PR #1, opened 2019-11-18
Refactoring/Upgrade: Exposing topics and other fixes

lintool:
This is the package currently deployed at TestPyPI: https://test.pypi.org/project/pyserini/0.6.1.post0/

The fatjar is based on `0ed488` in Anserini, dated `Sun Nov 17 19:08:26 2019 -0500`.

Working colab notebook at: https://colab.research.google.com/drive/1nY1bjwop3Enygrks-208EilhiCYmYKT6

Still need to update docs, but code itself is ready for CR I think.

I'd also appreciate opinions on publishing SNAPSHOT versions of the fatjar that aren't in Maven... Current package is on TestPyPI, so it really doesn't matter... but any strong opinions on publishing in official PyPI?

The tradeoff is between iteration speed vs. "doing it right". PyPI packages are supposed to be permanent, so depending on a snapshot jar is janky.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 .DS_Store
 __pycache__
+pyserini/resources/jars/*.jar
--- a/README.md
+++ b/README.md
@@ -15,17 +15,7 @@ Install via PyPI
 pip install pyserini
 ```
 
-Fetch the Anserini fatjar from Maven Central:
-
-```bash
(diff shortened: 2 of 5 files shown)


zeynepakkalyoncu:
I think we can hold off on publishing on PyPI for now? We can re-consider later when the Anserini/Pyserini connection is more mature and doesn't require as frequent updates?

lintool:
Okay, and continue to push out interim releases on TestPyPI for those on the real bleeding edge?

zeynepakkalyoncu:
Yes, sounds like a plan!

lintool:
I'll let @emmileaf take a look before merging...

emmileaf:
LGTM!

PR #5, opened 2019-11-22
Add/Refactoring IndexUtils #2

### LINKED ISSUE #2: Dump out tf-idf document vectors in pyserini
Is it possible to dump out tf-idf document vectors for retrieved documents in pyserini?

alirezamirzaeiyan:
Since DocVectorWeight enum is private in Anserini I couldn't port properly. This issue creates output files with wrong names(text.txt.docvector.(0-1).tar.gz). So it would be better to make DocVectorWeight public in IndexUtils.java. However, this piece of code works correctly.

### DIFF
--- a/pyserini/index/pyutils.py
+++ b/pyserini/index/pyutils.py
@@ -17,8 +17,7 @@
 '''
 Module for providing python interface to Anserini index utils
 '''
-from enum import Enum
-from ..pyclass import JIndexUtils, JString
+from ..pyclass import JIndexUtils, JDocumentVectorWeight, JString
 
 import logging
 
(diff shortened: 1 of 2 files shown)


pyserini/index/pyutils.py
```diff
@@ -0,0 +1,54 @@
...
+    '''
+
+    def __init__(self, index_dir):
+        self.object = JIndexUtils(JString(index_dir))
+
+    class DocVectorWeight(Enum):
+        NONE = 0
+        TF_IDF = 1
+
+    def dumpDocumentVectors(self, reqDocidsPath: str, weight: DocVectorWeight):
```
lintool: rename to Python conventions? i.e., `dump_document_vectors`?

lintool:
> So it would be better to make DocVectorWeight public in IndexUtils.java

Yes, makes sense. Please send a PR in anserini directly?
Should we rename to `DocumentVectorWeight` while we're at it to be consistent?

pyserini/index/pyutils.py
alirezamirzaeiyan: Yes :)

alirezamirzaeiyan:
Yes, I'll send a PR in anserini asap. It would be very great to be consistent, by the way, should I rename to DocumentVectorWeight in anserini?

lintool:
> should I rename to DocumentVectorWeight in anserini?

yes please.

lintool:
It would be even more amazing if you could write a test case!

alirezamirzaeiyan:
I've fixed enum problem by a PR in anserini and another commit in pyserini.

alirezamirzaeiyan:
> t would be even more amazing if you could write a test case!

Do you mean to write a test case for dumpDocumentVectors in anserini?

lintool:
Yes

PR #9, opened 2019-11-25
Update README.md

lintool:
To reduce duplication between Anserini and Pyserini, I've moved the focus of the notebook stuff to a separate repo: https://github.com/castorini/anserini-notebooks

Added pointer here.


zeynepakkalyoncu:
Just to check, should the AFFIRM notebooks be stored in the anserini-notebooks repo as well?

lintool:
Why don't we put them in a separate repo for now, e.g., `anserini-notebooks-afirm2020`? We can merge them back into this when things have stabilized a bit?

You should be able to create a repo in `castorini` - let me know if you can't...

PR #10, opened 2019-11-26
Add Python interface for IndexReaderUtils methods

zeynepakkalyoncu:


pyserini/index/pyutils.py
```diff
@@ -35,13 +35,105 @@ class IndexUtils:
...
+            if self.positions:
+                repr += ' [' + ','.join([str(p) for p in self.positions]) + ']'
+            return repr
+
+    def analyze_term(self, term):
+        '''
+        Parameters
+        ----------
+        term : str
+            Search term
```
lintool: just "term" - not necessary search, I think? Same below.

pyserini/index/pyutils.py
```diff
@@ -35,13 +35,105 @@ class IndexUtils:
...
+            Stemmed term
+        '''
+        return self.object.analyzeTerm(self.reader, JString(term))
+
+    def get_term_counts(self, term):
+        '''
+        Parameters
+        ----------
+        term : str
+            Search term
```
lintool: note that this is before analysis/stemming?

pyserini/index/pyutils.py
```diff
@@ -35,13 +35,105 @@ class IndexUtils:
...
+        for posting in postings_list.toArray():
+            result.append(self.Posting(posting.getDocid(), posting.getTF(), posting.getPositions()))
+        return result
+
+    def get_document_vector(self, docid):
+        '''
+        Parameters
+        ----------
+        docid : str
+            Lucene document ID
```
lintool: I believe this is the docid as defined in the collection? call it "collection docid" to be unambiguous. This is different from the Lucene internal docid.

So "Lucene document ID" is ambiguous.
Also below.

pyserini/index/pyutils.py
zeynepakkalyoncu: True, I just followed the order in Anserini. Should I change both then?

lintool: sorry, wasn't clear - I meant, "term prior to stemming/analysis"

zeynepakkalyoncu: Ahh sorry, my bad... Fixed.

PR #14, opened 2019-12-07
Add wrapper over iterator over terms

zeynepakkalyoncu:


pyserini/index/pyutils.py
```diff
@@ -37,6 +37,17 @@ class IndexReaderUtils:
     def __init__(self, index_dir):
         self.object = JIndexReaderUtils()
         self.reader = self.object.getReader(JString(index_dir))
+        self.term_iterator = self.object.getTermIterator(self.reader)
+
+    def __iter__(self):
```
lintool: I'm a bit confused... don't we want a `get_terms` method that returns the iterator?

zeynepakkalyoncu: Hmm could you elaborate on the method that you had in mind? My goal was to be able to do something like this by defining a Python iterator with `__iter__` and `__next__`:

```
index = pyutils.IndexReaderUtils('/path/to/index/')

for term in index:
  print(term.docFreq())
```

(It is also possible to implement a wrapper over the TermsEnum functions (e.g., docFreq(), postings()) of course.)

lintool: Yes, basically, but:

```python
for term in index.terms():
   # do something

# and you can also have...

for doc_vector in index.doc_vectors():
   # do something

# and

for doc in index.doc():
   # do something
```

How's that?

PR #28, opened 2020-01-07
Fetch raw document

zeynepakkalyoncu:
Add functionality to fetch raw document given its collection docid

Python wrapper over getRawDocument: https://github.com/castorini/anserini/pull/937


lintool:
hey @zeynepakkalyoncu can we add a test case on the Python end also? 
here: https://github.com/castorini/pyserini/blob/master/tests/test_indexutils.py

PR #30, opened 2020-01-16
Mirror update in analyze method

zeynepakkalyoncu:


lintool:
Please add/modify test case?

PR #37, opened 2020-02-15
Remove obsolete method dump_document_vectors

### LINKED ISSUE #35: Use case for dump_document_vectors in IndexReaderUtils?
Hi @zeynepakkalyoncu - do we have an actual use case for `dump_document_vectors` in `IndexReaderUtils`? I'm trying to write a use case for it, and I can't seem to even get it to work? Throws a mysterious `jnius.JavaException: JVM exception occurred` error.

If you're using this feature somewhere (in a notebook), we should make sure it works and write a test case... otherwise I would suggest removing it until a real use case comes up.

zeynepakkalyoncu:
Resolves #35

### DIFF
--- a/pyserini/index/pyutils.py
+++ b/pyserini/index/pyutils.py
@@ -150,17 +150,6 @@ class IndexReaderUtils:
         """
         return self.object.getBM25TermWeight(self.reader, JString(docid), JString(term))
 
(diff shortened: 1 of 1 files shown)


lintool:
hi @zeynepakkalyoncu let me fold this in as part of a branch I'm working on to avoid conflicts.

PR #40, opened 2020-02-15
Simplify requirements to include only pyjnius

lintool:
I think `pyjnius` is the only direct dependency we need? Others are pulled in transitively, but we're not particularly picky about versions, etc.? Is that right?


lintool:
hey @emmileaf did you want to comment on this?

lintool:
According to this, we should only include the direct dependencies:
https://www.reddit.com/r/Python/comments/32q1gy/should_i_include_transitive_dependencies_in_my/

I'm going to merge this because my local pycharm warning is driving me nuts.

PR #41, opened 2020-02-17
Update README.md

lintool:
Example rewritten using f-strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ hits = searcher.search('hubble space telescope')
 
 # Prints the first 10 hits
 for i in range(0, 10):
-    print('{} {} {}'.format(i+1, hits[i].docid, hits[i].score))
+    print(f'{i+1} {hits[i].docid} {hits[i].score}')
 
 # Grab the actual text
 hits[0].content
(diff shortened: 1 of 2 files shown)


lintool:
Is that a safe assumption to make you think? What's the acceptance/adoption in industry in your experience?

emmileaf:
> Is that a safe assumption to make you think? What's the acceptance/adoption in industry in your experience?

>=3.5 might be a safer assumption imo, although it probably isn't a big deal... I can't recall where else in the package has the >=3.6 dependency, but would probably feel less justified to enforce it if f-strings is the only version-dependent feature? If something else already requires the 3.6+ though, then that's a different story, of course.

lintool:
@emmileaf I asked Twitter... let's see what people say: https://twitter.com/lintool/status/1229480636479660034

lintool:
hey @emmileaf - @tuzhucheng makes a convincing case, I think:
https://twitter.com/tuzhucheng/status/1229529691519774720

<blockquote class="twitter-tweet"><p lang="en" dir="ltr">Given that Python 3.6 was first released in December 2016, I&#39;d say those using older releases should upgrade, given the pace of how fast things change now. It&#39;s also quite easy to create virtual environments with different Python 3.x versions.</p>&mdash; Michael Tu (@tuzhucheng) <a href="https://twitter.com/tuzhucheng/status/1229529691519774720?ref_src=twsrc%5Etfw">February 17, 2020</a></blockquote>

Thoughts?

emmileaf:
Hmm that's a good point too, legacy users gotta upgrade sometime or another. Let’s roll forward with the f-strings then?

PR #48, opened 2020-03-06
API for analyzers, option to change analyzer in pysearch

### LINKED ISSUE #43: Add functionality to search without preprocessing
We would like to be able to search an index without the query being processed by the stemmer. The specific use-case would be for the background linking task of TREC. We want to use the document vectors (that contain stemmed terms) to construct a new query.

### LINKED ISSUE #44: Clean API to expose Lucene Analyzers
Related to #43 - currently `IndexReaderUtils` exposes `analyze` method:
https://github.com/castorini/pyserini/blob/master/pyserini/index/pyutils.py

This is hard coded to a default. We should think about how to expose arbitrary Lucene analyzers in general... what would the API look like?

Chriskamphuis:
Getter function for anserini analyzers. 
In pysearch it is possible to set these analyzers such that they will be used when searching. 

fix #43 
fix #44


pyserini/analysis/pyanalysis.py
```diff
@@ -0,0 +1,131 @@
...
+from ..pyclass import JFreebaseAnalyzer
+from ..pyclass import JFrenchAnalyzer
+from ..pyclass import JGermanAnalyzer
+from ..pyclass import JHindiAnalyzer
+from ..pyclass import JSpanishAnalyzer
+from ..pyclass import JTokenizeOnlyAnalyzer
+from ..pyclass import JTweetAnalyzer
+
+
+def get_default_analyzer():
```
lintool: What about `get_analyzer()` as a function that takes a enum to specify which one? And if called with zero args, returns the default?

Chriskamphuis: Sure will implement

lintool:
@Chriskamphuis can we get some test cases please? 🙏

PR #49, opened 2020-03-06
Fixed broken tests due to Anserini 0.7.3-SNAPSHOT

lintool:
We've made changes to Anserini head that are breaking changes to Pyserini. These break existing Pyserini unit tests.

After merging PR, we'll have the opposite problem: Pyserini unit tests will now fail on the official Maven Anserini 0.7.2 artifact.

There's no good way around this issue.


Chriskamphuis:
works for me

PR #50, opened 2020-03-06
Fix for ValueError: chr() arg not in range

PepijnBoers:
Fix for ‘ValueError: chr() arg not in range(0x110000)’ when calling a method from IndexReaderUtils (in pyutils) that contains a JString(term). For example after calling the get_document_vector(docid) method. This can be caused by a rare character in a term (See: https://github.com/kivy/pyjnius/issues/437). 

Specifying utf-8 encoding for the term solved this error: JString(term.encode(‘utf-8’)).

### DIFF
--- a/pyserini/index/pyutils.py
+++ b/pyserini/index/pyutils.py
@@ -132,7 +132,7 @@ class IndexReaderUtils:
         Tuple[int, int]
             The document frequency and collection frequency of the term.
         """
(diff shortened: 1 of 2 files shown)


lintool:
@PepijnBoers can we have a test case?

PepijnBoers:
Yes! I also wrote a test case that fails when Pyjnius has fixed this internally. Indicating that we can remove the (then redundant) encoding.

PR #62, opened 2020-03-22
Refactoring to match underlying Anserini API changes

lintool:


lintool:
@nikhilro yes, I need to publish new Anserini artifact before these changes will work. Just lining everything up right now.

PR #64, opened 2020-03-22
added fake words and lexical LSH Analyzers to pyserini (for ANN search)

tteofili:


lintool:
Holding off merging, per discussion on Slack DM.

tteofili:
I'm closing it for now as this is not needed anymore since #66 addresses the aNN case better.

PR #66, opened 2020-04-05
added SimpleNNSearcher class

tteofili:
the `SimpleNNSearcher` class exposes Anserini's `SimpleNearestNeighborSearcher` features in pyserini.

### DIFF
--- a/pyserini/pyclass.py
+++ b/pyserini/pyclass.py
@@ -57,6 +57,8 @@ JTweetAnalyzer = autoclass('io.anserini.analysis.TweetAnalyzer')
 JDocument = autoclass('org.apache.lucene.document.Document')
 JSearcher = autoclass('io.anserini.search.SimpleSearcher')
 JResult = autoclass('io.anserini.search.SimpleSearcher$Result')
+JSnnSearcher = autoclass('io.anserini.search.SimpleNearestNeighborSearcher')
+JSnnResult = autoclass('io.anserini.search.SimpleNearestNeighborSearcher$Result')
 
 ### Topics
 
--- a/pyserini/pyclass.py
+++ b/pyserini/pyclass.py
@@ -55,9 +55,9 @@ JTweetAnalyzer = autoclass('io.anserini.analysis.TweetAnalyzer')
 ### Search
 
 JDocument = autoclass('org.apache.lucene.document.Document')
-JSearcher = autoclass('io.anserini.search.SimpleSearcher')
+JSimpleSearcher = autoclass('io.anserini.search.SimpleSearcher')
 JResult = autoclass('io.anserini.search.SimpleSearcher$Result')
-JSnnSearcher = autoclass('io.anserini.search.SimpleNearestNeighborSearcher')
+JSimpleNearestNeighborSearcher = autoclass('io.anserini.search.SimpleNearestNeighborSearcher')
 JSnnResult = autoclass('io.anserini.search.SimpleNearestNeighborSearcher$Result')
 
 ### Topics
--- a/pyserini/pyclass.py
+++ b/pyserini/pyclass.py
@@ -56,9 +56,9 @@ JTweetAnalyzer = autoclass('io.anserini.analysis.TweetAnalyzer')
 
 JDocument = autoclass('org.apache.lucene.document.Document')
 JSimpleSearcher = autoclass('io.anserini.search.SimpleSearcher')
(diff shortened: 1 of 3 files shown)


pyserini/search/pysearch.py
```diff
@@ -252,3 +252,25 @@ def get_topics(collection_name):
         for key in topics.get(topic).keySet().toArray():
             t[topic][key] = topics.get(topic).get(key)
     return t
+
+class SimpleNNSearcher:
```
lintool, line 256: `SimpleNearestNeighborSearcher` to be consistent with Java?

tteofili: ok, makes sense

tteofili:
I updated the class name to reflect the Anserini/Java one.

lintool:
hey @tteofili I'm wondering if the `Results` class can be shared?

```
JResult = autoclass('io.anserini.search.SimpleSearcher$Result')
JSnnResult = autoclass('io.anserini.search.SimpleNearestNeighborSearcher$Result')
```

Yes, would require more hacking on the Java side, but if you think it's a good idea, we should take the time to do it? Since we're introducing API changes anyway?

pyserini/pyclass.py
```diff
@@ -57,6 +57,8 @@
 JDocument = autoclass('org.apache.lucene.document.Document')
 JSearcher = autoclass('io.anserini.search.SimpleSearcher')
```
lintool, line 58: Should we rename to `JSimpleSearcher` and `JSimpleNearestNeighborSearcher` just to make everything consistent?

tteofili: sure, good point, done.

tteofili:
> hey @tteofili I'm wondering if the `Results` class can be shared?
> 
> ```
> JResult = autoclass('io.anserini.search.SimpleSearcher$Result')
> JSnnResult = autoclass('io.anserini.search.SimpleNearestNeighborSearcher$Result')
> ```
> 
> Yes, would require more hacking on the Java side, but if you think it's a good idea, we should take the time to do it? Since we're introducing API changes anyway?

I would avoid sharing the same class as it wouldn't make much sense to have `result.raw` and `result.contents` for the embedding case, so I had thought it was better to have a dedicated (and simpler) object for the nearest neighbor case.

lintool:
@tteofili sounds good, agreed.

pyserini/pyclass.py
```diff
@@ -55,8 +55,10 @@
 ### Search
 
 JDocument = autoclass('org.apache.lucene.document.Document')
-JSearcher = autoclass('io.anserini.search.SimpleSearcher')
+JSimpleSearcher = autoclass('io.anserini.search.SimpleSearcher')
 JResult = autoclass('io.anserini.search.SimpleSearcher$Result')
```
lintool, line 59: In the same way, `JSimpleSearcherResult` and `JSimpleNearestNeighborSearcherResult`? I know it's a bit long...
or `JSearchResult` or `JNearestNeighborResult`?

tteofili: sure, makes sense :+1:

pyserini/search/pysearch.py
```diff
@@ -292,3 +293,25 @@ def get_topics(collection_name):
         for key in topics.get(topic).keySet().toArray():
             t[topic][key] = topics.get(topic).get(key)
     return t
+
+class SimpleNearestNeighborSearcher:
+
+    def __init__(self, index_dir: str):
+        self.object = JSimpleNearestNeighborSearcher(JString(index_dir))
+
+    def search(self, q: str, k=10) -> List[List[JSimpleNearestNeighborSearcherResult]]:
```
lintool, line 302: Is the return type here correct?

It's dispatch to this method right? `public Result[] search(String id, int d)`
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleNearestNeighborSearcher.java#L78

tteofili: yes, correct. I was fixing it :)

lintool:
Would it make sense to write a test case?
Like what I do here?
https://github.com/castorini/pyserini/blob/master/tests/test_search.py#L30

I suggest we use CACM? We can store the index here also? https://github.com/castorini/anserini-data/tree/master/CACM

tteofili:
great point, I'll add such a test case.

tteofili:
I've added a test case that downloads a vectors index for the CACM data and runs a nearest neighbor query.

PR #71, opened 2020-04-17
Better alignment of API in SimpleSearcher and IndexReaderUtils for access docs

lintool:
Better testing. Uploaded new version of CACM index with `contents` stored, and thus, testable.

### DIFF
--- a/pyserini/search/pysearch.py
+++ b/pyserini/search/pysearch.py
@@ -42,6 +42,10 @@ class Document:
     def docid(self: JDocument) -> str:
         return self.object.getField('id').stringValue()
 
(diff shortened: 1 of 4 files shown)


pyserini/search/pysearch.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, document):
     def docid(self: JDocument) -> str:
         return self.object.getField('id').stringValue()
 
+    def id(self: JDocument) -> str:
```
rodrigonogueira4, line 45: I don't think it is common to tell explicitly the type of self (`JDocument`), but that is ok

PR #83, opened 2020-05-04
Expose phrase df in simple searcher 

PepijnBoers:
Exposing SimpleSearcher's [phraseDf](https://github.com/castorini/anserini/commit/1b45b6b859deced6470ce0a7150b5eceaf9972bd) method.

### DIFF
--- a/pyserini/search/pysearch.py
+++ b/pyserini/search/pysearch.py
@@ -254,6 +254,22 @@ class SimpleSearcher:
         """
         return Document(self.object.documentByField(JString(field), JString(q)))
 
+    def phrase_df(self, phrase: str) -> int:
+        """Constructs a phrase query and uses the TotalHitCountCollector to find the number of documents containing 
+        the phrase.
+
+        Parameters
+        ----------
+        phrase : str
+            phrase query
+
+        Returns
+        -------
+        int
+            number of documents containing phrase
+        """             
+        return self.object.phraseDf(JString(phrase))
+
     def close(self):
         self.object.close()
 
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -165,10 +165,15 @@ class TestSearch(unittest.TestCase):
         self.assertEqual(self.searcher.doc('CACM-3134').docid(),
                          self.searcher.doc_by_field('id', 'CACM-3134').docid())
 
+    def test_phrase_df(self):
+        self.assertEqual(74, self.searcher.phrase_df("inform retriev"))
+        self.assertEqual(135, self.searcher.phrase_df("search"))
+        
     def tearDown(self):
         self.searcher.close()
         os.remove(self.tarball_name)
         shutil.rmtree(self.index_dir)
+        
 
 
 if __name__ == '__main__':


tests/test_search.py
```diff
@@ -165,10 +165,15 @@ def test_doc_by_field(self):
...
 
+    def test_phrase_df(self):
+        self.assertEqual(74, self.searcher.phrase_df("inform retriev"))
+        self.assertEqual(135, self.searcher.phrase_df("search"))
+        
     def tearDown(self):
         self.searcher.close()
         os.remove(self.tarball_name)
         shutil.rmtree(self.index_dir)
+
```
lintool, line 176: I think PEP8 will complain this as too having too many empty lines?

pyserini/search/pysearch.py
```diff
@@ -254,6 +254,22 @@ def doc_by_field(self, field: str, q: str) -> str:
         """
         return Document(self.object.documentByField(JString(field), JString(q)))
 
+    def phrase_df(self, phrase: str) -> int:
+        """Constructs a phrase query and uses the TotalHitCountCollector to find the number of documents containing
```
lintool, line 258: I don't think there's a need to expose the implementation detail here? `TotalHitCountCollector`
Also be clear whether you expect an analyzed or unanalyzed phrase.
Any maybe we want a variant (optional arg) that does both?

tests/test_search.py
```diff
@@ -165,10 +165,15 @@ def test_doc_by_field(self):
         self.assertEqual(self.searcher.doc('CACM-3134').docid(),
                          self.searcher.doc_by_field('id', 'CACM-3134').docid())
 
+    def test_phrase_df(self):
+        self.assertEqual(74, self.searcher.phrase_df("inform retriev"))
```
lintool, line 169: I assume you've sanity checked the output by other means? i.e., that df _is_ correct?

tests/test_search.py
```diff
@@ -165,10 +165,15 @@ def test_doc_by_field(self):
...
+        self.assertEqual(135, self.searcher.phrase_df("search"))
+
```
lintool, line 171: can you add assert that confirms phrase df for a single term is the same as the original method?

Actually, now that I think about it... why do we need both methods? What about one method that either takes phrases or individual terms?

PepijnBoers: What are you exactly referring to when you say 'original method'? If you mean IndexReaderUtils' `get_term_counts`, then it would be more intuitive to have the phrase_df function in the IndexReaderUtils I think.

lintool: yea, do you think it'd make more sense to but directly in `get_term_counts` and allow it to take phrases as well?

Chriskamphuis: want me to refactor it in Anserini and move it to that method?

PepijnBoers: Yes! That is exactly how I imagined to use it.

lintool: Let's do it then?

lintool: but we'll also need to get cf from a phrase?

Chriskamphuis: I can get all hits from the phrase query, loop over the documents and sum the tf for each hit?

Chriskamphuis: Actually I am not sure if you can get the tf easily given a phrase and a document..

lintool: If it's not easy, then make sure we return `None`?

Chriskamphuis: Ok!

lintool:
Closing per @Chriskamphuis 's request. Superseded by https://github.com/castorini/anserini/pull/1154

PR #89, opened 2020-05-07
Added support for CORD-19 articles

lintool:
+ Added `Cord19Article` to wrap raw JSON articles.
+ Added sample documentation.


lintool:
@stephaniewhoo Can you do a CR here also? As I mentioned, you should start your analyses with the raw collection, because the indexer already blacklists a few outlier articles.

PR #90, opened 2020-05-08
Flexible class loader

Chriskamphuis:
Now its not possible to load classes through pyclass and also load in custom classes through Pyjnius, as it is not possible to configure the classpath twice. Using this helper function it also possible to load new Java classes through pyclass.


lintool:
Java classes can be loaded with `autoclass(class_name)`... which is basically what this PR does... so it just provides an alias? Not sure what this adds, or I'm missing something?

Chriskamphuis:
nvm you are right, dont know why I thought this was not possible.

PR #91, opened 2020-05-08
Query construction

Chriskamphuis:
Make it possible to construct queries in Pyserini

### DIFF
--- a/pyserini/search/pyquerybuilder.py
+++ b/pyserini/search/pyquerybuilder.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+#
+# Pyserini: Python interface to the Anserini IR toolkit built on Lucene
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module provides Pyserini's Python interface query building for Anserini.
+"""
+import logging
+from ..pyclass import JQueryGeneratorUtils, JTermQuery, JTerm, JBoostQuery
+from ..analysis.pyanalysis import get_lucene_analyzer, Analyzer
+
+logger = logging.getLogger(__name__)
+
+
+def get_boolean_query_builder():
+    """ Get a BooleanQueryBuilder object.
+
+    Returns
+    -------
+    JBooleanQueryBuilder
+    """
+    return JQueryGeneratorUtils.getBooleanQueryBuilder()
+
+
+def get_clause_should():
+    """ Get a BooleanClause.Occur.SHOULD statement
+
+    Returns
+    -------
+    BooleanClause.Occur.SHOULD
+    """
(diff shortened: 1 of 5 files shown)


pyserini/search/pyquerybuilder.py
```diff
@@ -0,0 +1,111 @@
...
+"""
+import logging
+from ..pyclass import JQueryGeneratorUtils, JTermQuery, JTerm, JBoostQuery
+from ..analysis.pyanalysis import get_lucene_analyzer, Analyzer
+
+logger = logging.getLogger(__name__)
+
+
+def get_boolean_query_builder():
+    """ Get a BooleanQueryBuilder object.
```
lintool, line 28: extra space here, per PEP 8. and below.

pyserini/search/pyquerybuilder.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Returns
+    -------
+    JBooleanQueryBuilder
+    """
+    return JQueryGeneratorUtils.getBooleanQueryBuilder()
+
+
+def get_clause_should():
+    """ Get a BooleanClause.Occur.SHOULD statement
```
lintool, line 38: PEP8 - docstring comments should end in periods.

pyserini/search/pysearch.py
```diff
@@ -72,14 +72,14 @@ class SimpleSearcher:
     def __init__(self, index_dir: str):
         self.object = JSimpleSearcher(JString(index_dir))
 
-    def search(self, q: str, k: int=10, t: int=-1,
+    def search(self, q, k: int=10, t: int=-1,
```
lintool, line 75: We can still provide a type hint here? using `Union`?

pyserini/search/pyquerybuilder.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Returns
+    -------
+    BooleanClause.Occur.SHOULD
+    """
+    return JQueryGeneratorUtils.getBooleanClauseShould()
+
+
+def get_clause_must():
```
lintool, line 47: Would it be better to implement an Enum for these? Like here? https://github.com/castorini/pyserini/blob/master/pyserini/pyclass.py#L103

lintool: Hrm. Wait, or maybe somehow bind them to singleton symbols?

Dunno...

tests/test_querybuilding.py
```diff
@@ -0,0 +1,125 @@
...
+
+        boolean_query1 = pyquerybuilder.get_boolean_query_builder()
+        boolean_query1.add(term_query1, should)
+        boolean_query1.add(term_query2, should)
+
+        bq1 = boolean_query1.build()
+        hits1 = self.searcher.search(bq1)
+        hits2 = self.searcher.search('information retrieval')
+
+        self.assertEqual(hits1[0].docid, hits2[0].docid)
```
lintool, line 115: Can you iterate through all the hits to check all scores and docids?

tests/test_querybuilding.py
```diff
@@ -0,0 +1,125 @@
...
+        boolean_query2.add(term_query1, should)
+        boolean_query2.add(term_query2, should)
+
+        bq2 = boolean_query2.build()
+        hits2 = self.searcher.search(bq2)
+
+        self.assertEqual(hits1[0].docid, hits2[0].docid)
+        self.assertEqual(hits1[0].score, hits2[0].score)
+
+        boost3 = pyquerybuilder.get_boost_query(term_query1, 2.)
```
lintool, line 70: Maybe add a test case where you assign both weight 2 - docid (in entire hits) should be the same, but scores different?

PR #92, opened 2020-05-09
Improved documentation for IndexReaderUtils

lintool:
Better interleaving of explanations for code.

@PepijnBoers take a look also?

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ To compute the tf-idf representation of a document, do something like this:
 ```python
 tf = index_utils.get_document_vector('FBIS4-67701')
 analyzer = pyanalysis.get_lucene_analyzer(stemming=False, stopwords=False)
-df = {term: (index_utils.get_term_counts(term, analyzer=analyzer))[1] for term in tf.keys()}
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,85 +71,136 @@ Pyserini exposes Lucene Analyzers in Python with the `Analyzer` class.
...
 print(doc_vector)
+```
+
+The result is a dictionary where the keys are the analyzed terms and the values are the term frequencies.
+To compute the tf-idf representation of a document, do something like this:
+
+```python
+tf = index_utils.get_document_vector('FBIS4-67701')
+analyzer = pyanalysis.get_lucene_analyzer(stemming=False, stopwords=False)
+df = {term: (index_utils.get_term_counts(term, analyzer=analyzer))[1] for term in tf.keys()}
```
PepijnBoers, line 183: Should probably take the first element from the `get_term_counts`, this refers to the cf right? I think I wrote it incorrectly in an issue, sorry!

fix:
```python 
df = {term: (index_utils.get_term_counts(term, analyzer=analyzer))[0] for term in tf.keys()}
```

lintool: yup, good catch!

PR #95, opened 2020-05-09
Create documentation for Query building

Chriskamphuis:
Create an example on how to construct queries in Pyserini

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ hits = searcher.search('hubble space telescope')
 # Print the first 10 hits:
 for i in range(0, 10):
     print(f'{i+1:2} {hits[i].docid:15} {hits[i].score:.5f}')
+    
+# 1 LA071090-0047   16.85690
+# 2 FT934-5418      16.75630
+# 3 FT921-7107      16.68290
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +230,39 @@ for (i, fs) in enumerate(collection):
...
+boolean_query = pyquerybuilder.get_boolean_query_builder()
+boolean_query.add(boost1, should)
+boolean_query.add(boost2, should)
+
+query = boolean_query.build()
+```
+
+Finally issue the constructed weighted query:
+```python
+hits1 = self.searcher.search(query)
```
lintool, line 263: Can you write this doc, making reference to the actual index above?

```
index_utils = pyutils.IndexReaderUtils('index-robust04-20191213/')
```

Then you can run queries, and compare against the default bow generator, and show that the results are the same?

Start with a query, run it through the analyzer, etc.?

Also, move to above "Usage of the Collection API"?

Chriskamphuis: I think the "Usage of the IndexReaderAPI" would be better suited then?

lintool: sg!

Issue #93, opened 2020-05-09
Make methods in IndexReaderUtils more consistent re: Analyzer


lintool:
We have:

```python
# Pass in a no-op analyzer:
analyzer = pyanalysis.get_lucene_analyzer(stemming=False, stopwords=False)
index_utils.get_term_counts(term, analyzer=analyzer)
df, cf = index_utils.get_term_counts(term)
```
Here, we take an `analyzer`.

And:

```python
# Fetch and traverse postings for an analyzed term:
postings_list = index_utils.get_postings_list(analyzed[0], analyze=False)
for posting in postings_list:
    print(f'docid={posting.docid}, tf={posting.tf}, pos={posting.positions}')
```

Here, we take a `bool`. Let's make both consistent?

How about both take `analyzer` and accepts `None`? Passing in a "no-op" analyzer seems a bit janky.

Thoughts? @PepijnBoers @Chriskamphuis

PepijnBoers:
Both taking an `analyzer` makes sense to me, especially since the `getPostingsListWithAnalyzer` method is already available in Anserini.

lintool:
@PepijnBoers do you have cycles to take this on?

PepijnBoers:
@lintool sure!

PepijnBoers:
Currently `get_term_counts` applies Anserini's default Lucene analyzer if no analyzer is specified (`analyzer=None`), then in order to skip term analysis you have to pass a dummy analyzer. Is this the requested behavior or do we want `analyzer=None` to mean that no analysis should take place?

lintool:
I'm thinking:
+ No analyzer specified - use default.
+ `analyzer=foo` - use `foo` as analyzer
+ `analyzer=None` - skip analysis

And we make this behavior across all methods (existing, and in the future).

Thoughts?

PepijnBoers:
I assumed that `None` would be the value for an unspecified analyzer, do you suggest we overload the methods in Pyserini and remove a default value for the `analyzer` parameter?

PepijnBoers:
We could do it like this:

```python
get_term_counts(self, term: str, analyzer='unspecified') -> Tuple[int, int]: 
    if analyzer is 'unspecified'":
        # use default
    elif analyzer is None:
        # skip analysis (pass dummy to Anserini)
    else:
       # perform analysis with given analyzer
    ...
```

lintool:
What about something simpler, like this?

```python
[...]

PR #96, opened 2020-05-09
Make methods in IndexReaderUtils more consistent

PepijnBoers:
Makes methods in IndexReadUtils more consistent (resolves https://github.com/castorini/pyserini/issues/93) + update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ print(f'term "{term}": df={df}, cf={cf}')
 ```
 
 What if we want to fetch term statistics for an analyzed term?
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -137,7 +137,7 @@ print(f'term "{term}": df={df}, cf={cf}')
 ```
 
 What if we want to fetch term statistics for an analyzed term?
-This can be accomplished by passing in an no-op `Analyzer`:
+This can be accomplished by setting `Analyzer` to None:
```
lintool, line 140: None -> `None`

PR #98, opened 2020-05-09
add a field to pysearcher made it expose to 'n' (number of docs in the index)

stephaniewhoo:
Questions: 
1. how to initialize a indexSearcher in pysearcher? 
I have checked available constructors either need a reader or readerContext. How can I access those two?

2. Is there any documents I can follow to run the whole pysenrini locally and test the changes?

### DIFF
--- a/pyserini/search/pysearch.py
+++ b/pyserini/search/pysearch.py
@@ -23,7 +23,7 @@ import logging
 from typing import Dict, List, Union
 
 from ..pyclass import JSimpleSearcher, JSimpleSearcherResult, JDocument, JString, JArrayList, JTopics, JTopicReader, \
-JQueryGenerator, JSimpleNearestNeighborSearcherResult, JSimpleNearestNeighborSearcher, JQuery
+    JQueryGenerator, JSimpleNearestNeighborSearcherResult, JSimpleNearestNeighborSearcher, JQuery
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


lintool:
1. I think you'll need to add a method in Anserini `SimpleSearcher`:
https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java

2. Unfortunately, because of the java/python repo split, you'll need to build a fatjar in Anserini, in `target/` and then copy fatjar over to replace the one in `pyserini/resources/jars` by hand.

Once everything is wired up correctly, I can generate a new Maven artifact.

pyserini/search/pysearch.py
```diff
@@ -23,7 +23,7 @@
 from typing import Dict, List, Union
 
 from ..pyclass import JSimpleSearcher, JSimpleSearcherResult, JDocument, JString, JArrayList, JTopics, JTopicReader, \
-    JQueryGenerator, JSimpleNearestNeighborSearcherResult, JSimpleNearestNeighborSearcher, JQuery
+JQueryGenerator, JSimpleNearestNeighborSearcherResult, JSimpleNearestNeighborSearcher, JQuery
```
lintool, line 26: I don't think this line needs to be changed?

pyserini/search/pysearch.py
```diff
@@ -71,6 +71,7 @@ class SimpleSearcher:
 
     def __init__(self, index_dir: str):
         self.object = JSimpleSearcher(JString(index_dir))
+        self.maxDoc = self.object.getTotalNumDocuments()
```
lintool, line 74: A variable name that is more pythonic? How about `number_of_docs` or even just `num_docs`?

lintool:
Can you change the name of your PR? Provide a more informative title?

PR #101, opened 2020-05-12
Fixed broken unit tests related to num_docs

lintool:
@Chriskamphuis @PepijnBoers I'd appreciate a sanity check on this... just _odd_.

On current HEAD, unit tests are broken: try with `python -m unittest`.

However, somewhat inexplicably, this patch fixes the bug...  and unit tests pass.

I don't really understand why... I suspect there's an underlying concurrency bug with how the `IndexReader` and `IndexSearcher` is being initialized on the Java end... any thoughts?

### DIFF
--- a/pyserini/search/pysearch.py
+++ b/pyserini/search/pysearch.py
@@ -71,9 +71,7 @@ class SimpleSearcher:
 
     def __init__(self, index_dir: str):
         self.object = JSimpleSearcher(JString(index_dir))
-
-    def num_docs(self):
-        return self.object.getTotalNumDocuments()
+        self.num_docs = self.object.getTotalNumDocuments()
 
     def search(self, q: Union[str, JQuery], k: int=10, t: int=-1,
                query_generator: JQueryGenerator=None) -> List[JSimpleSearcherResult]:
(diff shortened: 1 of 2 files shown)


PepijnBoers:
The main difference I see is that if we call the `getTotalNumDocuments` method before doing a search, the searcher gets initialized without setting a similarity via the `setSimilarity` method, whereas that would have been the case if we had done a search before calling the `num_docs`/`getTotalNumDocuments` method.

```java
   public int getTotalNumDocuments(){
     if (searcher == null) {
       searcher = new IndexSearcher(reader);
      }
     return searcher.getIndexReader().maxDoc();
   }
```

```java
  protected Result[] search(Query query, List<String> queryTokens, String queryString, int k,
                            long t) throws IOException {
    // Create an IndexSearch only once. Note that the object is thread safe.
    if (searcher == null) {
      searcher = new IndexSearcher(reader);
      searcher.setSimilarity(similarity);
    }
```

We can recreate the assertion error by calling the `num_docs` method before doing a search.

lintool:
@PepijnBoers Oh, I totally didn't notice this detail! Thanks for catching.

I sent this PR on the Pyserini end: https://github.com/castorini/anserini/pull/1188

@Chriskamphuis can you do a quick sign off?

PR #117, opened 2020-05-19
Pyserini Searchcollection

yuki617:


pyserini/search/__main__.py
```diff
@@ -0,0 +1,40 @@
+import re
+import argparse
+
+from pyserini.search.pysearch import get_topics,SimpleSearcher
+
+if __name__ == "__main__":
```
lintool, line 6: Do we still need this?

PR #118, opened 2020-05-19
Verify pyserini search results with anserini

x65han:
Introduce a new script to verify/compare search results between pyserini and anserini.

```
python scripts/verify_simplesearcher.py --anserini path/to/anserini_root
```

Currently supports:
- robust04
- robust05
- core17
- core18

### DIFF
--- a/pyserini/verify_search.py
+++ b/pyserini/verify_search.py
@@ -1,0 +1,107 @@
+import argparse
+import os
+import filecmp
+
+
+def remove_if_exist(path):
+    if os.path.exists(path):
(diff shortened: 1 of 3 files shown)


pyserini/verify_search.py
```diff
@@ -0,0 +1,107 @@
+import argparse
```
lintool, line 1: Add our boilerplate? https://github.com/castorini/pyserini/blob/master/pyserini/search/pysearch.py#L1

x65han: Added

PR #123, opened 2020-05-21
Pseudo Relevance Feedback Interpolation on pysearch

x65han:
Sample Command
----------------
```
python -m pyserini.search \
 -index /Users/johnson/anserini/indexes/lucene-index.robust04.pos+docvectors+raw \
 -topics robust04 \
 -output robust04-prf.txt \
 -prf -alpha 1
```
bm25 + LR on robust04 (alpha=1)
----------------
```
map                   	all	0.2550
P_30                  	all	0.3138
```

### DIFF
--- a/pyserini/search/__main__.py
+++ b/pyserini/search/__main__.py
@@ -1,3 +1,19 @@
+# -*- coding: utf-8 -*-
+#
+# Pyserini: Python interface to the Anserini IR toolkit built on Lucene
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import argparse
 import re
 from pyserini.search.pysearch import get_topics, SimpleSearcher
--- a/pyserini/search/__main__.py
+++ b/pyserini/search/__main__.py
@@ -36,6 +36,11 @@ if topics_dic == {}:
     print('Topic Not Found')
     exit()
 
+need_classifier = args.prf and len(args.prf) > 0 and args.alpha > 0
+if need_classifier is True:
+    ranker = PseudoRelevanceClassifierReranker(
+        args.index, args.prf, r=args.r, n=args.n, alpha=args.alpha)
+
 with open(args.output, 'w') as target_file:
     for index, topic in enumerate(sorted(topics_dic.keys())):
         print(f'Topic {topic}: {index + 1}/{len(topics_dic)}')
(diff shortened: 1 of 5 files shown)


pyserini/search/__main__.py
```diff
@@ -1,33 +1,55 @@
 import re
```
lintool, line 1: when you get a chance, please add in our standard boilerplate headers. and reranker below.

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+    LR = 'lr'
+    SVM = 'svm'
+
+
+class FusionMethod(enum.Enum):
+    AVG = 'avg'
+
+
+class PseudoRelevanceClassifierReranker:
+    def __init__(self, lucene_index: str, clf_type: List[ClassifierType], r=10, n=100, alpha=1):
```
lintool, line 20: you probably want alpha=0.5 as a default; defaults should do something reasonable...

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+from sklearn.linear_model import LogisticRegression
+from sklearn.svm import SVC
+from typing import List
+from ..vectorizer import TfidfVectorizer
+import uuid
+import os
+
+
+class ClassifierType(enum.Enum):
+    LR = 'lr'
```
lintool, line 11: Maybe we should just give it the full sklearn name?

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+    def set_classifier(self, clf_type: ClassifierType):
+        if clf_type == ClassifierType.LR:
+            self.clf = LogisticRegression()
+        elif clf_type == ClassifierType.SVM:
+            self.clf = SVC(kernel='linear', probability=True)
+        else:
+            raise Exception("Invalid classifier type")
+
+    def get_prf_vectors(self, doc_ids: List[str]):
+        train_docs = doc_ids[:self.r] + doc_ids[-self.n:]
```
lintool, line 40: Add a comment here to explain what you're doing here... notation might not be familiar to all - and line below.
This is actually a cute use of * notation.

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def set_classifier(self, clf_type: ClassifierType):
+        if clf_type == ClassifierType.LR:
+            self.clf = LogisticRegression()
+        elif clf_type == ClassifierType.SVM:
+            self.clf = SVC(kernel='linear', probability=True)
+        else:
+            raise Exception("Invalid classifier type")
+
+    def get_prf_vectors(self, doc_ids: List[str]):
```
lintool, line 39: this is intended to be private? use _ notation?

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+        return train_vecs, train_labels, test_vecs
+
+    def rerank_with_classifier(self, doc_ids: List[str], search_scores: List[float]):
+        train_vecs, train_labels, test_vecs = self.get_prf_vectors(
+            doc_ids)
+
+        # classification
+        self.clf.fit(train_vecs, train_labels)
+        pred = self.clf.predict_proba(test_vecs)
+        rank_scores = self.normalize([p[1] for p in pred])
```
lintool, line 55: rank_score -> classifier_scores to be more accurate

pyserini/search/reranker.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # interpolation
+        interpolated_scores = [a * self.alpha + b * (1-self.alpha)
+                               for a, b in zip(rank_scores, search_scores)]
+
+        return self.sort_dual_list(interpolated_scores, doc_ids)
+
+    def rerank(self, doc_ids: List[str], search_scores: List[float]):
```
lintool, line 63: I think it'd make sense to take as input an array of hits?
that way, other data (e.g., the document text) can come along for the ride as well (and be used as part of reranking in other reranker implementations).

x65han: Ok, I can do that. I was a little hesitant at first, since I don't know how to get the type of hits.

PR #126, opened 2020-05-22
Expose Anserini's IndexUtils (print stats only)

PepijnBoers:
Makes Anserini's printing index stats available.

Might cause some confusion when naming the IndexReaderUtils as `index_utils` (e.g. in the README).

This would probably be clearer:
```python

index_reader_utils = pyutils.IndexReaderUtils('indexes/index-robust04-20191213/')

index_utils = pyutils.IndexUtils('indexes/index-robust04-20191213/')
index_utils.print_index_stats()
```

### DIFF
--- a/pyserini/index/pyutils.py
+++ b/pyserini/index/pyutils.py
@@ -23,7 +23,7 @@ and methods provided are meant only to provide tools for examining an index and 
 import logging
 from typing import Dict, Iterator, List, Tuple
 
-from ..pyclass import JIndexReaderUtils, JString, JAnalyzerUtils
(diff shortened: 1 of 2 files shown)


lintool:
Hey @PepijnBoers I'm -1 on this. `IndexUtils` is legacy and `IndexReaderUtils` is the updated refactored class... so we shouldn't expose `IndexUtils`, just causes more headaches down the road.

Furthermore, we shouldn't just print, but actually return the stats in some data structure (e.g., map?).

So, to "do this right", I propose impl new functionality in `IndexReaderUtils` and then hooking into pyserini.

Thoughts?

PepijnBoers:
Ah yea, that's probably better!

PR #128, opened 2020-05-23
Expose index statistics

PepijnBoers:
Pyserini end of https://github.com/castorini/anserini/pull/1218

### DIFF
--- a/pyserini/index/pyutils.py
+++ b/pyserini/index/pyutils.py
@@ -337,30 +337,17 @@ class IndexReaderUtils:
         """
         return self.object.convertDocidToLuceneDocid(self.reader, docid)
 
-    def stats(self) -> dict:
+    def stats(self) -> Dict[str, int]:
         """Returns dictionary with index statistics.
 
         Returns
         -------
-        dict
+        Dict[str, int]
             Index statistics as a dictionary of statistic's name to statistic.
-
-            Quantitative stats:
             - documents: number of documents
             - non_empty_documents: number of non-empty documents
             - unique_terms: number of unique terms
             - total_terms: number of total terms
-
-            stored fields:
-            - fullCaption
-            - author
-            - raw
-            - article_url
-            - title
-            - contents
-            - id
-            - published_date
-            - kicker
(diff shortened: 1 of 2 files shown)


pyserini/index/pyutils.py
```diff
@@ -336,3 +336,22 @@ def convert_collection_docid_to_internal_docid(self, docid: str) -> int:
             The Lucene internal ``docid`` corresponding to the external collection ``docid``.
         """
         return self.object.convertDocidToLuceneDocid(self.reader, docid)
+
+    def stats(self) -> dict:
```
lintool, line 340: should we add type hints here?
and below, do some explicit type checking?

pyserini/index/pyutils.py
```diff
@@ -336,3 +336,22 @@ def convert_collection_docid_to_internal_docid(self, docid: str) -> int:
...
+        """Returns dictionary with index statistics.
+
+        Returns
+        -------
+        dict
+            Index statistics as a dictionary of statistic's name to statistic.
```
lintool, line 346: would be nice to just list the statistics we provide? so user doesn't need to go hunting in the Java code?

pyserini/index/pyutils.py
PepijnBoers: All keys have type `str` so that would be possible but, the values vary between `int` and `str`, I'm not sure how to hint/type check without the code getting messy..

```python
def stats(self) -> Dict[str, int or str]:
```

Looks kinda strange to me, thoughts?

lintool: Can the values be string too?

PepijnBoers: Yes, the original `printIndexStats` in Anserini's `IndexUtils` also showed the stored fields as string, e.g. for the 'title' field: 

`(indexOption: DOCS, hasVectors: false)`

lintool: Hrm, I see.
Yea, `def stats(self) -> dict:` is fine I think.

PR #129, opened 2020-05-23
cord19_length_outlier_analysis

stephaniewhoo:
This is a python script to analyze the length of cord19 documents and find the extreme outliers (which may have negative affect to index size) and temporarily return a csv.


scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
+from pyserini.collection import pycollection
+import pandas as pd
+
+
+
+if __name__ == '__main__':
```
lintool, line 6: don't hard code args, allow something like `-collection` and `-output` to specify input and output.

scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
+from pyserini.collection import pycollection
```
lintool, line 1: actually, why don't you start a `scripts/cord19` subdir and move this in there?

lintool, line 1: throw in our standard boilerplate? https://github.com/castorini/pyserini/blob/master/pyserini/setup.py#L1

scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
...
+import pandas as pd
+
+
+
+if __name__ == '__main__':
+    titles_withText = []
+    abstracts_withText = []
```
lintool, line 8: `_with_text` to be consistently pythonic?

lintool: Better yet, why not a dictionary?

scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
...
+                dois_withText.append(article.metadata()["doi"])
+                titles_withText.append(len(article.title()))
+                abstracts_withText.append(len(article.abstract()))
+                paragraph_num.append(len(article.body()))
+                abstracts_total.append(abstracts_withText[-1] * paragraph_num[-1])
+                titles_total.append(titles_withText[-1] * paragraph_num[-1])
+
+    documents_withText = {'docid': docids_withText, 'DOI': dois_withText, 'title': titles_total, 'abstract': abstracts_total, 'num_para': paragraph_num}
+    documents_withoutText = {'docid': docids_withoutText, 'DOI': dois_withoutText, 'title': titles_withoutText, 'abstract': abstracts_withoutText}
+    documents_withText_df = pd.DataFrame(documents_withText)
```
lintool, line 56: If you're going to use pd, which is fine, why not just use pd to compute all the above conditions?

scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
...
+    # using quantile to find outliers return a df
+    q1 = documents_withText_df['abstract'].quantile(0.25)
+    q3 = documents_withText_df['abstract'].quantile(0.75)
+    iqr = q3 - q1
+    # I only consider extreme big value, since small value will not cause file explosion
+    filter = documents_withText_df['abstract'] >= q3 + 1.5 * iqr  # could choose other coefficient to find more "extreme" outlier
+    # abstract outlier
+    outlier = documents_withText_df.loc[filter]
+    #temporarily choose 5
```
lintool, line 65: once again, should be arg passed in from command line, with defaults.

scripts/find_cord19_length_outliers.py
```diff
@@ -0,0 +1,68 @@
...
+    sorted = outlier.sort_values(by=['abstract'], ascending=False)
+    top_5 = sorted.head(5)
+    top_5.to_csv('length_outlier_top5.csv')
```
lintool, line 68: should produce some type of output to stdout... like "found X outlier documents, writing results to ...."

scripts/find_cord19_length_outliers.py
stephaniewhoo: done parser agrs

scripts/find_cord19_length_outliers.py
stephaniewhoo: I don't really know how should this import be in the standard boilerplate...

scripts/find_cord19_length_outliers.py
stephaniewhoo: I updated with a dictionary

scripts/find_cord19_length_outliers.py
stephaniewhoo: I am not sure how to use pd to compute above conditions. I iterate through raw documents and add to dictionaries. Then transform dic to pd because pd will be easier to computer later outliers.

PR #140, opened 2020-05-26
__main__.py should use tqdm for progress update and have better args naming

x65han:
- Updated python args to `pyserini/search/__main__.py.
- Utlizing `tqdm` for progress update

-----------------------------
- Leaving `-output` as optional so __main__.py can decide the output file name.
- Didn't write the test case yet, since `cacm` doesn't appear to have a pre-parsed topics in [Topics.java](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/topicreader/Topics.java). We also have `verify_simplesearcher.py` already that tests __main__.py.


pyserini/search/__main__.py
```diff
@@ -16,30 +16,30 @@
...
 import re
+from tqdm import tqdm
 from pyserini.search.pysearch import get_topics, SimpleSearcher
 from pyserini.search.reranker import ClassifierType, PseudoRelevanceClassifierReranker
 
-
 parser = argparse.ArgumentParser(description='Create a input schema')
-parser.add_argument('-index', metavar='path', required=True,
+parser.add_argument('-index', type=str, metavar='path', required=True,
                     help='the path to workspace')
```
lintool, line 25: path to index?

also, we're using line width of 100, so some of these can be on a single line.

pyserini/search/__main__.py
```diff
@@ -16,30 +16,30 @@
...
 parser.add_argument('-bm25',  action='store_true', default=True,
-                    help='use bm25 ranker')
+                    help='use bm25 ranker by default')
 parser.add_argument('-rm3',  action='store_true',
                     help='use rm3 ranker')
 parser.add_argument('-qld',  action='store_true',
                     help='use qld ranker')
-parser.add_argument('-prf',  type=ClassifierType, nargs='+',
+parser.add_argument('-prcl',  type=ClassifierType, nargs='+', default=[],
                     help='use pseudo relevance feedback ranker')
```
lintool, line 37: pseudo relevance classifier rerank - the name of the class?

tests/test_indexutils.py
```diff
@@ -41,8 +41,9 @@ def setUp(self):
         tarball.extractall(self.index_dir)
         tarball.close()
 
-        self.searcher = pysearch.SimpleSearcher(f'{self.index_dir}lucene-index.cacm')
-        self.index_utils = pyutils.IndexReaderUtils(f'{self.index_dir}lucene-index.cacm')
+        self.index_path = os.path.join(self.index_dir, 'lucene-index.cacm')
```
lintool, line 44: this shouldn't be part of the commit?

pyserini/search/__main__.py
```diff
@@ -53,32 +53,39 @@
...
     searcher.set_qld()
 
+# invalid topics name
 if topics_dic == {}:
-    print('Topic Not Found')
+    print(f'Topic {args.topics} Not Found')
     exit()
 
+# get re-ranker
+need_classifier = args.prcl and len(args.prcl) > 0 and args.alpha > 0
```
lintool, line 62: doesn't seem like the write variable name? how about `use_prcl`?

pyserini/search/__main__.py
```diff
@@ -53,32 +53,39 @@
     search_rankers.append('qld')
     searcher.set_qld()
 
+# invalid topics name
```
lintool, line 56: how about just `topics` instead of `topics_dic`?

tests/test_indexutils.py
x65han: Will remove and add in another commit.

pyserini/search/__main__.py
x65han:
```
parser.add_argument('-indexpath', type=str, metavar='path',
                    required=True, help='the path to lucene index')
```

Changed my editor to line width of 100.

PR #149, opened 2020-05-29
Introduce new BM25Vectorizer

x65han:
- Introduce new BM25Vectorizer
- Refactor Common logic of vectorizer to `Class VectorizerBase`
- Add test case in `20newsgroups-replication.py` with f1 score assertion
- Replace `MultinomialNB` with `LogisticRegression`, which renders better results

----------------------------

tfidf vectorizer with Linear Regression model score: 0.8359058
bm25 vectorizer with Linear Regression model score: 0.8421606

----------------------------
```
python scripts/20newsgroups-replication.py --vectorizer TfidfVectorizer
python scripts/20newsgroups-replication.py --vectorizer BM25Vectorizer
```

### DIFF
--- a/pyserini/vectorizer.py
+++ b/pyserini/vectorizer.py
@@ -157,19 +157,20 @@ class BM25Vectorizer(Vectorizer):
             if index % 1000 == 0 and num_docs > 1000 and self.verbose:
                 print(f'Vectorizing: {index}/{len(docids)}')
 
+            # Term Frequency
             tf = self.index_utils.get_document_vector(doc_id)
+            if tf is None:
+                continue
 
             # Filter out in-eligible terms
             tf = {t: tf[t] for t in tf if t in self.term_to_index}
 
-            # Retrieve bm25 term weight
-            tf = {term: self.index_utils.compute_bm25_term_weight(doc_id, term, analyzer=None) for term in tf.keys()}
-
             # Convert from dict to sparse matrix
             for term in tf:
+                bm25_weight = self.index_utils.compute_bm25_term_weight(doc_id, term, analyzer=None)
                 matrix_row.append(index)
                 matrix_col.append(self.term_to_index[term])
-                matrix_data.append(tf[term])
+                matrix_data.append(bm25_weight)
 
         vectors = csr_matrix((matrix_data, (matrix_row, matrix_col)), shape=(num_docs, self.vocabulary_size))
         return self._l2normalize(vectors)
--- a/pyserini/vectorizer.py
+++ b/pyserini/vectorizer.py
@@ -23,6 +23,18 @@ from pyserini.index import pyutils
 
 
 class VectorizerBase:
+    """Base class for vectorizer implemented on top of Pyserini.
+
+    Parameters
+    ----------
+    lucene_index_path : str
+        Path to lucene index folder
+    min_df : int
+        Minimum acceptable document frequency
+    verbose : bool
+        Whether to print out debugging information
+    """
+
     def __init__(self, lucene_index_path: str, min_df: int = 1, verbose: bool = False) -> None:
         self.min_df: int = min_df
         self.verbose: bool = verbose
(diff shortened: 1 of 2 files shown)


pyserini/vectorizer.py
```diff
@@ -22,8 +22,8 @@
 from pyserini.index import pyutils
 
 
-class TfidfVectorizer:
-    """Wrapper class for tf-idf vectorizer implemented on top of Pyserini.
+class VectorizerBase:
```
lintool, line 25: Just `Vectorizer`? no base?

pyserini/vectorizer.py
```diff
@@ -36,23 +36,21 @@ class TfidfVectorizer:
     """
 
     def __init__(self, lucene_index_path: str, min_df: int = 1, verbose: bool = False) -> None:
```
lintool, line 38: I suppose `min_df` is still applicable for filtering the vocab, even if we're ultimately computing BM25. Is that the intention?

pyserini/vectorizer.py
```diff
@@ -103,4 +118,53 @@ def get_vectors(self, doc_ids: List[str]):
...
+    def __init__(self, lucene_index_path: str, min_df: int = 1, verbose: bool = False) -> None:
+        super().__init__(lucene_index_path, verbose, min_df)
+
+    def get_vectors(self, doc_ids: List[str]):
+        """Get the bm25 vectors given a list of doc_ids
+
+        Parameters
+        ----------
+        doc_ids : List[str]
+            The piece of text to analyze.
```
lintool, line 146: isn't this the list of docids?
btw, can you change `doc_ids` to `docids`? I use the latter consistently.

pyserini/vectorizer.py
```diff
@@ -103,4 +118,53 @@ def get_vectors(self, doc_ids: List[str]):
...
+
+        Returns
+        -------
+        csr_matrix
+            L2 normalized sparse matrix representation of bm25 vectors
```
lintool, line 151: capitalize BM25 in this context.

pyserini/vectorizer.py
x65han: Yes, with BM25Vectorizer we can still use `min_df` to filter out rare words when building our vocabulary.

scripts/20newsgroups-replication.py
```diff
@@ -19,43 +21,44 @@ def get_info(path):
...
-        'sci.space',
-        'soc.religion.christian',
-        'talk.politics.guns',
-        'talk.politics.mideast',
-        'talk.politics.misc',
-        'talk.religion.misc',
-    ]
+    parser = argparse.ArgumentParser(
+        description='Replication script of pyserini vectorizer')
+    parser.add_argument('-vectorizer', type=VectorizerType, required=True, help='which vectorizer to use')
```
lintool, line 49: `--vectorizer`? per what we agreed to? https://github.com/lintool/guide/blob/master/coding-style.md

scripts/20newsgroups-replication.py
```diff
@@ -19,43 +21,44 @@ def get_info(path):
...
     target_to_index = {t: i for i, t in enumerate(target_names)}
 
     train_docs, train_labels = get_info('./20newsgroups/20news-bydate-train/')
     test_docs, test_labels = get_info('./20newsgroups/20news-bydate-test/')
 
-    vectorizer = TfidfVectorizer(
-        './20newsgroups/lucene-index.20newsgroup.pos+docvectors+raw', min_df=5)
+    lucene_index_path = '20newsgroups/lucene-index.20newsgroup.pos+docvectors+raw'
+    vectorizer = None
+    if args.vectorizer == VectorizerType.TFIDF:
```
lintool, line 62: how about we just pass in the class name from the cmd line and init from string? e.g., https://stackoverflow.com/questions/4821104/dynamic-instantiation-from-string-name-of-a-class-in-dynamically-imported-module

that way we don't need this enum...

x65han: Interesting. I didn't know this is possible.

PR #157, opened 2020-06-02
synchronize data download instructions with anserini

stephaniewhoo:


docs/working-with-cord19.md
```diff
@@ -14,12 +14,11 @@ DATE=2020-05-26
 DATA_DIR=./collections/cord19-"${DATE}"
 mkdir "${DATA_DIR}"
 
-wget https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/historical_releases/cord-19_2020-05-26.tar.gz -P "${DATA_DIR}"
+wget https://ai2-semanticscholar-cord-19.s3-us-west-2.amazonaws.com/2020-05-26/document_parses.tar.gz -P "${DATA_DIR}"
```
lintool, line 17: Since you specify the date in a variable above, why not use `${DATE}`? That way in the future you only need to change one thing.

PR #161, opened 2020-06-02
Adopt sklearn's L2 norm function

x65han:
- Adopting sklearn's L2 norm function which is designed for [csr_matrix](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/preprocessing/_data.py)

### DIFF
--- a/pyserini/vectorizer.py
+++ b/pyserini/vectorizer.py
@@ -14,10 +14,11 @@
 
 import math
 import numpy as np
-from typing import List
-from scipy.sparse import csr_matrix
 from pyserini.search import pysearch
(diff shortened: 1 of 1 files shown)


pyserini/vectorizer.py
```diff
@@ -14,10 +14,11 @@
 
 import math
 import numpy as np
-from typing import List
-from scipy.sparse import csr_matrix
 from pyserini.search import pysearch
 from pyserini.index import pyutils
+from scipy.sparse import csr_matrix
```
x65han, line 19: alphabetical order

PR #162, opened 2020-06-03
Custom implementation of reciprocal_rank_fusion

x65han:
- add a custom implementation of reciprocal_rank_fusion, so we have control over the implementation details rather than relying on others.

### DIFF
--- a/pyserini/search/fusion.py
+++ b/pyserini/search/fusion.py
@@ -1,0 +1,82 @@
+# Pyserini: Python interface to the Anserini IR toolkit built on Lucene
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from functools import cmp_to_key
+import os
+import sys
+from typing import List, Dict
+
+
+class FusionSearcher:
+    """FusionSearcher class to perform fusion on multiple Qruns.
+
+    Parameters
+    ----------
+    paths : List[str]
+        A list of paths to Qruns
+    """
+
+    def __init__(self, paths: List[str]):
+        if len(paths) <= 1:
+            raise Exception('Fusion requres at least 2 files.')
+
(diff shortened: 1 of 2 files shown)


pyserini/search/fusion.py
```diff
@@ -0,0 +1,82 @@
...
+            self._load_qrun(path, k, rrf_by_topics)
+
+        if os.path.exists(output_path):
+            os.remove(output_path)
+
+        with open(output_path, 'w') as f:
+            for topic in rrf_by_topics:
+                rrf_by_topic = rrf_by_topics[topic]
+
+                for index, (doc_id, score) in enumerate(sorted(rrf_by_topic.items(), key=lambda x: x[1], reverse=True)):
```
lintool, line 80: Add in the _k_ value that trectools also implements?

lintool:
hi @x65han - I was thinking: `reciprocal_rank_fusion` should be a function that takes in multiple hits lists and returns a fused list.

`SimpleFusionSearcher` should be a class in `pyserini.search` that behaves like `SimpleSearcher` but takes multiple indexes.

So something like:

```python
searcher = SimpleFusionSearcher('/path/to/index1/', '/path/to/index2/', ..., method=rrf)

# Print the first 10 hits:
for i in range(0, 10):
    print(f'{i+1:2} {hits[i].docid:15} {hits[i].score:.5f}')
```

And internally, it uses: `reciprocal_rank_fusion`

pyserini/search/fusion.py
```diff
@@ -0,0 +1,82 @@
...
+                topic, _, doc_id, rank, _, _ = line.split()
+                if topic not in res:
+                    res[topic] = dict()
+
+                if doc_id not in res[topic]:
+                    res[topic][doc_id] = 0
+
+                res[topic][doc_id] += (1/(k+float(rank)))
+
+    def reciprocal_rank_fusion(self, output_path: str, k: int = 60, tag: str = None):
```
x65han, line 48: I have the k value passed over here. and used in line 46.

```
res[topic][doc_id] += (1/(k+float(rank)))
```

pyserini/search/fusion.py
```diff
@@ -0,0 +1,82 @@
...
+        with open(path, 'r') as f:
+            for line in f:
+                topic, _, doc_id, rank, _, _ = line.split()
+                if topic not in res:
+                    res[topic] = dict()
+
+                if doc_id not in res[topic]:
+                    res[topic][doc_id] = 0
+
+                res[topic][doc_id] += (1/(k+float(rank)))
```
rodrigonogueira4, line 46: nit: no need to convert to float in python 3

x65han:
Abandoning after code structure change.

Issue #119, opened 2020-05-20
Break README.md into document subpages?


lintool:
Main `README.md` is getting pretty long. Should we break into separate pages?

+ Usage of the Analyzer API
+ Usage of the Query Builder API
+ ...

@Chriskamphuis @PepijnBoers @x65han thoughts?

lintool:
cf. https://github.com/castorini/pyserini/pull/163#discussion_r434584234

@rodrigonogueira4 I assume you'd support this?

rodrigonogueira4:
Yes, sure

Chriskamphuis:
Sorry missed this earlier, I agree to split

PR #163, opened 2020-06-03
Massive module structure refactoring

lintool:
I've always thought that the pyserini module structure was a bit confusing, in that it broke the "flat is better than nested" principle:

```python
from pyserini.search import pysearch

searcher = pysearch.SimpleSearcher('indexes/index-robust04-20191213/')
```

Why `search` and then `pysearch`? Furthermore, the `py` prefix is odd to me. I think the original intention was that the `py` prefix indicates bindings to Java class... but that's an implementation detail that should really be hidden? To me, this is a lot more intuitive:

```python
from pyserini import search

searcher = search.SimpleSearcher('indexes/index-robust04-20191213/')
...
```

Or:

```python
from pyserini.search import SimpleSearcher

searcher = SimpleSearcher('indexes/index-robust04-20191213/')
...
```

Note that `search` is a subpackage (i.e., with `__init__.py`) as opposed to just a module, for future proofing, i.e., we can further define modules inside `search`, like `pyserini.search.querybuilder`.

This is a massive refactoring along these lines: basically, I "undo" the nesting by one level. Check out the new `README` to see how all the import statements change; nothing else really changes.

Thoughts? @PepijnBoers @Chriskamphuis @x65han @rodrigonogueira4 @ronakice


PepijnBoers:
Much cleaner! I’m in favour of these changes, and I think it will be even better with @Chriskamphuis his suggestion.

lintool:
@Chriskamphuis Do you like sklearn's  `_base.py` notation?
https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/ensemble

If we follow that, under search we'd have:

+ `_base.py` - `Document` class
+ `_searcher.py` - `SimpleSearcher` class (and we're going to implement `FusionSearcher` soon)
+ `_nearest_neighbor.py` - `SimpleNearestNeighborSearcher`

For the other subpackages (`analysis` etc.) we'd just have `__init__` and `_base`?

Chriskamphuis:
Yeah I like that better :)

lintool:
I've implement proposed changes for `pyserini.search`. I'll wait for others to comment before doing the rest.

I very much like modeling our library organization after `sklearn`.

README.md
```diff
@@ -217,10 +217,9 @@ For example, if a method expects the unanalyzed term and is called with an analy
 Initialize the class as follows:
 
 ```python
-from pyserini.index import pyutils
-from pyserini.analysis import pyanalysis
+from pyserini import analysis, index
 
-index_utils = pyutils.IndexReaderUtils('indexes/index-robust04-20191213/')
+index_utils = index.IndexReaderUtils('indexes/index-robust04-20191213/')
```
rodrigonogueira4, line 222: For brevity, should this be called `index.Reader()`? or `index.IndexReader()`

lintool: I like just `IndexReader`. I suppose I'll change it while I'm doing this massive refactoring.

lintool:
@rodrigonogueira4 but you like the general approach?

rodrigonogueira4:
> @rodrigonogueira4 but you like the general approach?

Definitely!

README.md
```diff
@@ -64,9 +64,9 @@ rm index-robust04-20191213.tar.gz
 Use the `SimpleSearcher` for searching:
 
 ```python
-from pyserini.search import pysearch
+from pyserini.search import SimpleSearcher
```
rodrigonogueira4, line 67: As for the readme.md examples, I think it should reflect more on what will be the most common use cases:
1- How to build an index
2- How to query the index
3- How to retrieve a document (contents, fields) given its id

The other examples (e.g. Usage of the Analyzer API) can be kept, but I think most users only care about those first 3.

lintool: Yes, will do this as part of #119 later?

rodrigonogueira4: SG!

lintool:
@PepijnBoers @Chriskamphuis please 👍 if you're happy also?

I'll wait for @x65han and @ronakice to see if they have any comments.

tests/test_analysis.py
```diff
@@ -19,10 +19,10 @@
 from random import randint
 from urllib.request import urlretrieve
 
-from pyserini.analysis import pyanalysis
-from pyserini.index import pyutils
+from pyserini import analysis
+from pyserini import index
+from pyserini import search
```
ronakice, line 24: Is there any reason why we have these imports separately?

tests/test_collection.py
```diff
@@ -19,32 +19,32 @@
 from random import randint
 from urllib.request import urlretrieve
 
-from pyserini.collection import pycollection
-from pyserini.index import pygenerator
+from pyserini import collection
+from pyserini import index
```
ronakice, line 23: same here

lintool:
`python -m unittest` passes but it's likely I broke other non-tested stuff. Will circle back and fix in later PR.

PR #165, opened 2020-06-03
Add fusion & trectools sub-modules

x65han:
- add `fusion` and `trectools` submodules.


pyserini/trectools/_base.py
```diff
@@ -0,0 +1,114 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from copy import deepcopy
+from typing import Dict, List, Set
+from functools import cmp_to_key
+
+
+class TrecRunDoc:
```
lintool, line 22: No sure why I se the need for both `TrecRun` and `TrecRunDoc`. Why not just add methods to `TrecRun` for reading from file and writing to file?

lintool: On I get it, it the intent that `TrecRunDoc` is a single line in a `TrecRun`? Not sure I like this design...

lintool:
@rodrigonogueira4 you should take a look also...

pyserini/fusion/_base.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from ..trectools import TrecRun, TrecRunDoc, sum_runs_by_score
+from typing import List
+
+
+def set_reciprocal_rank_fusion_score(doc: TrecRunDoc, k):
```
lintool, line 21: No sure I like two methods at top level here...

pyserini/trectools/_base.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if docid not in self.trec_run[topic]:
+            self.trec_run[topic][docid] = TrecRunDoc(topic=topic, docid=docid)
+
+    def add_score(self, topic: str, docid: str, delta: float) -> None:
+        self.initialize_doc_if_not_exist(topic, docid)
+        self.trec_run[topic][docid].score += delta
+
+
+def sum_runs_by_score(runs: List[TrecRun]):
```
lintool, line 101: Not sure I like have this method that isn't OO...

lintool:
As a suggestion, would it make sense to use a DataFrame for store a TREC run?

pyserini/trectools/_base.py
x65han: I can make it a static method in TrecRun.

pyserini/trectools/_base.py
lintool: After thinking about this a bit, I'm more of the opinion that DF is the way to go... you've basically implemented what amounts to a "Row" in a DF here?

x65han: Open to suggestions. `TrecRunDoc` is an abstraction to represent a document or each line in a TrecRun. We can print it as an object and it also takes care of the string serialization logic. 

It also works nicely to instantly create a placeholder or empty TrecRunDoc.

pyserini/fusion/_base.py
x65han: I can wrap it in a class? I wanted to keep it simple for now. As it becomes more complex, we can add more?

pyserini/fusion/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+
+import argparse
+from pyserini.fusion import reciprocal_rank_fusion
+from ..trectools import TrecRun
+
+parser = argparse.ArgumentParser(description='Create a input schema')
+parser.add_argument('--runs', type=str, nargs='+', default=[], required=True, help='A list of run files.')
+parser.add_argument('--output', type=str, required=True, help="Path to Lucene index.")
```
rodrigonogueira4, line 23: Does the description need to be changed?

pyserini/fusion/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+from pyserini.fusion import reciprocal_rank_fusion
+from ..trectools import TrecRun
+
+parser = argparse.ArgumentParser(description='Create a input schema')
+parser.add_argument('--runs', type=str, nargs='+', default=[], required=True, help='A list of run files.')
```
rodrigonogueira4, line 22: describe how runs should be separated?

x65han:
Changed my implementation to pandas dataframe. Requesting a review. Thanks!

tests/test_fusion.py
```diff
@@ -0,0 +1,40 @@
...
+
+class TestSearch(unittest.TestCase):
+    def setUp(self):
+        self.qruns = ['covidex.t5', 'covidex.sim']
+        self.cmps = ['myfused.txt', 'fused.txt']
+
+        for qrun in self.qruns:
+            os.system(f'wget -q -nc https://ir.nist.gov/covidSubmit/archive/round2/{qrun}')
+
+        os.system('wget -q -nc https://www.dropbox.com/s/scl0lm7x47jrsxh/fused.txt')
```
lintool, line 28: If you're going to use this testing strategy, how about use these?
https://github.com/castorini/anserini/blob/master/docs/experiments-covid.md#round-2

> reciprocal rank fusion(1, 3, 5)

pyserini/fusion/_base.py
```diff
@@ -0,0 +1,44 @@
...
+    fused_run : TrecRun
+        The fused TrecRun using reciprocal rank fusion.
+    """
+
+    if len(trec_runs) < 2:
+        raise Exception('Fusion requres at least 2 runs.')
+
+    fused_run = TrecRun()
+    for trec_run in trec_runs:
+        trec_run.set_rrf_scores(k)
```
lintool, line 40: if I understand your logic correctly, this is destructive - i.e., it messes with the original run... this probably will cause subtle bugs down the road... can we clone the df?

tests/test_fusion.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_reciprocal_rank_fusion(self):
+        os.system('python -m pyserini.fusion --runs covidex.sim covidex.t5 --output myfused.txt --tag reciprocal_rank_fusion_k=60')
+        self.assertTrue(filecmp.cmp(self.cmps[0], self.cmps[1]))
+
+    def tearDown(self):
```
lintool, line 34: Maybe a test case to read in a simple run file - just a few lines - to make sure loading functionality is working?

lintool: maybe create a new dir called `tests/resources` to store these test files?

pyserini/fusion/_base.py
x65han: Assuming you are referring to `set_rrf_scores`:

I will change the function name to `clone_with_rrf_scores(k)` so that it's clear we are cloning the TrecRun.

pyserini/trectools/_base.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, filepath: str = None):
+        self.run_data = dict()
+        self.filepath = filepath
+
+        if filepath is not None:
+            self.read_run(self.filepath)
+
+    def read_run(self, filepath: str) -> None:
+        run_data = pd.read_csv(filepath, sep="\s+", names=["topic", "q0", "docid", "rank", "score", "tag"])
```
lintool, line 38: changed to single quotes? we've standardized rest of code base on single quotes.

lintool:
You probably want to add `pandas` to requirements and also https://github.com/castorini/pyserini/blob/master/setup.py

tests/test_fusion.py
x65han: Will do

pyserini/trectools/_base.py
x65han: done

pyserini/trectools/_base.py
```diff
@@ -0,0 +1,105 @@
...
+
+        all_topics = all_topics.sort_values(by=['topic', 'score'], ascending=[True, False])
+        if tag is not None:
+            all_topics['tag'] = tag
+        all_topics.to_csv(output_path, sep=' ', header=False, index=False)
+
+    def get_docs_by_topic(self, topic: str):
+        return self.run_data[topic] if topic in self.run_data else None
+
+    def merge_runs_by_sum_scores(self, _run):
```
lintool, line 83: I like this, but should we change aggregation method to be controllable?
i.e., `aggregation='sum'` allowing for other options in the future?

tests/test_fusion.py
```diff
@@ -0,0 +1,47 @@
...
+from pyserini.trectools import TrecRun
+import unittest
+
+
+class TestSearch(unittest.TestCase):
+    def setUp(self):
+        self.qruns = ['tests/resources/covidex.t5', 'tests/resources/covidex.sim']
+        self.output_path = 'tests/resources/output.txt'
+
+    def test_reciprocal_rank_fusion(self):
```
lintool, line 28: I don't like checking large run files into the repo... bloats the repo, which I'd like to keep small.

I actually like your previous testing strategy better. How about we take this:
https://github.com/castorini/anserini/blob/master/docs/experiments-covid.md#round-2

Fuse rows 1, 3, 5 and make sure we get row 7?
Fuse rows 2, 4, 6 and make sure we get row 8?

I will ensure over the long term that the DropBox links are stable.

tests/test_fusion.py
```diff
@@ -0,0 +1,47 @@
...
+
+import filecmp
+import os
+from pyserini.trectools import TrecRun
+import unittest
+
+
+class TestSearch(unittest.TestCase):
+    def setUp(self):
+        self.qruns = ['tests/resources/covidex.t5', 'tests/resources/covidex.sim']
```
lintool, line 25: no need to put in `setUp` since you're not initializing state - you can move to each test case.

pyserini/fusion/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+#
+
+import argparse
+from pyserini.fusion import reciprocal_rank_fusion
+from ..trectools import TrecRun
+
+parser = argparse.ArgumentParser(description='Create a input schema')
+parser.add_argument('--runs', type=str, nargs='+', default=[], required=True, help='A list of run files separated by space.')
+parser.add_argument('--output', type=str, required=True, help="Path to resulting fused txt.")
+parser.add_argument('--tag', type=str, default="pyserini.fusion", help="Tag name of fused run.")
```
lintool, line 24: `--runtag`, to make consistent with anserini.

pyserini/trectools/_base.py
lintool: If you add `aggregation='avg'`, we can reimplement the `clprf` averaging technique here!

pyserini/fusion/_base.py
```diff
@@ -0,0 +1,43 @@
...
+    fused_run : TrecRun
+        The fused TrecRun using reciprocal rank fusion.
+    """
+
+    if len(trec_runs) < 2:
+        raise Exception('Fusion requres at least 2 runs.')
+
+    fused_run = TrecRun()
+    for trec_run in trec_runs:
+        fused_run.merge_runs_by_sum_scores(trec_run.clone_with_rrf_scores(k))
```
lintool, line 40: seems like `merge_runs` can be a static? So you can do something like:

```
fused_run = TrecRun.merge([trec_run.clone_with_rrf_scores(k) for trec_run in trec_runs])
```

lintool:
Okay, after thinking this through, I think I've come up with a clean API. How about something like:

```
run.clone().rescore(method='rrf',rrf.k=60)
```

I think the idea of rescoring a run is pretty general, because we could also have this:

```
run.clone().rescore(method='norm')
```

Which would normalize the scores to between 0 and 1.

Methods for combining multiple runs should be static, so we could do things like:

```
merged_run = TrecRun.merge(run1, run2, run3)
```

And we can implement interpolation later, e.g.:

```
interpolated_run = TrecRun.interpolate(run1, run2, 0.6)
```

This way, `clprf` and be cleanly rewritten in terms of these primitives.

Thus, RRF can be written as a one-liner:

```
fused_run = TrecRun.merge([run.clone().rescore(method='rrf') for run in runs])
```

Thoughts?

rodrigonogueira4:
> You probably want to add `pandas` to requirements and also https://github.com/castorini/pyserini/blob/master/setup.py



> Okay, after thinking this through, I think I've come up with a clean API. How about something like:
> 
> ```
> run.clone().rescore(method='rrf',rrf.k=60)
> ```
> 
> I think the idea of rescoring a run is pretty general, because we could also have this:
> 
> ```
> run.clone().rescore(method='norm')
> ```
> 
> Which would normalize the scores to between 0 and 1.
> 
> Methods for combining multiple runs should be static, so we could do things like:
> 
> ```
> merged_run = TrecRun.merge(run1, run2, run3)
> ```
> 
> And we can implement interpolation later, e.g.:
> 
> ```
> interpolated_run = TrecRun.interpolate(run1, run2, 0.6)
> ```
> 
> This way, `clprf` and be cleanly rewritten in terms of these primitives.
> 
> Thus, RRF can be written as a one-liner:
> 
> ```
> fused_run = TrecRun.merge([run.clone().rescore(method='rrf') for run in runs])
> ```
> 
> Thoughts?

I like it.

x65han:
> Okay, after thinking this through, I think I've come up with a clean API. How about something like:
> 
> ```
> run.clone().rescore(method='rrf',rrf.k=60)
> ```
> 
> I think the idea of rescoring a run is pretty general, because we could also have this:
> 
> ```
> run.clone().rescore(method='norm')
> ```
> 
> Which would normalize the scores to between 0 and 1.
> 
> Methods for combining multiple runs should be static, so we could do things like:
> 
> ```
> merged_run = TrecRun.merge(run1, run2, run3)
> ```
> 
> And we can implement interpolation later, e.g.:
> 
> ```
> interpolated_run = TrecRun.interpolate(run1, run2, 0.6)
> ```
> 
> This way, `clprf` and be cleanly rewritten in terms of these primitives.
> 
> Thus, RRF can be written as a one-liner:
> 
> ```
> fused_run = TrecRun.merge([run.clone().rescore(method='rrf') for run in runs])
> ```
> 
> Thoughts?

I like this! 👍   I'm going to merge this PR with a working solution, then address your comments above in the next PR. Thanks!

PR #167, opened 2020-06-04
Refactoring docs to break into subpages

### LINKED ISSUE #119: Break README.md into document subpages?

lintool:
Closes #119 - also adopts @rodrigonogueira4 's suggestion of what to focus documentation on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ doc = searcher.doc('LA071090-0047')
 
 From `doc`, you can access its `contents` as well as its `raw` representation.
 The `contents` hold the representation of what's actually indexed; the `raw` representation is usually the original "raw document".
-A simple example can illustrate this distinction: for an article from CORD-19, `raw` holds the complete JSON of the article, which obviously includes the artcle contents, but has metadata and other information as well.
+A simple example can illustrate this distinction: for an article from CORD-19, `raw` holds the complete JSON of the article, which obviously includes the article contents, but has metadata and other information as well.
 The `contents` are extracts from the article that's actually indexed (for example, the title and abstract).
 In most cases, `contents` can be deterministically reconstructed from the `raw`.
 When building the index, we specify flags to store `contents` and/or `raw`; it's rarely the case we store both, since it's usually a waste of space.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -112,267 +112,48 @@ for i in range(0, 10):
...
-index_reader = index.IndexReader('indexes/index-robust04-20191213/')
+doc = searcher.doc('LA071090-0047')
 ```
 
-Use `terms()` to grab an iterator over all terms in the collection, i.e., the dictionary.
-Note that these terms are _analyzed_.
-Here, we only print out the first 10:
+From `doc`, you can access its `contents` as well as its `raw` representation.
+The `contents` hold the representation of what's actually indexed; the `raw` representation is usually the original "raw document".
+A simple example can illustrate this distinction: for an article from CORD-19, `raw` holds the complete JSON of the article, which obviously includes the artcle contents, but has metadata and other information as well.
```
PepijnBoers, line 124: artcle -> article

docs/usage-collection.md
```diff
@@ -0,0 +1,28 @@
...
+collection = collection.Collection('HtmlCollection', 'collections/cacm/')
+generator = index.Generator('DefaultLuceneDocumentGenerator')
+
+for (i, fs) in enumerate(collection):
+    for (j, doc) in enumerate(fs):
+        parsed = generator.create_document(doc)
+        docid = parsed.get('id')            # FIELD_ID
+        raw = parsed.get('raw')             # FIELD_RAW
+        contents = parsed.get('contents')   # FIELD_BODY
+        print('{} {} -> {} {}...'.format(i, j, docid, contents.strip().replace('\n', ' ')[:50]))
```
PepijnBoers, line 27: Use f-strings for formatting?

README.md
lintool: good catch!

docs/usage-collection.md
lintool: I tried, but according to: https://www.python.org/dev/peps/pep-0498/#escape-sequences

> Backslashes may not appear anywhere within expressions.

PepijnBoers: Ah okay I didn't know that!

PR #171, opened 2020-06-06
Fusion searcher

x65han:
# Todo

- We need a test case with more than one Lucene index. @lintool any suggestions?
- Doc str after the implementation is finalized.

# Sample Usage

```
from pyserini.search import SimpleFusionSearcher
from pyserini.trectools import FusionMethod


index_dirs = [
    '../anserini/indexes/lucene-index.core17.pos+docvectors+raw',
    '../anserini/indexes/lucene-index.core18.pos+docvectors+raw',
    '../anserini/indexes/lucene-index.robust04.pos+docvectors+raw',
    '../anserini/indexes/lucene-index.robust05.pos+docvectors+raw'
]

searcher = SimpleFusionSearcher(index_dirs, method=FusionMethod.RRF)

for i in range(0, 10):
    print(f'{i+1:2} {hits[i].docid:15} {hits[i].score:.5f}')

```

### DIFF
--- a/pyserini/search/__main__.py
+++ b/pyserini/search/__main__.py
@@ -15,13 +15,11 @@
 #
 
 import argparse
-
-from tqdm import tqdm
-
 from pyserini.search import get_topics, SimpleSearcher
 from pyserini.search.reranker import ClassifierType, PseudoRelevanceClassifierReranker
+from tqdm import tqdm
 
-parser = argparse.ArgumentParser(description='Create a input schema')
+parser = argparse.ArgumentParser(description='Perform search on a given lucene index with various options.')
 parser.add_argument('--index', type=str, metavar='path to index', required=True, help="Path to Lucene index.")
 parser.add_argument('--topics', type=str, metavar='topic_name', required=True,
                     help="Name of topics. Available: robust04, robust05, core17, core18.")
--- a/pyserini/search/__main__.py
+++ b/pyserini/search/__main__.py
@@ -19,7 +19,7 @@ from pyserini.search import get_topics, SimpleSearcher
 from pyserini.search.reranker import ClassifierType, PseudoRelevanceClassifierReranker
 from tqdm import tqdm
 
-parser = argparse.ArgumentParser(description='Perform search on a given lucene index with various options.')
+parser = argparse.ArgumentParser(description='Search a Lucene index.')
 parser.add_argument('--index', type=str, metavar='path to index', required=True, help="Path to Lucene index.")
 parser.add_argument('--topics', type=str, metavar='topic_name', required=True,
                     help="Name of topics. Available: robust04, robust05, core17, core18.")
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -258,47 +258,35 @@ class SimpleFusionSearcher:
         self.method = method
         self.searchers = [SimpleSearcher(index_dir) for index_dir in index_dirs]
 
-    def search(self, q: Union[str, JQuery], k: int = 10, query_generator: JQueryGenerator = None) -> TrecRun:
+    def get_searchers(self) -> List[SimpleSearcher]:
+        return self.searchers
+
+    def search(self, q: Union[str, JQuery], k: int = 10, query_generator: JQueryGenerator = None) -> List[JSimpleSearcherResult]:
         trec_runs = []
-        docid_to_search_result = dict()
+        docid_to_score, docid_to_search_result = dict(), dict()
 
         for searcher in self.searchers:
-            hits = searcher.search(q, k, query_generator)
+            hits = searcher.search(q, k=k, query_generator=query_generator)
             for hit in hits:
                 docid_to_search_result[hit.docid] = hit
-            run = TrecRun.generate_trec_run_from_search_results(hits)
+                docid_to_score[hit.docid] = hit.score
+            run = TrecRun.from_search_results(docid_to_score)
             trec_runs.append(run)
 
         if self.method == FusionMethod.RRF:
-            fused_run = reciprocal_rank_fusion(trec_runs, rrf_k=60, depth=1000, k=1000)
+            fused_run = reciprocal_rank_fusion(trec_runs, rrf_k=60, depth=1000, k=k)
         else:
             raise NotImplementedError()
 
-        return fused_run.map_trec_run_to_simple_search_result(docid_to_search_result)
+        return SimpleFusionSearcher.convert_to_search_result(fused_run, docid_to_search_result)
 
-    def set_analyzer(self, analyzer):
-        for searcher in self.searchers:
-            searcher.set_analyzer(analyzer)
+    @staticmethod
+    def convert_to_search_result(run: TrecRun, docid_to_search_result: Dict[str, JSimpleSearcherResult]) -> List[JSimpleSearcherResult]:
+        search_results = []
 
-    def set_rm3(self, fb_terms=10, fb_docs=10, original_query_weight=float(0.5), rm3_output_query=False):
-        for searcher in self.searchers:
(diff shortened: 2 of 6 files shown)


pyserini/search/_searcher.py
```diff
@@ -248,4 +250,55 @@ def bm25(k1=0.9, b=0.4):
...
+    def __init__(self, index_dirs: List[str], method: FusionMethod):
+        self.method = method
+        self.searchers = [SimpleSearcher(index_dir) for index_dir in index_dirs]
+
+    def search(self, q: Union[str, JQuery], k: int = 10, query_generator: JQueryGenerator = None) -> TrecRun:
+        trec_runs = []
+        docid_to_search_result = dict()
+
+        for searcher in self.searchers:
+            hits = searcher.search(q, k, query_generator)
```
lintool, line 266: I think this _k_ is actually the depth, right?

pyserini/search/_searcher.py
```diff
@@ -248,4 +250,55 @@ def bm25(k1=0.9, b=0.4):
...
+            for hit in hits:
+                docid_to_search_result[hit.docid] = hit
+            run = TrecRun.generate_trec_run_from_search_results(hits)
+            trec_runs.append(run)
+
+        if self.method == FusionMethod.RRF:
+            fused_run = reciprocal_rank_fusion(trec_runs, rrf_k=60, depth=1000, k=1000)
```
lintool, line 273: and this _k_ is the _k_ passed in from the method call?

pyserini/trectools/_base.py
```diff
@@ -90,36 +116,45 @@ def get_all_topics_from_runs(runs) -> Set[str]:
...
         res = TrecRun() if run is None else run
 
         df = pd.DataFrame(rows)
         df.columns = TrecRun.columns
         res.run_data = df.copy()
 
         return res
+
+    @staticmethod
+    def generate_trec_run_from_search_results(hits: List[JSimpleSearcherResult]) -> TrecRun:
```
lintool, line 154: how about just `from_search_results`? The fact that it's a static of `TrecRun` makes the return type obvious?

pyserini/search/__main__.py
```diff
@@ -15,13 +15,11 @@
...
 import argparse
-
-from tqdm import tqdm
-
 from pyserini.search import get_topics, SimpleSearcher
 from pyserini.search.reranker import ClassifierType, PseudoRelevanceClassifierReranker
+from tqdm import tqdm
 
-parser = argparse.ArgumentParser(description='Create a input schema')
+parser = argparse.ArgumentParser(description='Perform search on a given lucene index with various options.')
```
lintool, line 22: "Search a Lucene index" - omit needless words:
+ "perform" not necessary
+ "give"... well, of course it's given
+ "options"... well, of course there are various options

pyserini/search/_searcher.py
```diff
@@ -248,4 +250,55 @@ def bm25(k1=0.9, b=0.4):
...
\ No newline at end of file
+        return autoclass('org.apache.lucene.search.similarities.LMDirichletSimilarity')(mu)
+
+
+class SimpleFusionSearcher:
+    def __init__(self, index_dirs: List[str], method: FusionMethod):
+        self.method = method
+        self.searchers = [SimpleSearcher(index_dir) for index_dir in index_dirs]
+
+    def search(self, q: Union[str, JQuery], k: int = 10, query_generator: JQueryGenerator = None) -> TrecRun:
```
lintool, line 261: Return type not accurate?

pyserini/search/_searcher.py
```diff
@@ -248,4 +250,55 @@ def bm25(k1=0.9, b=0.4):
...
+            trec_runs.append(run)
+
+        if self.method == FusionMethod.RRF:
+            fused_run = reciprocal_rank_fusion(trec_runs, rrf_k=60, depth=1000, k=1000)
+        else:
+            raise NotImplementedError()
+
+        return fused_run.map_trec_run_to_simple_search_result(docid_to_search_result)
+
+    def set_analyzer(self, analyzer):
```
lintool, line 279: add `get_searchers` method? Returns list of individual searchers.

pyserini/search/_searcher.py
```diff
@@ -248,4 +250,55 @@ def bm25(k1=0.9, b=0.4):
...
+        for searcher in self.searchers:
+            searcher.set_analyzer(analyzer)
+
+    def set_rm3(self, fb_terms=10, fb_docs=10, original_query_weight=float(0.5), rm3_output_query=False):
```
lintool, line 283: Remove all of these methods? Users can call `get_searchers` and configure individual searcher themselves?

pyserini/trectools/_base.py
```diff
@@ -14,9 +14,25 @@
...
+
 from copy import deepcopy
+from enum import Enum
 import pandas as pd
-from typing import Set
+from pyserini.pyclass import autoclass
+from typing import Dict, List, Set
+
+# Wrappers around Anserini classes
+JSimpleSearcherResult = autoclass('io.anserini.search.SimpleSearcher$Result')
```
lintool, line 26: We've created this in search already; just import, otherwise you'll get weird variable shadowing effects...

pyserini/trectools/_base.py
```diff
@@ -70,16 +86,26 @@ def get_docs_by_topic(self, topic: str, max_docs: int = None):
...
             for topic, _, docid, rank, _, tag in self.run_data.to_numpy():
                 rows.append((topic, 'Q0', docid, rank, 1 / (rrf_k + rank), tag))
 
-            return TrecRun.get_trec_run_by_list(rows, self)
+            return TrecRun.get_trec_run_from_list(rows, self)
         else:
-            raise Exception(f'Unknown rescore method {method} detected')
+            raise NotImplementedError()
+
+    def map_trec_run_to_simple_search_result(self, docid_to_search_result: Dict[str, JSimpleSearcherResult]) -> List[JSimpleSearcherResult]:
```
lintool, line 100: map -> convert: save map for the technical functional programming sense.

just `convert_to_search_results`? `trec_run` is implict give it's a method?

lintool: On second thought, this method should be directly inside the fusion searcher... because we're passing in `Dict[str, JSimpleSearcherResult]`, breaks separation of concerns since `TrecRun` doesn't deal with a result object anywhere else?

pyserini/trectools/_base.py
```diff
@@ -90,36 +116,45 @@ def get_all_topics_from_runs(runs) -> Set[str]:
...
         else:
-            raise Exception(f'Unknown aggregation method {aggregation} detected.')
+            raise NotImplementedError()
 
-        return TrecRun.get_trec_run_by_list(rows)
+        return TrecRun.get_trec_run_from_list(rows)
 
     @staticmethod
-    def get_trec_run_by_list(rows, run=None):
+    def get_trec_run_from_list(rows, run: TrecRun = None) -> TrecRun:
```
lintool, line 144: what type is `rows`? not immediately obvious what this method is doing?

pyserini/search/_searcher.py
x65han: Oh, sounds good!

pyserini/trectools/_base.py
x65han: slow. The faster way is to dump df to a Numpy array, perform all the necessary operations, and append results to a list.

```rows.append((topic, 'Q0', docid, rank, 1 / (rrf_k + rank), tag))```


With the variable `rows`, we can this function to convert it to a TrecRun.

If `run` is given, then we update the given TrecRun with `rows`.

x65han: Added proper doc string to explain what this function is doing.

PR #173, opened 2020-06-06
Citation graph script

HangCui0510:


scripts/cord19/citation_graph.py
```diff
@@ -0,0 +1,45 @@
+import pyserini.collection
```
lintool, line 1: Add in standard boilerplate.

scripts/cord19/citation_graph.py
```diff
@@ -0,0 +1,45 @@
...
+import json
+import csv
+collection = pyserini.collection.Collection(
```
lintool, line 4: we use 120 char width, so it can fit on a line.

scripts/cord19/citation_graph.py
```diff
@@ -0,0 +1,45 @@
...
+    'Cord19AbstractCollection', 'collections/cord19')
```
lintool, line 5: use argparse - make path a variable on the command line

scripts/cord19/citation_graph.py
```diff
@@ -0,0 +1,45 @@
...
+articles = collection.__next__()
+article = None
+with open("node.csv", 'w') as file1:
```
lintool, line 8: make these args on the command line, with defaults.

scripts/cord19/citation_graph.py
```diff
@@ -0,0 +1,45 @@
...
+    writer = csv.writer(file1)
+    writer.writerow(['Id'])
+with open("edge.csv", 'w') as file2:
+    writer = csv.writer(file2)
+    writer.writerow(['Source', 'Target'])
+
+for (i, d) in enumerate(articles):
+    article = pyserini.collection.Cord19Article(d.raw)
+    # if i == 100:     #change this number
```
lintool, line 17: what does this mean?

lintool:
We should map node ids back to `cord_uid` so we can tell what paper in the collection the corresponds to each vertex.

We'll need to think about references to articles outside the collection... maybe generate unique ids for them?

scripts/cord19/citation_graph.py
HangCui0510: In case we don't want to run the entire dataset. I delete this in new commit.

scripts/cord19/citation_graph.py
```diff
@@ -26,48 +26,54 @@ def dir_path(path):
...
-collection = pyserini.collection.Collection('Cord19AbstractCollection', args.path)
-articles = collection.__next__()
-article = None
-with open(args.node_file, 'w') as file1:
-    writer = csv.writer(file1)
-    writer.writerow(['Id'])
-with open(args.edge_file, 'w') as file2:
-    writer = csv.writer(file2)
-    writer.writerow(['Source', 'Target'])
+def parse_arguments():
```
lintool, line 29: typically, this goes directly in `__name__`

lintool:
rename to `extract_citation_graph.py` please.

PR #176, opened 2020-06-08
Script and instructions to load CORD-19 citation data into Neo4j 

mrkarezina:
The script prepares the csv files that the cypher queries use to load the citation data into Neo4j.

Originally used `py2neo` to load the articles directly from the script but that was very slow.

Currently the only reliable citation data is titles. The script initially creates nodes for the articles in the dataset, and then merges cited articles (creates new nodes if the title is new). This leaves the possibility of there being duplicate titles that have the same node. We could try to use citation metadata to separate these but the additional citation metadata is unreliable.

A quick script shows there's a total of 59839 articles and 57099 after removing duplicate titles.


docs/working-with-cord19.md
```diff
@@ -85,3 +85,52 @@ article.json
...
+The easiest way to get started with Neo4j and start an instance is to [download the Neo4j desktop](https://neo4j.com/download-center/).
+
+In the desktop app create a new project and add a database (create local graph). Give it any name and password and use Neo4j version `4.0.4`. The click "start" to start running database locally.
+
+Once you have the [Pyserini development environment](https://github.com/castorini/pyserini#development-installation) setup run the `neo4j_loader.py` script in the root of the pyserini project.
+
+Due to security reasons Neo4j only allows Cypher queries to acess files in certain directories. Move the generated csv files, `articles.csv` and `edges.csv` to the import directory of Neo4j. Follow [this guide](https://neo4j.com/docs/operations-manual/current/configuration/file-locations/) to find the import directory on your machine.
+
+
+### Cypher Queries to load the data
```
lintool, line 101: don't need another section heading here since it's a singleton.

scripts/cord19/neo4j_loader.py
```diff
@@ -0,0 +1,43 @@
+# Pyserini: Python interface to the Anserini IR toolkit built on Lucene
```
lintool, line 1: rename into a verb phrase, e.g., `load_citation_graph_into_neo4j` for clarity.

I like scripts that have `main` to be verb phrases.

scripts/cord19/neo4j_loader.py
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pyserini.collection
+import csv
+
+collection = pyserini.collection.Collection('Cord19AbstractCollection', 'collections/cord19-2020-05-26')
```
lintool, line 18: make path be an arg from the command line.

PR #178, opened 2020-06-09
Introduce new fusion method by average

x65han:
- introduce new fusion method by averaging
- include test case

### DIFF
--- a/pyserini/fusion/__init__.py
+++ b/pyserini/fusion/__init__.py
@@ -16,4 +16,4 @@
 
 from ._base import average, FusionMethod, interpolation, reciprocal_rank_fusion
 
-__all__ = ['average', 'FusionMethod', 'interpolation', 'reciprocal_rank_fusion']
+__all__ = ['FusionMethod', 'average', 'interpolation', 'reciprocal_rank_fusion']
--- a/pyserini/fusion/__init__.py
+++ b/pyserini/fusion/__init__.py
@@ -14,6 +14,6 @@
 # limitations under the License.
 #
 
-from ._base import FusionMethod, interpolation, reciprocal_rank_fusion
+from ._base import average, FusionMethod, interpolation, reciprocal_rank_fusion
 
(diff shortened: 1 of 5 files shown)


pyserini/fusion/__init__.py
```diff
@@ -14,6 +14,6 @@
 # limitations under the License.
 #
 
-from ._base import FusionMethod, interpolation, reciprocal_rank_fusion
+from ._base import average, FusionMethod, interpolation, reciprocal_rank_fusion
 
-__all__ = ['FusionMethod', 'interpolation', 'reciprocal_rank_fusion']
+__all__ = ['average', 'FusionMethod', 'interpolation', 'reciprocal_rank_fusion']
```
lintool, line 19: can you look up to see if there are best practices on how to order these? alphabetical mixed case seems a bit odd to me...

pyserini/fusion/_base.py
```diff
@@ -84,3 +85,28 @@ def interpolation(runs: List[TrecRun], alpha: int = 0.5, depth: int = None, k: i
     scaled_runs.append(runs[1].clone().rescore(method=RescoreMethod.SCALE, scale=(1-alpha)))
 
     return TrecRun.merge(scaled_runs, AggregationMethod.SUM, depth=depth, k=k)
+
+
+def average(runs: List[TrecRun], depth: int = None, k: int = None):
+    """Perform fusion by averaging on a list of ``TrecRun`` objects.
+    new_score = sum(scores) / num_runs
```
lintool, line 92: I don't think you need to explain how to compute average... :)

x65han: Hahaha. Will remove.

pyserini/fusion/__init__.py
x65han: I didn't find anything specific on pep8 style guide. But it seems like we should put classes first, e.g. https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/svm/__init__.py

PR #194, opened 2020-06-12
Add MS MARCO passage retrieval replication in pure Python

lintool:
I think this is going to make a lot of people happy!

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -110,7 +110,8 @@ map                   	all	0.1957
 recall_1000           	all	0.8573
 ```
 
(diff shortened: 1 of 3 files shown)


docs/experiments-msmarco-passage.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+The output should be:
+
+```
+map                   	all	0.1957
+recall_1000           	all	0.8573
+```
+
+Average precision and recall@1000 are the two metrics we care about the most.
```
ronakice, line 113: Maybe an explanation of why we care about these instead of MRR @ 10 which is their official metric? (i.e. something about multi-stage ranking)

lintool: Will do!

PR #196, opened 2020-06-12
Replace verify_simplesearcher.py with integrations

lintool:
There's code duplication... do you think it's worth creating a generic class of the test itself?


x65han:
Maybe we can merge it for now since it's a working solution. In this step, we are changing it from a standalone script to integration testing. Then in the next step, we can think about how to create a generic class.

PR #199, opened 2020-06-15
Rewrite Neo4j script and queries for better load time

mrkarezina:
Rewrite the citation graph script to make it significantly faster by making better use of the Cypher queries. Instead of checking for replication in the script and creating an `articles.csv` containing CORD-19 and citation title, keep `articles.csv` only for CORD-19 titles.

In the `edges.csv` file store the source title and target title with all the available metadata. Only some citations have DOIs, by merging on the title in the query we can combine metadata from multiple citations.

Create a unique constraint on titles. There are some accidental duplicates in the dataset (possibly some actual duplicate titles) but if we create a unique constraint on titles the load speed is magnitudes better.

Loading everything should now take about 10min.


lintool:
@mrkarezina you have merge conflicts to resolve... I'll let @HangCui0510 take a look first?

mrkarezina:
I think that was because of the minor issue PR approved in the meantime. Merge conflict should be resolved now.

HangCui0510:
@lintool Yes, this is good. It works fine with neoj4 because you can manage a lot of things using cypher queries. For example, the output 'nodes.csv' file from the scripts contains full-text articles and 'edge.csv' file contains the bib_entries. Using cypher queries, @mrkarezina creates two sets of nodes and makes connections between them. This can not be done in Gephi, I have to store all nodes in one file. That's why my original scripts were slow. However, we can stream from neoj4 to Gephi anyways so I think marko's scripts are good.

PR #200, opened 2020-06-17
Update MS MARCO documentation

lintool:
+ Update passage retrieval task
+ Add doc retrieval task


scripts/cord19/find_cord19_length_outlier.py
```diff
@@ -19,7 +19,7 @@
```
rodrigonogueira4, line 19: Maybe add a description of what this script does?

lintool: Yup, good call. This script was written by an URA. I'll ask for improvement on next iteration.

PR #210, opened 2020-06-20
TrecRun.remove/keep docs by qrels

x65han:
# Add two functions to TrecRun

- keep docs by qrels so that we get 100% judged
- remove docs by qrels so e.g. we don't want rnd1 docs when submitting round 2

### DIFF
--- a/pyserini/trectools/_base.py
+++ b/pyserini/trectools/_base.py
@@ -30,6 +30,56 @@ class RescoreMethod(Enum):
     SCALE = 'scale'
 
 
+class Qrels:
+    """Wrapper class for TREC Qrels.
+
+    Parameters
+    ----------
+    filepath : str
+        File path of a given TREC Qrels.
+    """
+
+    columns = ['topic', 'q0', 'docid', 'relevance_grade']
+
+    def __init__(self, filepath: str = None):
+        self.filepath = filepath
+        self.qrels_data = pd.DataFrame(columns=Qrels.columns)
+
+        if filepath is not None:
+            self.read_run(self.filepath)
+
+    def read_run(self, filepath: str):
+        self.qrels_data = pd.read_csv(filepath, sep='\s+', names=Qrels.columns)
+
+    def get_relevance_grades(self) -> Set[str]:
+        """Return a set with all relevance grades."""
+
+        return set(sorted(self.qrels_data["relevance_grade"].unique()))
+
+    def topics(self) -> Set[str]:
+        """Return a set with all topics."""
+
+        return set(sorted(self.qrels_data["topic"].unique()))
+
+    def get_docids(self, topic, relevance_grades=None) -> List[str]:
+        """"Return a list of docids for a given topic and a list relevance grades.
+
+        Parameters:
+        ----------
+        relevance : List[int]
+            E.g. [0, 1, 2]. If not provided, then all relevance will be returned.
+        topic : int
+        """
+
+        if relevance_grades is None:
+            relevance_grades = self.get_relevance_grades()
+
+        filtered_df = self.qrels_data[self.qrels_data['topic'] == topic]
+        filtered_df = filtered_df[filtered_df['relevance_grade'].isin(relevance_grades)]
+
+        return filtered_df['docid'].tolist()
+
+
 class TrecRun:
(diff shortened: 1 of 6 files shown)


lintool:
hey @x65han is this meant to be a replacement for these scripts?

x65han:
No replacement is happening. I am moving the code around a little bit.


1. I moved Qrels to be before TrecRun, but still in the same file. This way, we can refer to the Qrels type in TrecRun.

2. I also moved a couple test cases from test_fusion.py to test_trectools.py since the two test cases are testing Qrels & TrecRun class, which belong to the trectools submodules.

lintool:
Sorry, it appears that I missed the URL: https://github.com/castorini/anserini-tools/blob/master/scripts/filter_run_with_qrels.py

It seems like the filter qrels portion is a cleaner, more integration version of the above script?

pyserini/trectools/_base.py
```diff
@@ -99,6 +152,71 @@ def rescore(self, method: RescoreMethod, rrf_k: int = None, scale: float = None)
     def to_numpy(self) -> np.ndarray:
         return self.run_data.to_numpy(copy=True)
 
+    def remove_by_qrels(self, qrels: Qrels, copy=True):
```
lintool, line 155: how about `discard_qrels` and `retain_qrels` to be consistent with https://github.com/castorini/anserini-tools/blob/master/scripts/filter_run_with_qrels.py ?

pyserini/trectools/_base.py
```diff
@@ -99,6 +152,71 @@ def rescore(self, method: RescoreMethod, rrf_k: int = None, scale: float = None)
...
+        """Given a Qrels object, remove each docid in self if docid is also in the qrels.
+        This operation is performed on each topic separately.
```
lintool, line 157: I think docstring convention is to begin description with a verb, so something like "Discard docids that are present in a Qrels object..."

pyserini/trectools/_base.py
```diff
@@ -99,6 +152,71 @@ def rescore(self, method: RescoreMethod, rrf_k: int = None, scale: float = None)
...
+
+        Parameters:
+        ----------
+        qrels : Qrels
+            Qrels with docids to remove from TrecRun.
+        copy : Bool
```
lintool, line 163: I think `clone` is more standard?

pyserini/trectools/_base.py
```diff
@@ -99,6 +152,71 @@ def rescore(self, method: RescoreMethod, rrf_k: int = None, scale: float = None)
...
+
+    def _filter_from_qrels(self, qrels: Qrels, keep: bool, copy=True):
+        """Private helper function to remove/keep each docid in self if docid is also in the given Qrels object.
+        This operation is performed on each topic separately.
+
+        Parameters:
+        ----------
+        qrels : Qrels
+            Qrels with docids to remove from or keep in TrecRun.
+        copy : Bool
```
lintool, line 192: typo here? two "copy"

pyserini/trectools/_base.py
```diff
@@ -99,6 +152,71 @@ def rescore(self, method: RescoreMethod, rrf_k: int = None, scale: float = None)
...
+            Return a new TrecRun object if True, else self will be modified and returned.
+        """
+
+        df_list = []
+        for topic in self.topics():
+            if topic not in qrels.topics():
+                continue
+
+            tabu_list = qrels.get_docids(topic)
```
lintool, line 203: what's "tabu" mean?

x65han: Coming from [Tabu Search](https://www.sciencedirect.com/topics/computer-science/tabu-search#:~:text=2.2%20Tabu%20Search%20Algorithm%20(TS,composed%20by%20the%20next%20elements.), tabu just means something we don't want. Maybe not the best naming here. I am going to name it `qrels_docids`.

x65han:
> Sorry, it appears that I missed the URL: https://github.com/castorini/anserini-tools/blob/master/scripts/filter_run_with_qrels.py
> 
> It seems like the filter qrels portion is a cleaner, more integration version of the above script?

Yes, integrating the logic of `filter_run_with_qrels.py` into our TrecRun and Qrels object.

PR #223, opened 2020-07-27
Add get position list function

nsndimt:
expose [getTermPositions](https://github.com/castorini/anserini/pull/1337) in the IndexReader class. Aside from the term positions mapping, this function also returns a reconstructed document using the mapping. This should help users to the effect of stop words removal and stemming. Currently, the document returned by doc.contents() is the document before it is processed by Lucene. The reconstructed document will provide a view of the document after Lucene's processing.


docs/usage-indexreader.md
```diff
@@ -72,6 +72,15 @@ print(doc_vector)
 ```
 
 The result is a dictionary where the keys are the analyzed terms and the values are the term frequencies.
+
+If you want to know the positions of each term in the document, you can use `get_term_positions`:
```
lintool, line 76: "of each term" -> "of every term"?

docs/usage-indexreader.md
```diff
@@ -72,6 +72,15 @@ print(doc_vector)
...
+```python
+term_positions, indexed_doc = index_reader.get_term_positions('FBIS4-67701')
+print(term_positions)
+print(indexed_doc)
+```
+The result is a tuple. The first member is a dictionary where the keys are the analyzed terms and the values are the positions each term occur in the document. The second member is a string containing the recovered document content using the position information.
```
lintool, line 82: Here I think you can just write `Tuple[Dict[str, int], str]` - a Python programmer should be able to interpret type signatures, and it's more concise.

pyserini/index/_base.py
```diff
@@ -260,6 +260,35 @@ def get_document_vector(self, docid: str) -> Optional[Dict[str, int]]:
...
+        Parameters
+        ----------
+        docid : str
+            Collection ``docid``.
+
+        Returns
+        -------
+        Optional[Tuple[Dict[str, int], str]]
+            A tuple contains a dictionary with analyzed terms as keys and corresponding posting list as values, and a
+            string representing the recovered document
```
lintool, line 278: Why do you want to return "string representing the recovered document"? What's the use case? Users can easily do this also if they want?

nsndimt: Currently, the document returned by `doc.contents()` is the document before it is processed by Lucene. The reconstructed document will provide a view of the document after Lucene's processing. 
Yes. Users can easily do this if they want, and not every user needs this. I think we can move this piece of code to `usage-indexreader.md` as an example of how to use this function.

lintool: Agreed - can you do that?

nsndimt: done

Issue #226, opened 2020-08-05
Baseline LTR in pyserini


lintool:
Discussion with @nsndimt on building LTR in Pyserini. I think we're in agreement:

+ Use Python as primary language
+ Features extractors in Java
+ Python for "gluing" everything together and interfacing with different ML toolkits

Our initial target, MS MARCO passage, two feature LTR pipeline:
+ f1 = BM25 score (already implemented)
+ f2 = something simple, like fraction of matching query terms

So Python constructs something like:

```
judgment_for_doc1 f1 f2
judgment_for_doc2 f1 f2
judgment_for_doc3 f1 f2
...
```

Could be a DataFrame, could be CSV, could be svmlight format... whatever is reasonable - and we run the pipeline end to end.

@nsndimt thoughts?

lintool:
We have #321 and #438, so closing.

PR #227, opened 2020-08-19
New Learning to Rank module

nsndimt:
As discussed in #226, I expose `FeatureExtractorUtils` in Pyserini. I create a baseline model using BM25 score and query length as features. When evaluated on 500 randomly sampled queries from the dev set of MS MARCO dataset, the baseline model's MRR reaches 0.185 using 1% of queries in the training set.


lintool:
Currently, each feature extractor has method signature `extract(queryTokens, docIds)`: we should discuss whether that's the right design.

For example, what if there are annotations on the query - e.g., we run NER on it? We only want to run NER once...

What about annotations on the text? If we run NER on it, and want to use the NER output as the basis for multiple features? We only want to run NER once...

What if both the query and the document were JSON structures? And we could add arbitrary annotations to it? Is that too unconstrained?

stephaniewhoo:
closed for new [PR](https://github.com/castorini/pyserini/pull/438) raised for LTR

PR #230, opened 2020-09-19
Add automate downloading of indexes

qguo96:
Currently, this change supports 'ms-marco-passage', 'ms-marco-doc' and 'TREC Disks 4 & 5'.

- If the index exists, skip the download and use the index under '(pyserini)/indexes'.
- If not, download the index to cache(`~/.cache/pyserini/indexes`) and extract the index to `(pyserini)/indexes`. Finally, delete the gz file in cache. Should we keep the gz file in cache?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,22 @@ for i in range(0, 10):
     print(f'{i+1:2} {hits2[i].docid:15} {hits2[i].score:.5f}')
 ```
 
+## How Do I build searchers?
+
+There are two ways to build searchers.
++ To build a searcher with a path to a directory containing the index.
+```python
+searcher = SimpleSearcher('indexes/index-robust04-20191213/')
+```
++ To build a searcher with the index's identifier name.
+```python
+searcher = SimpleSearcher.from_prebuilt_index('trec45')
+```
+It currently supports:
++ trec45 (TREC Disks 4 & 5)
++ ms-marco-passage (MS MARCO Passage)
++ ms-marco-doc (MS MARCO Doc)
+
 ## How Do I Fetch a Document?
 
 The other commonly used feature is to fetch a document given its `docid`.
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ searcher = SimpleSearcher.from_prebuilt_index('trec45')
 ```
 It currently supports:
 + trec45 (TREC Disks 4 & 5)
++ robust04 (TREC Disks 4 & 5)
 + ms-marco-passage (MS MARCO Passage)
 + ms-marco-doc (MS MARCO Doc)
 
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -20,6 +20,7 @@ class, which wraps the Java class with the same name in Anserini.
 """
 
 import logging
+import os
 from typing import Dict, List, Optional, Union
 
 from ._base import Document, JQuery, JQueryGenerator
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -27,7 +27,7 @@ from ._base import Document, JQuery, JQueryGenerator
 from pyserini.pyclass import autoclass, JString, JArrayList
 from pyserini.trectools import TrecRun
 from pyserini.fusion import FusionMethod, reciprocal_rank_fusion
-from pyserini.util import download_index
+from pyserini.util import download_prebuilt_index
 
 logger = logging.getLogger(__name__)
 
@@ -51,11 +51,8 @@ class SimpleSearcher:
         self.num_docs = self.object.getTotalNumDocuments()
 
     @classmethod
-    def from_prebuilt_index(cls, prebuilt_index_name_or_path: str):
-        if not os.path.isdir(prebuilt_index_name_or_path):
-            index_dir = download_index(prebuilt_index_name_or_path)
-        else:
-            index_dir = prebuilt_index_name_or_path
+    def from_prebuilt_index(cls, prebuilt_index_name: str):
+        index_dir = download_prebuilt_index(prebuilt_index_name)
         return cls(index_dir)
 
     def search(self, q: Union[str, JQuery], k: int = 10, query_generator: JQueryGenerator = None, strip_segment_id=False, remove_dups=False) -> List[JSimpleSearcherResult]:
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -20,7 +20,6 @@ class, which wraps the Java class with the same name in Anserini.
 """
 
 import logging
-import os
 from typing import Dict, List, Optional, Union
 
 from ._base import Document, JQuery, JQueryGenerator
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -26,6 +26,7 @@ from ._base import Document, JQuery, JQueryGenerator
 from pyserini.pyclass import autoclass, JString, JArrayList
 from pyserini.trectools import TrecRun
 from pyserini.fusion import FusionMethod, reciprocal_rank_fusion
+from pyserini.util import download_index
 
 logger = logging.getLogger(__name__)
 
@@ -44,7 +45,11 @@ class SimpleSearcher:
         Path to Lucene index directory.
     """
 
-    def __init__(self, index_dir: str):
+    def __init__(self, index_name_or_dir: str):
+        if not os.path.isdir(index_name_or_dir):
+            index_dir = download_index(index_name_or_dir)
+        else:
+            index_dir = index_name_or_dir
         self.object = JSimpleSearcher(JString(index_dir))
         self.num_docs = self.object.getTotalNumDocuments()
 
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -46,13 +46,17 @@ class SimpleSearcher:
         Path to Lucene index directory.
     """
 
-    def __init__(self, index_name_or_dir: str):
-        if not os.path.isdir(index_name_or_dir):
-            index_dir = download_index(index_name_or_dir)
-        else:
-            index_dir = index_name_or_dir
+    def __init__(self, index_dir: str):
         self.object = JSimpleSearcher(JString(index_dir))
         self.num_docs = self.object.getTotalNumDocuments()
+
+    @classmethod
+    def from_prebuilt_index(cls, prebuilt_index_name_or_path: str):
+        if not os.path.isdir(prebuilt_index_name_or_path):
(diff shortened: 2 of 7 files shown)


lintool:
> Currently, this change supports 'ms-marco-passage', 'ms-marco-doc' and 'TREC Disks 4 & 5'.
> 
> * If the index exists, skip the download and use the index under '(pyserini)/indexes'.

I think we should _always_ look for the prebuilt index in `~/.cache/pyserini/indexes`, since the method is `from_prebuilt_index`. If the user also happens to have another index lying around on the same collection, there shouldn't be any confusion.

What if we did something like this: download, check md5, and then append the md5 to the name of the index?

> * If not, download the index to cache(`~/.cache/pyserini/indexes`) and extract the index to `(pyserini)/indexes`. Finally, delete the gz file in cache. Should we keep the gz file in cache?

Yes, we should remove the gz file in the cache.

pyserini/util.py
```diff
@@ -21,7 +21,24 @@
...
 from tqdm import tqdm
 from urllib.request import urlretrieve
-
+from collections import OrderedDict
+
+INDEX_MAPPING = OrderedDict(
+    [
+        (
+            'ms-marco-passage',
+            'https://git.uwaterloo.ca/jimmylin/anserini-indexes/raw/master/index-msmarco-passage-20191117-0ed488.tar.gz'
```
lintool, line 30: this should be a list, because we might have mirror locations for the index.

pyserini/util.py
```diff
@@ -21,7 +21,24 @@
...
+        ),
+        (
+            'ms-marco-doc',
+            'https://www.dropbox.com/s/awukuo8c0tkl9sc/index-msmarco-doc-20200527-a1ecfa.tar.gz?dl=1'
+        ),
+        (
+            'trec4_5',
```
lintool, line 37: `trec45`

qguo96:
Now, a searcher can be built like this `searcher = SimpleSearcher.from_prebuilt_index('ms-marco-passage')`.

I write a new function `download_to_cache_and_unpack_index` so that it downloads the index to cache; Otherwise, the old function would be messy.

The current index name is `index_name` plus md5.

pyserini/util.py
```diff
@@ -21,7 +21,24 @@
...
 from tqdm import tqdm
 from urllib.request import urlretrieve
-
+from collections import OrderedDict
+
+INDEX_MAPPING = OrderedDict(
+    [
+        (
+            'ms-marco-passage',
+            {'uwaterloo': 'https://git.uwaterloo.ca/jimmylin/anserini-indexes/raw/master/index-msmarco-passage-20191117-0ed488.tar.gz'}
```
qguo96, line 30: I use a dictionary here. We may choose the mirror used for download later., but now I use the first one by default.

pyserini/util.py
```diff
@@ -21,7 +21,33 @@
...
 from urllib.request import urlretrieve
-
+from collections import OrderedDict
+
+INDEX_MAPPING = OrderedDict(
+    [
+        (
+            'ms-marco-passage',
+            {
+                'urls': {'uwaterloo': 'https://git.uwaterloo.ca/jimmylin/anserini-indexes/raw/master/index-msmarco-passage-20191117-0ed488.tar.gz'},
```
qguo96, line 31: I use a dictionary here. We may choose the mirror used for download later, but now I use the first one by default.
I also add md5 of the index.

pyserini/search/_searcher.py
```diff
@@ -48,6 +50,14 @@ def __init__(self, index_dir: str):
         self.object = JSimpleSearcher(JString(index_dir))
         self.num_docs = self.object.getTotalNumDocuments()
 
+    @classmethod
+    def from_prebuilt_index(cls, prebuilt_index_name_or_path: str):
```
lintool, line 54: why `_path` here? "ms-marco-passage", stored in `INDEX_MAPPING`,  is just a name right?

pyserini/util.py
lintool: sg

pyserini/util.py
```diff
@@ -21,7 +21,33 @@
 import tarfile
 from tqdm import tqdm
 from urllib.request import urlretrieve
-
+from collections import OrderedDict
+
+INDEX_MAPPING = OrderedDict(
```
lintool, line 26: Why does this need to be ordered?

pyserini/util.py
```diff
@@ -71,6 +97,10 @@ def download_url(url, save_dir, md5=None, force=False, verbose=True):
     if md5:
         assert compute_md5(destination_path) == md5, f'{destination_path} does not match checksum!'
 
+def get_cache_home():
+    return os.path.expanduser(
+        os.getenv("PYS_HOME", os.path.join(os.getenv("XDG_CACHE_HOME", f'~{os.path.sep}.cache'), "pyserini"))
```
lintool, line 102: Are `PYS_HOME` and `XDG_CACHE_HOME` commonly used?

What about something simpler like https://stackoverflow.com/questions/4028904/how-to-get-the-home-directory-in-python

pyserini/util.py
```diff
@@ -103,3 +133,55 @@ def download_and_unpack_index(url, index_directory='indexes', force=False, verbo
     tarball.extractall(index_directory)
     tarball.close()
     os.remove(local_tarball)
+
+def download_to_cache_and_unpack_index(url, md5, force=False, verbose=True):
```
lintool, line 137: There's a fair bit of code duplication here between this method and `download_and_unpack_index`. Wouldn't it make sense to refactor to use common code?

Isn't this method just `download_and_unpack_index` to a fixed directory (user cache) with a bit error checking? Do we need a separate code path?

lintool: Also, can we make sure `prebuilt` is in the name to make it clear what the method does?

pyserini/util.py
```diff
@@ -103,3 +133,55 @@ def download_and_unpack_index(url, index_directory='indexes', force=False, verbo
...
+        print(f'Extracting {local_tarball} into {destination_path}...')
+    tarball = tarfile.open(local_tarball)
+    tarball.extractall(cache_index_directory)
+    tarball.close()
+    os.remove(local_tarball)
+    os.rename(index_path, destination_path)
+    return destination_path
+
+
+def download_index(index_name, force=False, verbose=True, mirror=None):
```
lintool, line 177: Can we make sure `prebuilt` is in the name to make it clear what the method does?

lintool (changes requested):

lintool:
hi @shaneding and @adamyy - can you hop in, comment on this, and actually try out the feature when it's ready?

pyserini/search/_searcher.py
qguo96: It's just a name. Users can also provide an index directory to build a searcher. I can remove `_path` if you want to use prebuilt_index only.

lintool: yea, this method should be "name" only - following hgf, right? With hgf, you initialize with an identifier, not an explicit path...?

qguo96: hgf `from_pretrained` method also supports a path to a directory.

from_pretrained(pretrained_model_name_or_path, *model_args, **kwargs):
pretrained_model_name_or_path –
either:
+ shortname/identifier
+ a path to a directory containing model weights saved using save_pretrained(), e.g.: ./my_model_directory/


I'm a bit curious here, do we want this function to follow hgf convention or support name only?

qguo96:
+ change `from_prebuilt_index` to support index_name only
+ refactor `download_and_unpack_index` function

pyserini/search/_searcher.py
lintool: hrm. didn't realize that about hgf. 

Let's just go with "shortname/identifier".

Because the path is simply our normal constructor:

```python
searcher = SimpleSearcher('indexes/index-robust04-20191213/')
```

pyserini/util.py
```diff
@@ -22,6 +22,31 @@
...
+        ),
+        (
+            'trec45',
+            {
+                'urls': {'uwaterloo': 'https://git.uwaterloo.ca/jimmylin/anserini-indexes/raw/master/index-robust04-20191213.tar.gz'},
+                'md5': '15f3d001489c97849a010b0a4734d018'
+            }
+        )
+    ]
+)
```
adamyy, line 49: super nitpick: why not just

```python
INDEX_MAPPING = {
    "ms-marco-passage": {
        "urls":{},
        "md5": ""
    }
}
```

lintool: Good suggestion.

lintool:
@adamyy can you pull the branch and make sure it works also?
Can you also make sure it doesn't break anything here? https://github.com/castorini/anserini/blob/master/docs/experiments-covid.md

I want a serverless version of CORD-19 also soonish, so this is in your critical path. Thanks!

adamyy:
@lintool Noted! Will do!

adamyy:
Meanwhile, perhaps we should update the documentation to include the usage of `from_prebuilt_index `? @qguo96

lintool:
👍

qguo96:
Sure. @adamyy @lintool Where would you like me to update the document? Is a section in homepage good?

README.md
```diff
@@ -112,6 +112,22 @@ for i in range(0, 10):
     print(f'{i+1:2} {hits2[i].docid:15} {hits2[i].score:.5f}')
 ```
 
+## How Do I build searchers?
```
lintool, line 115: yea, this is fine for now... I can always tweak later...

adamyy:
Bc I am physically located on the other side of the planet, downloading the index has been tremendously slow... And it might take a while for me to verify this manually.

A small issue I discovered is that if the previous download has failed, then one must delete `~/.cache/pyserini/<index-md5>` directory before trying to redownload the index. Otherwise the script throws error like

```
EOFError: Compressed file ended before the end-of-stream marker was reached
```

@qguo96 Perhaps something to look out for? I mean the `ms-marco-doc` index seems to be >10GB and there is a nontrivial chance that the download can fail.

lintool:
Let's ask @salman1993 to help vet the end-to-end downloading process then...

qguo96:
@adamyy that should be fixed now, but you still need to rerun the program if your download fails, like function `download_url`.

shaneding:
End to end regression was fine! Lgtm.

lintool:
@qguo96 sorry, I lost track of this - one final request. Can we add the ability to have multiple aliases referring to the same index, e.g., `robust04` as an alternative name to `trec45`?

Thanks!

qguo96:
Sure, I will update it later.

lintool (approved):

lintool:
This is awesome, thanks!

PR #235, opened 2020-09-27
More prebuilt indexes features

qguo96:
- Add `SimpleSearcher.list_prebuilt_indexes()`, `IndexReader.from_prebuilt_index()` and `IndexReader.list_prebuilt_indexes()`
- I use a JSON file to replace the dictionary


qguo96:
This is a demo of function `list_prebuilt_indexes()`:
Screenshot (1223)
Screenshot (1224)

lintool:
Nice! A few suggestions/comments:

- what if the output is just a pandas dataframe?
- size: just make exact number of bytes? (uncompressed = sum of bytes of all files)?
- `unique terms = -1` - that might be an issue with an underlying method... can you file an issue separate and I'll look into it?
- `pyserini/indexes_info.json`: can we just take the JSON structures and move into directly into Python (i.e., basically make it a `.py' file)? I think this will make PyPI distribution easier, so we don't forget about an additional file...

qguo96:
hi! @lintool, I just created an issue here https://github.com/castorini/pyserini/issues/236.

qguo96:
- Both `uncompressed size` and `compressed size` are displayed in bytes.
- Now the output is a pandas data frame.
- That JSON file is rewritten in a py file.

Here is a demo of output:
Screenshot (1229)
Screenshot (1228)

pyserini/index/_base.py
```diff
@@ -150,6 +151,15 @@ def __init__(self, index_dir):
         self.object = JIndexReader()
         self.reader = self.object.getReader(JString(index_dir))
 
+    @classmethod
```
lintool, line 154: Can we add docstrings to this and below?

lintool (changes requested):
lg. please add docstrings, and we're ready for merging.

pyserini/index/_base.py
qguo96: I add docstrings to `IndexReader` and `SimpleSearcher`. Ready for review.

qguo96:
add information about `enwiki` and `zhwiki`.

lintool (approved):

PR #249, opened 2020-11-19
Index download bugfix

lintool:
Definition `local_tarball` needs to be pulled out of the if branch to top level, otherwise when the else branch is taken, the variable won't be defined when used later.

### DIFF
--- a/pyserini/util.py
+++ b/pyserini/util.py
@@ -97,6 +97,11 @@ def download_and_unpack_index(url, index_directory='indexes', force=False, verbo
 
         if not os.path.exists(index_directory):
             os.makedirs(index_directory)
+
+        # If there's a local tarball, it's likely corrupted, because we remove the local tarball on success (below).
+        # So, we want to remove.
+        if os.path.exists(local_tarball):
+            os.remove(local_tarball)
     else:
         index_path = os.path.join(index_directory, f'{index_name}')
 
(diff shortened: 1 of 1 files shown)


rodrigonogueira4 (approved):
LGTM, but don't you want to remove the local_tarball as you did before inside the `if prebuild`?

pyserini/util.py
```diff
@@ -89,17 +89,14 @@ def get_cache_home():
...
 
     if prebuilt:
         index_directory = os.path.join(get_cache_home(), 'indexes')
         index_path = os.path.join(index_directory, f'{index_name}.{md5}')
 
         if not os.path.exists(index_directory):
             os.makedirs(index_directory)
-
-        local_tarball = os.path.join(index_directory, f'{index_name}.tar.gz')
-        if os.path.exists(local_tarball):
```
qguo96, line 101: You may still need to keep this.
if the previous download has failed, then one must delete ~/.cache/pyserini/<index-md5> directory before trying to redownload the index. Otherwise the script throws error like
```
EOFError: Compressed file ended before the end-of-stream marker was reached
```

If the previous download is successful, the `local_tarball` should be deleted automatically. If there exists `local_tarball`, the previous download fails.

lintool: Yea, good call!

PR #265, opened 2020-11-26
PyPI direct package replication instructions

lintool:
We're able to directly replicate runs from the PyPI package!

As simple, as:

```bash
$ python -m pyserini.search --topics robust04 --index robust04 --output run.robust04.txt --bm25
```

Using pre-built indexes, stored topics, etc.


ronakice (approved):
LGTM.

PR #267, opened 2020-12-02
Improved downloading of prebuilt indexes

lintool:
+ Robust04 index moved, fixed location
+ Added ability to download from alternate URLs
+ Added method to validate index stats
+ Fixed check to see if index has been downloaded

### DIFF
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -67,8 +67,8 @@ class SimpleSearcher:
         print(f'Attempting to initialize pre-built index {prebuilt_index_name}.')
         try:
             index_dir = download_prebuilt_index(prebuilt_index_name)
-        except ValueError:
-            print(f'Unknown pre-built index: {prebuilt_index_name}')
+        except ValueError as e:
+            print(str(e))
             return None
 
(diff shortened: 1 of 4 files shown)


pyserini/search/_searcher.py
```diff
@@ -67,8 +67,8 @@ def from_prebuilt_index(cls, prebuilt_index_name: str):
         print(f'Attempting to initialize pre-built index {prebuilt_index_name}.')
         try:
             index_dir = download_prebuilt_index(prebuilt_index_name)
-        except ValueError:
-            print(f'Unknown pre-built index: {prebuilt_index_name}')
+        except ValueError as e:
+            print(str(e))
```
rodrigonogueira4, line 71: Shouldn't we raise an exception?

rodrigonogueira4 (approved):
LGTM, just a minor comment.

pyserini/search/_searcher.py
lintool: I think we're good here because we return a `None`. This is so that the caller can just check for `None` instead of needing to trap an exception.

PR #280, opened 2020-12-20
add trec-covid pre-built index to pyserini

yuki617:
Add trec-covid round1-5  pre-built index

### DIFF
--- a/pyserini/prebuilt_index_info.py
+++ b/pyserini/prebuilt_index_info.py
@@ -159,9 +159,9 @@ INDEX_INFO = {
         ],
         "md5": "c883571ccc78b4c2ce05b41eb07f5405",
         "size compressed (bytes)": 2796524,
-        "total_terms": -1,
-        "documents": 18,
-        "unique_terms": -1,
+        "total_terms": 22100404,
+        "documents": 192459,
+        "unique_terms": 195875,
         "downloaded": False
     },
     "trec-covid4": {
@@ -171,9 +171,9 @@ INDEX_INFO = {
         ],
         "md5": "029bd55daba8800fbae2be9e5fcd7b33",
         "size compressed (bytes)": 2584264,
-        "total_terms": -1,
-        "documents": 18,
-        "unique_terms": -1,
+        "total_terms": 18724353,
+        "documents": 158226,
+        "unique_terms": 179937,
         "downloaded": False
     },
     "trec-covid3": {
@@ -183,9 +183,9 @@ INDEX_INFO = {
         ],
         "md5": "37bb97d0c41d650ba8e135fd75ae8fd8",
         "size compressed (bytes)": 2190328,
-        "total_terms": -1,
-        "documents": 18,
-        "unique_terms": -1,
+        "total_terms": 16278419,
+        "documents": 128465,
+        "unique_terms": 168291,
         "downloaded": False
     },
     "trec-covid2": {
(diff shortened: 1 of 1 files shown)


pyserini/prebuilt_index_info.py
```diff
@@ -151,5 +151,65 @@
...
+    },
+    "trec-covid5": {
+        "description": "Trec Covid Round 5",
+        "urls": [
+            "https://www.dropbox.com/s/9hfowxi7zenuaay/lucene-index-cord19-abstract-2020-07-16.tar.gz?dl=1"
+        ],
+        "md5": "c883571ccc78b4c2ce05b41eb07f5405",
+        "size compressed (bytes)": 2796524,
+        "total_terms": -1,
+        "documents": 18,
```
lintool, line 163: these stats don't seem right?

lintool: This method should check out:
https://github.com/castorini/pyserini/blob/master/pyserini/index/_base.py#L180

lintool (changes requested):
Also, mirror in gitlab.

pyserini/prebuilt_index_info.py
yuki617: Sorry, the number should be fixed now.

pyserini/prebuilt_index_info.py
```diff
@@ -151,5 +151,65 @@
         "documents": 4170312,
         "unique_terms": -1,
         "downloaded": False
+    },
+    "trec-covid5": {
```
lintool, line 155: Please change to `trec-covid-r5-abstract`
Description should be "TREC-COVID Round 5: abstract index"
Add gitlab link here: https://git.uwaterloo.ca/jimmylin/cord19-indexes

We should also have:
+`trec-covid-r5-full-text` description: "TREC-COVID Round 5: full-text index"
+`trec-covid-r5-paragraph` description: "TREC-COVID Round 5: paragraph index"

pyserini/prebuilt_index_info.py
```diff
@@ -151,5 +151,65 @@
...
+            "https://www.dropbox.com/s/9hfowxi7zenuaay/lucene-index-cord19-abstract-2020-07-16.tar.gz?dl=1"
+        ],
+        "md5": "c883571ccc78b4c2ce05b41eb07f5405",
+        "size compressed (bytes)": 2796524,
+        "total_terms": 22100404,
+        "documents": 192459,
+        "unique_terms": 195875,
+        "downloaded": False
+    },
+    "trec-covid4": {
```
lintool, line 167: For all rounds, we should have three separate indexes, like above.

lintool (changes requested):

lintool:
Please add additional indexes - per comments.

pyserini/prebuilt_index_info.py
yuki617: All full-text and paragraph pre-built index are added

lintool:
Gitlab URLs need to be changed to this form `https://git.uwaterloo.ca/jimmylin/anserini-indexes/raw/master/index-robust04-20191213.tar.gz` so that we're downloading the raw tarball, not the HTML landing page.

yuki617:
> the raw tarball, not the HTML landing page.
Thanks for catching that. The urls should be the raw tarball ones now

lintool (approved):

PR #287, opened 2021-01-01
Replication guide for KILT BM25 baseline

yuxuan-ji:
These are the scripts used to create runs for KILT with the anserini baseline. I placed the instructions and results in the docs. 
I tested 3 indexing level: document (entire doc), sections (header separation) and passage (newline separation).

I also tried the naive bigram implementation where we just create bigrams (filtering out punctuation and stopwords) and append it to the documents (doing this for input queries as well), but apart from the document-level index, this made performance worse throughout.

image


docs/kilt-baseline-replication.md
```diff
@@ -0,0 +1,136 @@
+# Replication guide KILT
+
+## Setup env
+
+```bash
+# Create a virtual env
+conda create -n kilt37 -y python=3.7 && conda activate kilt37
```
lintool, line 7: We're using Python 3.6 project wide, can you please switch?

docs/kilt-baseline-replication.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Get pyserini scripts and install package
```
lintool, line 9: Can you just say, we need the development installation, see instructions here:
https://github.com/castorini/pyserini/#development-installation

docs/kilt-baseline-replication.md
```diff
@@ -0,0 +1,136 @@
...
+git clone https://github.com/castorini/pyserini.git
+pip install pyserini
+
+# Get KILT scripts and install package
+git clone https://github.com/facebookresearch/KILT.git
```
lintool, line 14: Can you clarify the directory structure where you want to put everything?

I think it might make sense to have everything under `pyserini/`? For example, the KILT corpus can go into `collections/`?

docs/kilt-baseline-replication.md
```diff
@@ -0,0 +1,136 @@
...
+python -m nltk.downloader stopwords
+```
+
+## Index the knowledge source
+
+Convert to passage or document level JSONL format indexable by Pyserini (takes 1-2hr if bigrams is enabled):
+
+```bash
+# We'll split it in multiple files to make this faster
+mkdir kilt_knowledge_split
```
lintool, line 39: Unclear what directory you're in at the moment... this should probably be under `collections/` subdirectory so we can keep the directory structure clean?

docs/kilt-baseline-replication.md
```diff
@@ -0,0 +1,136 @@
...
+# Once it's done, convert back into 1 file:
+cat dump/passage_dump/kilt_ks.?? > dump/passage_dump/dump.jsonl
+rm dump/passage_dump/kilt_ks.??
+``` 
+
+Index into Anserini (about 1hr). These instructions are the same whether you are using document or passage level index (just change the path accordingly).:
+
+```bash
+nohup python -m pyserini.index -collection JsonCollection -generator DefaultLuceneDocumentGenerator \
+ -threads 40 -input <path to dump.jsonl e.g. dump/doc_dump/> \
```
lintool, line 91: put in the actual file name here...

lintool (changes requested):
Thanks, this is great! Some comments for you...

lintool:
Let me know explicitly when you're ready for re-review.

docs/kilt-baseline-replication.md
yuxuan-ji: This is the environment KILT uses, so I'm not sure if there'd be any dependency issues with their scripts by switching to 3.6. My current setup uses both scripts in pyserini and scripts from the KILT repository.

lintool: okay, let's keep then.

docs/kilt-baseline-replication.md
yuxuan-ji: I was planning on updating these instructions with specific paths once we had a baseline that beats DrQA, but I was quite surprised to see that bigrams actually reduced the precision for the passage level index.

scripts/kilt/anserini_retriever.py
```diff
@@ -0,0 +1,155 @@
...
+            # doc_scores = [hit.score for hit in hits]
+
+        except RuntimeError as e:
+            if logger:
+                logger.warning("RuntimeError: {}".format(e))
+        except jnius.JavaException as e:
+            if logger:
+                logger.warning("{query} jnius.JavaException: {}".format(query_element, e))
+            if 'maxClauseError' in str(e):
+                query = " ".join(query.split()[:950])
```
yuxuan-ji, line 85: A hacky way to handle maxClauseError (1024), which happens mostly for Entity Linking tasks (inputs are entire paragraphs)

lintool (approved):

PR #291, opened 2021-01-04
replicate pypi doc and update result

MXueguang:
please see results diff in MS MARCO Document Ranking
I think the result I got is too high?

### DIFF
--- a/docs/pypi-replication.md
+++ b/docs/pypi-replication.md
@@ -36,8 +36,8 @@ To evaluate:
 $ wget -O jtreceval-0.0.5-jar-with-dependencies.jar https://search.maven.org/remotecontent?filepath=uk/ac/gla/dcs/terrierteam/jtreceval/0.0.5/jtreceval-0.0.5-jar-with-dependencies.jar
 $ wget https://raw.githubusercontent.com/castorini/anserini/master/src/main/resources/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt
 $ java -jar jtreceval-0.0.5-jar-with-dependencies.jar -m map -c -m recall.1000 -c qrels.msmarco-passage.dev-subset.txt run.msmarco-passage.txt
-map                   	all	0.1958
-recall_1000           	all	0.8573
+map                   	all	0.1926
+recall_1000           	all	0.8526
 ```
 
 MS MARCO passage ranking task, BM25 baseline with [docTTTTTquery expansions](http://doc2query.ai/):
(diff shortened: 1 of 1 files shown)


lintool:
Hrm... odd - can you try replicating these first:
https://github.com/castorini/anserini/blob/master/docs/experiments-msmarco-doc-leaderboard.md

There's some overall between the runs, which will help us debug.

Also, can you check if you're returning 100 hits or 1000 hits?
Also, did you create a new conda env with `pyserini==0.10.0.1` (i.e., latest)? Wondering if it's an issue of what's published on PyPI vs. what's in the current repo (i.e., dev copy).

lintool:
Ah, here's a clue:
https://github.com/castorini/pyserini/blob/pyserini-0.10.0.0/pyserini/search/__main__.py
Hits hard coded to 1000 - https://github.com/castorini/pyserini/blob/pyserini-0.10.0.0/pyserini/search/__main__.py#L110
Even though for MS MARCO doc we only evaluate top 100

In `0.10.0.1` - the hits arg has been parameterized:
https://github.com/castorini/pyserini/blob/pyserini-0.10.0.1/pyserini/search/__main__.py

lintool:
@MXueguang can you run the same commands on master HEAD and compare results?

MXueguang:
> @MXueguang can you run the same commands on master HEAD and compare results?

sure

MXueguang:
I replicated with master HEAD, which gives the results in this PR.

The results with pyserini==0.10.0.1 are identical to the document

lintool:
So which version was causing the issue? Can you make of note of that in the docs? Update your PR then please.

docs/pypi-replication.md
```diff
@@ -70,6 +70,36 @@ To evaluate:
...
+  E.g., `msmarco-doc` should be change to  `ms-marco-doc`. Exactly same results are expected to obtain.
+  
+- WARNING: In `pyserini==0.10.0.0`, hits are hard coded to 1000. see [here](https://github.com/castorini/pyserini/blob/pyserini-0.10.0.0/pyserini/search/__main__.py#L110).
+  Even though for MS MARCO doc we only evaluate top 100.
+  In `pyserini==0.10.0.1`, the hits are has been parameterized. see [here](https://github.com/castorini/pyserini/blob/pyserini-0.10.0.1/pyserini/search/__main__.py#L112).
+  
+CAVEAT: current HEAD of master is getting different results.
+```
+MS MARCO passage ranking task, BM25 baseline:
+map                   	all	0.1958
```
lintool, line 89: Wait, aren't these matching the numbers above at (a) ^^^^?

docs/pypi-replication.md
```diff
@@ -36,8 +36,8 @@ To evaluate:
 $ wget -O jtreceval-0.0.5-jar-with-dependencies.jar https://search.maven.org/remotecontent?filepath=uk/ac/gla/dcs/terrierteam/jtreceval/0.0.5/jtreceval-0.0.5-jar-with-dependencies.jar
 $ wget https://raw.githubusercontent.com/castorini/anserini/master/src/main/resources/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt
 $ java -jar jtreceval-0.0.5-jar-with-dependencies.jar -m map -c -m recall.1000 -c qrels.msmarco-passage.dev-subset.txt run.msmarco-passage.txt
-map                   	all	0.1958
```
lintool: (a) here ->

lintool (commented):
Questions? Aren't the scores matching?

docs/pypi-replication.md
MXueguang: I made an update but revert it back. This is the result of HEAD of master branch. I change it back to result of `pyserini==0.10.0.1`

docs/pypi-replication.md
MXueguang: no, we should compare to the scores here, https://github.com/castorini/pyserini/blob/master/docs/pypi-replication.md

lintool: Oh shoot, you're right. The difference comes from different default parameters...

ARG, this is annoying, but this is fine as is - I'll circle back and redo once we publish `0.10.0.2`.

Thanks for the replication.

lintool (approved):
lgtm - please go ahead and merge.

PR #299, opened 2021-01-05
draft of hybrid search

MXueguang:
very first draft
create the draft pr to discuss


lintool:
What do you think about `pyserini.hsearch` to parallel `search` and `dsearch`

MXueguang:
sg

MXueguang:
replicated hybrid `tct_colbert hnsw index`  + `doct5query index`
```
6980/6980 [14:48<00:00,  7.86it/s]
MRR @10: 0.3634984877427564
map                     all     0.3700
recall_1000             all     0.9733
```

MXueguang:
replicated hybrid `tct_colbert brute force index` + `doct5query index`
```
582/582 [1:38:09<00:00, 10.12s/it]
MRR @10:                        0.36390321553645266
map                     all     0.3704
recall_1000             all     0.9734
```

expected result in paper:
```
mrr@10:      0.364 
recall@1000: 0.973
```

matched!

justram:
LGTM, wait for Jack's confirmation?

docs/dense-retrieval.md
```diff
@@ -76,3 +76,80 @@ recall_1000           	all	0.9637
...
+- dense retrieval with TCT-ColBERT, HNSW index.
+- sparse retrieval with doc2query-T5 expanded index.
+
+```
+python -m pyserini.hsearch --topics msmarco_passage_dev_subset \
+                             --dindex msmarco-passage-tct_colbert-hnsw \
+                             --sindex msmarco-passage-expanded \
+                             --encoded-queries msmarco-passage-dev-subset-tct_colbert \
+                             --output runs/run.msmarco-passage.tct_colbert.hnsw.doc2queryT5.tsv \
+                             --msmarco \
```
justram, line 92: Should we set batch size here?

lintool:
So, currently, dense and sparse methods are hardcoded, e.g., how many hits, possible weighting between dense/sparse... this is all fine, we can circle back and add options later.

docs/dense-retrieval.md
MXueguang: `batch_search` & `simple_search` both gives an acceptable result, I was trying to do one without `batch_search`. But we can set the batch, which will make it faster than `simple_search`. What do you think @lintool ?

docs/dense-retrieval.md
```diff
@@ -76,3 +76,80 @@ recall_1000           	all	0.9637
...
+### Hybrid Dense-Sparse Ranking
+MS MARCO passage ranking task, 
+Hybrid
+- dense retrieval with TCT-ColBERT, HNSW index.
+- sparse retrieval with doc2query-T5 expanded index.
+
+```
+python -m pyserini.hsearch --topics msmarco_passage_dev_subset \
+                             --dindex msmarco-passage-tct_colbert-hnsw \
+                             --sindex msmarco-passage-expanded \
```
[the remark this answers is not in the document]
MXueguang: sure

docs/dense-retrieval.md
lintool: See my comment below -

MXueguang:
sg

lintool:
So, I was thinking, and it doesn't need to be in this iteration, of implementation something like multiple subcommands, e.g.

```
python -m pyserini.hsearch --topics msmarco_passage_dev_subset 
 dense --index ... [all the usual dense options...]
 sparse --index ... [all the usual sparse options...]
 fusion [fusion options, like fusion weights, etc.]
```

See: https://stackoverflow.com/questions/10448200/how-to-parse-multiple-nested-sub-commands-using-python-argparse

This way, we don't need to re-implement options, just "dispatch" the the dense and sparse argparsers...

Thoughts?

lintool:
So the answer to the above comment is, in my proposed design, `--batch` and `--threads` would only apply to "dense".

justram:
It would be good to separate dense options from sparse options.
We may want to use other options of FAISS or change the backend in the future.

jacklin64:
It looks good to me! Agree to separate dense and sparse argparsers

pyserini/hsearch/_hybrid.py
```diff
@@ -0,0 +1,71 @@
...
+        return hybrid_result
+
+    @staticmethod
+    def _hybrid_results(dense_results, sparse_results, alpha):
+        dense_hits = {hit.docid: hit.score for hit in dense_results}
+        sparse_hits = {hit.docid: hit.score for hit in sparse_results}
+        hybrid_result = []
+        for doc in set(dense_hits.keys()) | set(sparse_hits.keys()):
+            if doc not in dense_hits:
+                score = alpha * sparse_hits[doc] + min(dense_hits.values())
```
jacklin64, line 65: Just a very minor comment here. min(dense_hits.values()) and min(sparse_hits.values()) can be precomputed before the for loop so they don't be computed many times.

MXueguang: yeah! update rn!

lintool (approved):
lgtm, let's get this merged in!

PR #296, opened 2021-01-05
Add multi-threading support

mrkarezina:
Added multithreading support with `batch_search`.

Tested to make sure output is the same. Environment: macOS 11.1 and Python 3.7.5

## Tests
**Original output**
```
python -m pyserini.search --topics robust04 --index robust04 --output run.robust04.txt --bm25
```
MD5 (run.robust04.txt) = d38797d0a284525d1d7ee35c00a9e5fd

```
python -m pyserini.search --topics msmarco_passage_dev_subset --index msmarco-passage --output runs/run.msmarco-passage.1.txt --bm25
```
MD5 (runs/run.msmarco-passage.1.txt) = 48ecace50020910580fc72d58fb9086c

**Modified script**
```
python -m pyserini.search --topics robust04 --index robust04 --output run.robust04-batch.1.txt --bm25
```
MD5 (run.robust04-batch.1.txt) = d38797d0a284525d1d7ee35c00a9e5fd

```
python -m pyserini.search --topics robust04 --index robust04 --output run.robust04-batch.2.txt --bm25 --threads 3 --batch-size 64
```
MD5 (run.robust04-batch.2.txt) = d38797d0a284525d1d7ee35c00a9e5fd

```
python -m pyserini.search --topics msmarco_passage_dev_subset --index msmarco-passage --output runs/run.msmarco-passage.1-batch.txt --bm25 --threads 4 --batch-size 64
```
MD5 (runs/run.msmarco-passage.1-batch.txt) = 48ecace50020910580fc72d58fb9086c

PR #301, opened 2021-01-06
Add multi-threading support

mrkarezina:
Re-opening since https://github.com/castorini/pyserini/pull/296 was marked as merged. I will also add integration tests.


pyserini/search/__main__.py
```diff
@@ -130,37 +137,68 @@
...
+    if args.msmarco:
+        for i, docid in enumerate(docids):
+            target_file.write(f'{topic}\t{docid}\t{i + 1}\n')
+    else:
+        for i, (docid, score) in enumerate(zip(docids, scores)):
+            target_file.write(
+                f'{topic} Q0 {docid} {i + 1} {score:.6f} {tag}\n')
+
+
+def write_result_max_passage(result: Tuple[str, List[JSimpleSearcher]]):
```
lintool, line 158: This is actually more of a general Python question: How does this method get access to `target_file`, which is part of `with open(output_path, 'w') as target_file:` from the caller? It doesn't appear to be explicitly passed?

pyserini/search/__main__.py
```diff
@@ -130,37 +137,68 @@
...
-                    target_file.write(f'{topic}\t{docid}\t{rank}\n')
-                else:
-                    target_file.write(f'{topic} Q0 {docid} {rank} {hit.score:.6f} {tag}\n')
-                rank = rank + 1
-                unique_docs.add(docid)
-                if rank > args.max_passage_hits:
-                    break
+            batch_topics.append(topic)
+            if (index + 1) % args.batch_size == 0 or index == len(topics.keys()) - 1:
+                queries = [topics[topic].get('title')
```
lintool, line 188: We standardize on 120 character lines, so you can reformat to make more compact...

lintool:
I pulled your branch and ran:

```
python -m unittest
python -m unittest discover -s integrations
```

All tests pass.

pyserini/search/__main__.py
mrkarezina: In Python variables that are only referenced inside a function are [implicitly declared global](https://docs.python.org/3/faq/programming.html#id9). If you were to try assigning to `target_file` inside of `write_result` for example it would throw an error.

lintool:
Just in terms of sequencing this - our replication tests are a bit of a mess right now, see this issue: #291

I want to push a new PyPI artifact, re-sync all replications, then merge this issue in. So that we're only change one thing at a time...

mrkarezina:
When converting the command
```
python -m pyserini.search --topics msmarco_passage_dev_subset --index msmarco-passage --output runs/run.msmarco-passage.1.txt --bm25
```
into an integration test:
```
self.checker_msmarco = SimpleSearcherChecker(
            anserini_root=anserini_root,
            index=os.path.join(
                # TODO: Name of the index?
                anserini_root, 'indexes/lucene-index.index-msmarco-passage-20201117-f87c94'),
            topics=os.path.join(
                pyserini_root, 'tools/topics-and-qrels/topics.msmarco-doc.dev.txt'),
            pyserini_topics='msmarco_passage_dev_subset',
            qrels=os.path.join(pyserini_root, 'tools/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt'))
```
What is the name of the index?


Converting the `robust04` test like this works:
```
self.checker = SimpleSearcherChecker(
            anserini_root=anserini_root,
            index=os.path.join(
                anserini_root, 'indexes/lucene-index.robust04.pos+docvectors+raw'),
            topics=os.path.join(
                pyserini_root, 'tools/topics-and-qrels/topics.robust04.txt'),
            pyserini_topics='robust04',
            qrels=os.path.join(pyserini_root, 'tools/topics-and-qrels/qrels.robust04.txt'))
```

However for `msmarco_passage_dev_subset` although I have an index called `indexes/lucene-index.index-msmarco-passage-20201117-f87c94` in anserini, during the unit test it shows 0 run time:
```
2021-01-06 18:23:35,418 INFO  [main] search.SearchCollection (SearchCollection.java:727) - Total run time: 00:00:00
```

and `verify.anserini.msmarco_passage_dev_subset.txt` is empty. Same issue for `msmarco_doc_dev`.

lintool:
These are pre-built indexes - check out here: https://github.com/castorini/pyserini/blob/master/docs/prebuilt-indexes.md

mrkarezina:
Hmm I tried using the indexes directly in the `.cache` folder. Works for `robust04`:
```
self.checker = SimpleSearcherChecker(
            anserini_root=anserini_root,
            index=os.path.join('~/.cache/pyserini/indexes/index-robust04-20191213.15f3d001489c97849a010b0a4734d018'),
            topics=os.path.join(
                pyserini_root, 'tools/topics-and-qrels/topics.robust04.txt'),
            pyserini_topics='robust04',
            qrels=os.path.join(pyserini_root, 'tools/topics-and-qrels/qrels.robust04.txt'))
```

But for `msmarco_passage_dev_subset` there's no error, however same thing, 0 topics processed:
```
2021-01-06 19:52:09,273 INFO  [pool-2-thread-1] search.SearchCollection$SearcherThread (SearchCollection.java:254) - 0 topics processed in 00:00:00
``` 
Am I missing something here?
```
self.checker_msmarco = SimpleSearcherChecker(
            anserini_root=anserini_root,
            index='~/.cache/pyserini/indexes/index-msmarco-passage-20201117-f87c94.1efad4f1ae6a77e235042eff4be1612d',
            topics=os.path.join(
                pyserini_root, 'tools/topics-and-qrels/topics.msmarco-passage.dev-subset.txt'),
            pyserini_topics='msmarco_passage_dev_subset',
            qrels=os.path.join(pyserini_root, 'tools/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt'))
```

lintool:
You might need to write a new harness... `SimpleSearcherChecker` was designed to check the output of anserini with pyserini... I suppose you could co-opt it to do what you want... but probably a bit janky...

mrkarezina:
The issue was in `simplesearcher_checker.py` the Trec topic reader is specific in the anserini base command.
```
self.anserini_base_cmd = os.path.join(self.anserini_root,
                                              'target/appassembler/bin/SearchCollection -topicreader Trec')
``` 

Maybe the topic reader could be made a parameter with default value of Trec. Since two of the tests use msmarco which requires TsvInt topic reader. It currently works by specifying `-topicreader TsvInt` again in the command although it's a bit hacky.

lintool:
As I mentioned, `SimpleSearcherChecker` isn't relevant in this case because it was designed to verify the output of anserini against pyserini... which isn't our goal here...

mrkarezina:
I added a new harness, `CompareSimpleSearcher`. It compares multiple `pyserini.search` runs.

integrations/compare_simplesearcher.py
```diff
@@ -0,0 +1,69 @@
...
+        self.topics = topics
+        self.pyserini_base_cmd = 'python3 -m pyserini.search'
+
+    @staticmethod
+    def _cleanup(files: List[str]):
+        for file in files:
+            if os.path.exists(file):
+                os.remove(file)
+
+    def run(self, runtag: str, extras_one: str, extras_two: List[str]):
```
lintool, line 34: I'm not sure I like `extras_one` and `extras_two`...

How about this as a re-design:
- the simple harness takes a bunch of args, performs a run, and returns the MD5 checksum of the run.
- the actual test case takes the harness, submits two separate runs (with different threads and batch sizes), and compares the checksum; tests pass if the checksums match.
-

integrations/test_simplesearcher_multithread.py
```diff
@@ -0,0 +1,65 @@
...
+from integrations.compare_simplesearcher import CompareSimpleSearcher
+
+
+class TestSearchIntegration(unittest.TestCase):
+    def setUp(self):
+        index_cache_root = "~/.cache/pyserini/indexes/"
+
+        self.checker_robust = CompareSimpleSearcher(
+            index=os.path.join(index_cache_root,
+                               'index-robust04-20191213.15f3d001489c97849a010b0a4734d018'),
```
lintool, line 29: path and index name might not be stable... can we figure out how do use this?

```
searcher = SimpleSearcher.from_prebuilt_index(INDEX_NAME)
```

lintool (commented):
Suggestions... open to discussion.

lintool:
Check out this PR: #307

We should add replications for all of these: https://github.com/castorini/pyserini/blob/master/docs/pypi-replication.md

Some you have already...

Once we get the harness built, adding new runs should be pretty easy...

mrkarezina:
I refactored the test harness to return the checksum, access the indexes by name, and added 7 tests for the pypi replications.

lintool (approved):

PR #305, opened 2021-01-08
add query encoder and add script to create embeddings and index

MXueguang:
draft of query encoder
& scripts for creating embedding and index


lintool:
Are these scripts specific to MS MARCO? If so, I think we should put under `pyserini/scripts/msmarco-passage`?

Maybe later down the road we generalize...?

scripts/msmarco-passage/create_embedding.py
```diff
@@ -0,0 +1,67 @@
...
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+from transformers import BertModel, BertTokenizer
+
+
+def encode_query(text, tokenizer, model, device='cpu'):
+    max_length = 36  # hardcode for now
+    inputs = tokenizer(
+        '[CLS] [Q] ' + text + " [MASK]" * max_length,
```
lintool, line 12: why switch to double quotes here?

scripts/msmarco-passage/create_embedding.py
```diff
@@ -0,0 +1,67 @@
...
+        id_, text = line.rstrip().split('\t')
+        id_ = id_.strip()
+        text = text.strip()
+        if args.store_text:
+            embeddings['text'].append(text)
+        embeddings['id'].append(id_)
+        if args.type == 'passage':
+            embeddings['embedding'].append(encode_passage(text, tokenizer, model, args.device))
+        elif args.type == 'query':
+            embeddings['embedding'].append(encode_query(text, tokenizer, model, args.device))
```
lintool, line 65: I'm a bit unclear why you would want to encode documents into in pickled structure?

Also, MS MARCO passages are in a JSON structure...

From what I can tell, this is just a generic script that encodes text in a `id\ttext` format...? Is my understanding correct?

lintool: If this is the case, shouldn't this script just be named `encode_text.py`?

scripts/msmarco-passage/create_faiss_index.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+
+import faiss
+import numpy as np
+import pandas as pd
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--embedding", type=str, help='encoder name or path', required=True)
+    parser.add_argument("--index-type", type=str, help='doc file to get encoded.', required=True)
```
lintool, line 11: help string doesn't make sense?

scripts/msmarco-passage/create_faiss_index.py
```diff
@@ -0,0 +1,34 @@
...
+    parser.add_argument("--index-dir", type=str, help='doc type, [passage, query]', required=True)
```
lintool, line 12: Just `--index` for index location? To be consistent with sparse indexes and anserini.

scripts/msmarco-passage/create_faiss_index.py
```diff
@@ -0,0 +1,34 @@
...
+    args = parser.parse_args()
+
+    print('Loading Embeddings')
+    embedding_df = pd.read_pickle(args.embedding)
```
lintool, line 16: Oh, I think I see now... you pickle the encoded text, and then you run through here...

The corpus is big... 

I think your rationale is that you encode the corpus once, as pickled file, and then run through indexing script twice, once to generate flat index, the second time to generate hnsw index?

If that's the idea - how about the alternative: you encode and directly generate flat index. Then in a second pass you go over the flat index to generate the hnsw index. That way you don't need to materialize intermediate data in pickled form.

lintool (commented):
My comments might be a bit confusing... took a bit to figure out the logic of your scripts. Open to discussion.

lintool:
So, my suggestion would be to have three scrips:
1. to encode the queries (from tsv)
2. to encode the corpus (take directly from JSONL format) and build faiss flat index.
3. to build hnsw index from faiss flat index.

From what I understand, the faiss flat index is lossless (i.e., just stores the vectors), right? If so, this saves you from having to save pickled intermediate data...

scripts/msmarco-passage/create_faiss_index.py
MXueguang: sure

scripts/msmarco-passage/create_embedding.py
MXueguang: yes, I am assuming we are handling the text with `id\ttext`. 
I usually download ms marco passages from https://github.com/castorini/docTTTTTquery#data-and-trained-models-ms-marco-passage-ranking-dataset , it gives tsv file.

sure, ill rename it to `encode_text.py`

scripts/msmarco-passage/create_faiss_index.py
MXueguang: will update soon.

scripts/msmarco-passage/create_embedding.py
lintool: I see. Since users of Pyserini (and our URAs) will likely have gone through our onboarding process, we should base the input on this: https://github.com/castorini/pyserini/blob/master/docs/experiments-msmarco-passage.md

In other words, I would prefer JSONL.

scripts/msmarco-passage/create_faiss_index.py
MXueguang: good suggestion! we can always keep the brute force index as raw `embedding` data.
so we dont need the pickle for passage embedding. will do!

I worked on pickle because I feel it is easier to handle than tfrecord here.
For query embeddings, we create a single pickle file to store id and embeddings together? Maybe with the raw query text as well, since we may want to do the `text->embedding` map. I.e. `load_encoded_queries` load everyting from a pickle file

lintool: see my comment below - yes, queries are small, pickle would be fine.

In general, I tend to avoid pickle for large amounts of data...

Actually, for reference - the corpus encoded as vectors - what's the size of picked data, FAISS flat index, and TF record?

scripts/msmarco-passage/create_embedding.py
MXueguang: i see. sure!

scripts/msmarco-passage/create_faiss_index.py
MXueguang: 26G, 26G and 26G...

lintool: Interesting. Pickled data isn't any larger.

I was always under the impression that Pickle wasn't very space efficient... guess I was wrong.

MXueguang:
seems there is a way to store doc ids with index, so we don't need an extra file to store doc ids.
but seems it only support integer. (it is fine for msmarco-passage, because the ids are integers)
should we utilize that?

MXueguang:
I feel put the docid with index is better.
If one day we have to use "string" id, we can just convert that to Hexdecimal ASCII i guess

MXueguang:
emmm, it not works......revert back.

MXueguang:
@lintool encoded query contains id, raw and embedding in a single pickle file
do we still  need to compress it as `tar.gz`?

lintool:
Probably not... how large is it uncompressed?

MXueguang:
22Mb

lintool:
Why don't we keep compressed just to save a bit of bandwidth, since we already have all the machinery to download and decompress tarballs?

MXueguang:
sure

lintool (approved):
👍 
🚢 it!

PR #309, opened 2021-01-09
add query iterator

MXueguang:
in current version, we added a flag `--ordered` to the pypi replication to use the query order array.
```bash
$ python -m pyserini.search --topics msmarco_passage_dev_subset \
                                             --index msmarco-passage \
                                             --output run.msmarco-passage.txt \
                                             --bm25 --msmarco \
                                             --ordered
```


lintool (approved):
Can you kill the `--ordered` option? I can't think of a reason why you _wouldn't_ want ordered?

But otherwise, 🚢 it!

lintool:
Hrm, conflicts need to be resolved.

After this gets merged we need to re-do replication numbers.

Actually @MXueguang - one more thing, please document in query_iterator why we actually need to do this - a reference to the GitHub issue should do.

Happy to review it again if you want, but otherwise just merge please.

lintool:
We'll also need to rerun multi-threaded search and make sure there are no weird interactions.

MXueguang:
updated scores in `pypi-replication.md`

multithreading gives same results

lintool (approved):
lgtm!

PR #322, opened 2021-01-15
update re pickled query embedding info

MXueguang:
recreate query embedding pickle file from py36


lintool (approved):
I'll test after you merge.

Remember to close the relevant issue when you merge.

PR #323, opened 2021-01-16
update bm25Vectorizer

MXueguang:
- fix bug: Each time of initializing, the BM25 Vectorizer gives vocabulary list with different order, which will make results not replicable. Solved by sort the vocabulary list.
- change: remove `normalize`. If we want to normalize, it should be done externally.
- add: `get_query_vector`

### DIFF
--- a/pyserini/vectorizer/_base.py
+++ b/pyserini/vectorizer/_base.py
@@ -48,11 +48,10 @@ class Vectorizer:
 
         # build vocabulary
         self.vocabulary_ = set()
-        self.term_to_df = {}
         for term in self.index_reader.terms():
             if term.df > self.min_df:
                 self.vocabulary_.add(term.term)
-                self.term_to_df[term.term] = term.df
+        self.vocabulary_ = sorted(self.vocabulary_)
 
         # build term to index mapping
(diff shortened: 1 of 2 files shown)


pyserini/vectorizer/_base.py
```diff
@@ -115,7 +127,7 @@ def get_vectors(self, docids: List[str]):
                 matrix_data.append(tfidf)
 
         vectors = csr_matrix((matrix_data, (matrix_row, matrix_col)), shape=(num_docs, self.vocabulary_size))
-        return normalize(vectors, norm='l2')
```
lintool, line 118: how about `norm=l2` as an optional parameter so it doesn't break downstream code that depend on this, e.g., 20newsgroup classification?

Then caller can set `norm=None` to turn off?

lintool (commented):
Turn norm into optional parameter?

lintool:
@MXueguang One more thing... can we add a test case please? Can probably be added here: https://github.com/castorini/pyserini/blob/master/tests/test_index_reader.py

lintool (approved):
lgtm

Issue #302, opened 2021-01-06
Hybrid search: parameter design


lintool:
Pulled out comment from #299 

I was thinking, for hybrid search, an implementation with multiple subcommands, e.g.

```
python -m pyserini.hsearch --topics msmarco_passage_dev_subset 
 dense --index ... [all the usual dense options...]
 sparse --index ... [all the usual sparse options...]
 fusion [fusion options, like fusion weights, etc.]
```

See: https://stackoverflow.com/questions/10448200/how-to-parse-multiple-nested-sub-commands-using-python-argparse

This way, we don't need to re-implement options, just "dispatch" the the dense and sparse searchers.

We could design so that the dense/spare searchers take `--index` etc. options in a constructor. Alternatively, if this were Java, this would be a factory/builder class, but I don't know if this is idiomatic Python.

MXueguang:
looks good, will do that in the next iteration (after integrating query encoder). 
I feel its good to do subcommands for hybrid search like that.

MXueguang:
closed by https://github.com/castorini/pyserini/pull/330

PR #330, opened 2021-01-18
redesign hybrid search parameter

MXueguang:
re https://github.com/castorini/pyserini/issues/302
Now the hybrid search command looks like:

```bash
python -m pyserini.hsearch   dense --index msmarco-passage-tct_colbert-bf \
                                                         --batch 36 --threads 12 \
                                                         --encoded-queries msmarco-passage-dev-subset-tct_colbert  \
                             sparse --index msmarco-passage-expanded \
                             fusion --alpha 0.24 \
                             run  --topics msmarco_passage_dev_subset \
                                  --output runs/run.msmarco-passage.tct_colbert.bf.doc2queryT5.tsv \
                                  --msmarco
```

### DIFF
--- a/docs/dense-retrieval.md
+++ b/docs/dense-retrieval.md
@@ -40,10 +40,10 @@ map                     all     0.3410
 recall_1000             all     0.9618
 ```
 
-> To evaluate with on-the-fly query encoding, replace `--encoded-queries` with our pretrained encoder model
-> on [Hugging Face](https://huggingface.co/castorini/tct_colbert-msmarco/tree/main) 
-> `--encoder castorini/tct_colbert-msmarco`. The encoding will run on CPU by default. To enable GPU, add `--device cuda:0`.
-> NOTE: Using GPU query encoding will give slightly different result. (E.g. MRR @10: 0.3349694137444839)
+To evaluate with on-the-fly query encoding, replace `--encoded-queries` with our pretrained encoder model
+on [Hugging Face](https://huggingface.co/castorini/tct_colbert-msmarco/tree/main) 
+`--encoder castorini/tct_colbert-msmarco`. The encoding will run on CPU by default. To enable GPU, add `--device cuda:0`.
+NOTE: Using GPU query encoding will give slightly different result. (E.g. MRR @10: 0.3349694137444839)
 
 
 MS MARCO passage ranking task, dense retrieval with TCT-ColBERT, brute force index.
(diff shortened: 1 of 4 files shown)


lintool:
Hrm... the diffs quite big - it's difficult to work out in my head if it breaks anything... but I think we have enough test coverage to catch any obvious bugs...

For example, do these still run: https://github.com/castorini/pyserini/blob/master/docs/pypi-replication.md
(i.e., down to the exact MD5 checksums?)

MXueguang:
```
python -m unittest discover -s integrations
----------------------------------------------------------------------
Ran 27 tests in 5432.511s

OK

```

docs/dense-retrieval.md
```diff
@@ -125,22 +125,24 @@ $ tools/eval/trec_eval.9.0.4/trec_eval -c -mrecall.1000 -mmap \
 map                     all     0.3702
 recall_1000             all     0.9734
 ```
+> To evaluate with on-the-fly query encoding, replace `--encoded-queries` with our pretrained encoder model
```
lintool, line 128: Why ">" here? Applies above also.

I forgot to ask this last CR.

docs/dense-retrieval.md
```diff
@@ -125,22 +125,24 @@ $ tools/eval/trec_eval.9.0.4/trec_eval -c -mrecall.1000 -mmap \
...
-python -m pyserini.hsearch --topics msmarco_passage_dev_subset \
-                             --dense-index msmarco-passage-tct_colbert-bf \
-                             --sparse-index msmarco-passage-expanded \
-                             --encoded-queries msmarco-passage-dev-subset-tct_colbert \
-                             --output runs/run.msmarco-passage.tct_colbert.bf.doc2queryT5.tsv \
-                             --msmarco \
-                             --alpha 0.24 \
-                             --batch 36 \
-                             --threads 12
+python -m pyserini.hsearch   dense --index msmarco-passage-tct_colbert-bf \
```
lintool, line 138: Also, comment I forgot to make from last time:

Is `--index msmarco-passage-tct_colbert-bf` and `--encoded-queries msmarco-passage-dev-subset-tct_colbert` both needed? As in, aren't the two options redundant? Would you ever want to use a "mismatched" set of queries? (Especially since we already have `--topics msmarco_passage_dev_subset`.)

How about just `--index` and `--encoder`? `--index` is required, `--encoder` is optional. If `--encoder` not specified, it uses pre-encoded queries. Since you know the index (and the topics), you know what pre-encoded queries you should use...

Would simply options a bit?

(This applies to dense retrieval alone also.)

lintool (commented):
A few comments for discussion... leftover from last PR when I forgot to raise the points...

docs/dense-retrieval.md
MXueguang: I'll update this in next PR, after integrating DPR. It may have some conflicts otherwise.

lintool (approved):

PR #339, opened 2021-01-21
Updated Wikipedia DPR 100 Word Passage Indices

KaiSun314:


MXueguang (approved):
LGTM!

lintool:
@KaiSun314 did we check in the script that converts the raw corpus into something Pyserini can index?

MXueguang:
@lintool should we leave the script in pyserini or put into tools? since it is for anserini

lintool:
Actually, you're right - tools.

KaiSun314:
I created a pull request with the script in anserini-tools: https://github.com/castorini/anserini-tools/pull/28

PR #340, opened 2021-01-21
Add document retrieval for TCT_ColBERT

MXueguang:


docs/dense-retrieval.md
```diff
@@ -166,6 +166,31 @@ To evaluate with on-the-fly query encoding with our pretrained encoder model
...
+                             --threads 72
+```
+
+We can use the official TREC evaluation tool `trec_eval` to compute metrics such as recall and ndcg
+For that we first need to convert runs and qrels files to the TREC format:
+```bash
+$ python tools/scripts/msmarco/convert_msmarco_to_trec_run.py --input runs/run.dl19.tct_colbert.txt --output runs/run.dl19.tct_colbert.trec
+$ tools/eval/trec_eval.9.0.4/trec_eval -c -mrecall.100 -mndcg_cut.3 tools/topics-and-qrels/qrels.dl19-doc.txt runs/run.dl19.tct_colbert.trec
+recall_100            	all	0.3415
+ndcg_cut_3            	all	0.6800
```
lintool, line 191: I think ndcg@10 is the official metric? look in the overview paper to verify? report that?

pyserini/search/_base.py
```diff
@@ -100,6 +100,10 @@ def get_topics(collection_name):
         topics = JTopicReader.getTopicsWithStringIds(JTopics.CAR17V15_BENCHMARK_Y1_TEST)
     elif collection_name == 'car17v2.0_benchmarkY1test':
         topics = JTopicReader.getTopicsWithStringIds(JTopics.CAR17V20_BENCHMARK_Y1_TEST)
+    elif collection_name == 'dl19_doc':
+        topics = JTopicReader.getTopicsWithStringIds(JTopics.TREC2019_DL_DOC)
+    elif collection_name == 'dl19_passage':
+        topics = JTopicReader.getTopicsWithStringIds(JTopics.TREC2019_DL_PASSAGE)
```
lintool, line 106: this means we need to do a artifact deploy on the Anserini end for this to work right?

docs/dense-retrieval.md
```diff
@@ -166,6 +166,31 @@ To evaluate with on-the-fly query encoding with our pretrained encoder model
 on [Hugging Face](https://huggingface.co/castorini/tct_colbert-msmarco/tree/main) add
 `--encoder castorini/tct_colbert-msmarco`. The encoding will run on CPU by default. To enable GPU, add `--device cuda:0`.
 
+## TREC2019-DL Document Ranking (Zero Shot)
+TREC-2019 DL document ranking task, dense retrieval with TCT-ColBERT trained on MS MARCO passages, brute force index.
```
lintool, line 170: actually, TREC 2019 is for our internal sanity check - externally, I'd rather present replication on MS MARCO doc.

MXueguang: sure, I'll change to msmarco

pyserini/search/_base.py
MXueguang: yeah, we already have these on the anserini end

docs/dense-retrieval.md
MXueguang: will change to msmsrco and report the corresponding metrics

pyserini/dsearch/__main__.py
```diff
@@ -106,11 +112,27 @@ def init_query_encoder(encoder, topics_name, device):
...
+                    unique_docs = set()
+                    rank = 1
                     for idx, hit in enumerate(hits[topic]):
-                        if args.msmarco:
-                            target_file.write(f'{topic}\t{hit.docid}\t{idx + 1}\n')
+                        if args.max_passage:
+                            docid, _ = hit.docid.split(args.max_passage_delimiter)
+                            if docid in unique_docs:
+                                continue
+                            if args.msmarco:
```
lintool, line 122: I noticed this conditional: if x write this way, else write that way... is duplicated in several places. This is potentially dangerous if we change something, leading to inconsistent code paths.

Is there an obvious clean way to fix this?

lintool (approved):
one comment.

if there's a simple clean fix, do it. otherwise, this is potential future technical debt.

pyserini/dsearch/__main__.py
MXueguang: should be able to clean, like what we have done in `search.__main__.py`. But I prefer to that in a separate PR, together with the hybrid document search. `hsearch.__main__.py` will have similar things to deal with so I feel it's good to be cleaned together.

lintool: sg

PR #348, opened 2021-01-27
clean code for dsearch/hybrid search

MXueguang:
1. use query iterator for dense and hybrid search
2. clean code


lintool (approved):
lgtm

PR #363, opened 2021-02-10
store dense/sparse index saperately

MXueguang:


pyserini/util.py
```diff
@@ -150,8 +154,11 @@ def check_downloaded(index_name):
     return os.path.exists(index_path)
 
 
-def get_indexes_info():
-    df = pd.DataFrame.from_dict(INDEX_INFO)
+def get_indexes_info(dense=False):
```
lintool, line 157: How about two separate methods to be explicit? E.g., `get_dense_indexes_info` and `get_sparse_indexes_info`?

Seems cleaner than this optional arg?

lintool (commented):
thoughts?

pyserini/util.py
MXueguang: yeah, sure

lintool (approved):

PR #365, opened 2021-02-10
Cleanup of TCT-ColBERT replication guide

lintool:
@MXueguang we're missing one condition - doc, hybrid with no expansions. After this gets merged, can you add that one in?


MXueguang:
> we're missing one condition - doc, hybrid with no expansions. After this gets merged, can you add that one in?

sure

MXueguang (approved):
LGTM!

PR #376, opened 2021-02-15
Fix environment

MXueguang:
I fix faiss-cpu as 1.6.5 since 1.7.0 gives slightly different result. We can update to 1.7.0 in future release.


lintool:
How different are the results?

`faiss-cpu` doesn't install at 1.6.5 and 1.7.0 on `orca`, getting [this error](https://github.com/facebookresearch/faiss/issues/1560).

This does work, however:

```
$ sudo pip install faiss-cpu==1.6.1
```

I'm running the experiments on `orca` right now.

lintool:
@ronakice your call on this?

ronakice:
@lintool both 1.6.5 and 1.7 work on `orca` with `virtualenv`?

ronakice:
Also, sure we can stick with `1.6.5` for now.

ronakice (approved):
LGTM!

PR #400, opened 2021-03-03
Refactoring of TCT-ColBERT and DPR integration tests

lintool:
Renamed to simplify name. Separated otf vs. pre-encoded queries so we can run separately.


lintool:
Name of the technique should be sufficient for filename. Previous file naming wasn't accurate anyway, since they included hybrid searching.

MXueguang (approved):

Issue #414, opened 2021-03-12
Anyway to indicate the cache directory for downloaded indexes?


ArvinZhuang:
Hi,
Is there any way to instruct the cache directory to download the index? instead of the default .cache in the home folder?

Thanks

lintool:
Not currently... but you can change the directory in the code? PRs to generalize this welcome!

MXueguang:
Yes, It is reasonable to make cache dir configurable. Since some of our pre-built index is to large to simply store under HOME.

transformers can configure cache dir by setting environment dir 'TRANSFORMERS_CACHE'. We can implement something similar, i.e.. 'PYSERINI_CACHE'

@ArvinZhuang for now, you can modify the code here https://github.com/castorini/pyserini/blob/229647961ac973753bd0e23355850d9473f74718/pyserini/util.py#L91

ArvinZhuang:
Hi, I have modified the code and submitted a PR to solve the problem.

MXueguang:
closed by https://github.com/castorini/pyserini/pull/416
thank you @ArvinZhuang

PR #416, opened 2021-03-12
specifying cache directory to from_prebuilt_index function

ArvinZhuang:
To solve issue [#414](https://github.com/castorini/pyserini/issues/414), I added cache_dir argument to the function from_prebuilt_index to specify the wanted cache directory. The same keyword has been used in huggingface transformers to specify the directory of cached models.

### DIFF
--- a/pyserini/search/_searcher.py
+++ b/pyserini/search/_searcher.py
@@ -51,14 +51,15 @@ class SimpleSearcher:
         self.num_docs = self.object.getTotalNumDocuments()
 
     @classmethod
-    def from_prebuilt_index(cls, prebuilt_index_name: str):
+    def from_prebuilt_index(cls, prebuilt_index_name: str, cache_dir: str = None):
         """Build a searcher from a pre-built index; download the index if necessary.
 
         Parameters
         ----------
         prebuilt_index_name : str
             Prebuilt index name.
-
+        cache_dir : str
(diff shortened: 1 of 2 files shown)


MXueguang:
Hi @ArvinZhuang, Thank you for creating the PR.
However, I would still suggest configuring the cache directory via environment variable instead of argument pass.
There are two reasons.
1. the ways to configure model and index cache dirs are better to be same. I.e. we already have models's cache dir configured by `TRANSFORMERS_CACHE`, so it might be better to make index configured by env variable as well.
2. Via env variable setting, we only need to do modification and deal with cache path at _one_ place. i.e. get_cache_home(). It can reduce the number of nested argument pass and keep the code cleaner. 

a potential solution can be as follows:
```
def get_cache_home():
    if os.environ["PYSERINI_CACHE"]:
        return os.environ["PYSERINI_CACHE"]
    return os.path.expanduser(os.path.join(f'~{os.path.sep}.cache', "pyserini"))
```

open for discussion.

MXueguang (approved):
Looks good to me. Thanks for your contribution.

ArvinZhuang:
Hi @MXueguang I agree with you otherwise from_prebuilt_index in dsearch also needs to be changed.
I tried what you suggest and it works fine.

PR #419, opened 2021-03-12
fix download cache and test

MXueguang:


tests/test_index_download.py
```diff
@@ -0,0 +1,40 @@
...
+import shutil
+from pyserini import search
+from pyserini.util import get_cache_home
+
+
+class TestIndexDownload(unittest.TestCase):
+
+    def test_default_cache(self):
+        os.unsetenv('PYSERINI_CACHE')
+        search.SimpleSearcher.from_prebuilt_index('robust04')
```
lintool, line 28: I think it'll look cleaner if you import `SimpleSearcher` directly so you don't need the `search.` prefix.

lintool (approved):
Minor nit.

Issue #396, opened 2021-02-26
topics and encoded-queries mapping 


MXueguang:
Currently, we are attaching encoded-queries to topic name. i.e. `dpr-nq-test` is going to use `query-embedding-nq-test-dpr-multi`.
It worked since we were having 1-to-1 mapping between topics and encoded-queries.

But now we have multiple encoders for the same topic. I.e. DPR and ANCE both have encoder for `dpr-nq-test`.

maybe add back the option `--encoded-queries`?

lintool:
Related:

```
$ ls ~/.cache/pyserini/queries/
query-embedding-curated_trec-test-dpr-multi-20210129-3a276c.e4c6e96e9f4a733c32941a6288b92316	query-embedding-squad-test-dpr-multi-20210129-3a276c.e1c41fe5261c8b09c442ee57f525be53
query-embedding-msmarco-passage-dev-subset-20210112-be7119.c9398a713ff13e1780ea2aa2747deb2f	query-embedding-trivia_qa-test-dpr-multi-20210129-3a276c.c7b3c2a415ce9ace27da13432a817824
query-embedding-msmarco-passage-dev-subset-20210115-cd5034.0e448fda77344e7b892e861f416a8870	query-embedding-wq-test-dpr-multi-20210129-3a276c.8b8a9054ec1e6ce57446acdc133fecbe
query-embedding-nq-test-dpr-multi-20210129-3a276c.dd92355c7661cf23a802e793cd5a73ff
```

We should stick to the convention of `model-corpus` so that the queries sort nicely... so `dpr` should come first...

MXueguang:
closed by https://github.com/castorini/pyserini/pull/458

PR #420, opened 2021-03-12
add ANCE encoded queries and update ANCE doc

MXueguang:


pyserini/encoded_query_info.py
```diff
@@ -29,7 +39,27 @@
         "total_queries": 3610,
         "downloaded": False
     },
-    "dpr-trivia_qa-dev-multi": {
+    "dpr-nq-dev-ance-multi": {
```
lintool, line 42: Related: https://github.com/castorini/pyserini/issues/396#issuecomment-789761717

i.e., consistent naming... but doesn't have to be in this PR.

lintool (approved):

pyserini/encoded_query_info.py
MXueguang: Let's fix in this PR.
will update soon

MXueguang: fixed the encoded query by moving "dpr" to front

PR #421, opened 2021-03-13
add SBERT passage encode script

MXueguang:


lintool (approved):

lintool:
hey @MXueguang - at some point, we might want to write a more general encoder script that spans multiple models...

MXueguang:
yeah we should. 
Once we integrated SBERT i think?

lintool:
Not a big rush because these are util scripts that aren't pushed to pip, so don't really need to be clean and polished...

PR #428, opened 2021-03-15
sbert query encoder

MXueguang:


docs/experiments-sbert.md
```diff
@@ -0,0 +1,35 @@
+# Pyserini: Replicating S-BERT MSMARCO Results
```
lintool, line 1: Do we want to use "SBERT" (without dash) to be consistent with everything else?

lintool (approved):
lgtm, minor comment.

docs/experiments-sbert.md
MXueguang: sure

Issue #432, opened 2021-03-16
Integrate Margin-MSE trained DistilBERT dense retrieval model


sebastian-hofstaetter:
Hi,

I just published our Margin-MSE ensemble-trained, DistilBERT-based checkpoint for dense passage retrieval here: https://huggingface.co/sebastian-hofstaetter/distilbert-dot-margin_mse-T2-msmarco

Would be cool if we could integrate that model in pyserini! I created a minimal usage example here: https://github.com/sebastian-hofstaetter/neural-ranking-kd/blob/main/minimal_bert_dot_usage_example.ipynb

Best,
Sebastian

MXueguang:
@lintool I think this issue can be close by https://github.com/castorini/pyserini/pull/433?
CC: @sebastian-hofstaetter

lintool:
👍

PR #433, opened 2021-03-17
Integrate Margin-MSE trained DistilBERT dense retrieval model

MXueguang:
re https://github.com/castorini/pyserini/issues/432

get:
```
#####################
MRR @10: 0.32505867103288255
QueriesRanked: 6980
#####################
map                     all     0.3308
recall_1000             all     0.9553
```

expect:
```
MRR@10: 0.323
recall_1000: 0.957
```


lintool:
The expected effectiveness is:

> Margin-MSE BERT_Dot (Retrieval) .323

Is that correct, @sebastian-hofstaetter ?

pyserini/dsearch/_dsearcher.py
```diff
@@ -192,6 +192,36 @@ def encode(self, query: str):
             return super().encode(query)
 
 
+class AutoQueryEncoder(QueryEncoder):
```
lintool, line 195: Is the intention here that AutoQueryEncoder will replace other model specific ones?

docs/experiments-ensemble_kd.md
```diff
@@ -0,0 +1,35 @@
+# Pyserini: Replicating Ensemble_KD MSMARCO Results
```
lintool, line 1: Let's retain the original model name `distilbert-dot-margin_mse-T2` to be consistent, and call it

And use the more human readable name "DistilBERT KD" in the appropriate context, e.g., 

"Pyserini: Replicating DistilBERT KD Results" <--- leave out MS MARCO to be consistent with TCT-ColBERT

Also add link from README (for ANCE also)

lintool (changes requested):

sebastian-hofstaetter:
> The expected effectiveness is:
> 
> > Margin-MSE BERT_Dot (Retrieval) .323
> 
> Is that correct, @sebastian-hofstaetter ?

Yes, although in the paper I used the large 49K query DEV set (so a minimal change like 0.002 is expected if you use the 7K query DEV set). So I think these numbers are ok 👍

pyserini/dsearch/_dsearcher.py
MXueguang: yes, the most common dense encoder is taking CLS from the encoder model
We can try to group these models as Auto I think?

docs/experiments-ensemble_kd.md
MXueguang: updated, I'll re-upload the index with name changed later

docs/experiments-distilbert_kd.md
```diff
@@ -0,0 +1,35 @@
+# Pyserini: Replicating DistilBERT KD Results
+
+## Dense Retrieval
+
+Dense retrieval with distilbert-dot-margin_mse-T2, brute-force index:
```
lintool, line 5: If it's the literal model name, should be in monospace found `distilbert-dot-margin_mse-T2`, otherwise, call it DistilBERT KD in normal font.

pyserini/dsearch/_dsearcher.py
lintool: File an issue, doesn't need to be part of this PR.

scripts/ensemble_kd/encode_corpus_msmarco_passage.py
```diff
@@ -0,0 +1,62 @@
+import argparse
```
lintool, line 1: rename directory to be consistent...

lintool (approved):
minor nits - please fix and merge.

pyserini/dsearch/_dsearcher.py
MXueguang: > File an issue, doesn't need to be part of this PR.

Should I rename it to something like `DistilBertQueryEncoder` in this PR?

lintool: Nah... I think it's okay for master to be in a slightly inconsistent state as long as we fix if before the next release.

PR #434, opened 2021-03-17
script for hybird result externally

MXueguang:
script for hybrid results 
cc: @ronakice


lintool (changes requested):
scripts should be named as verb phrases... i.e., "do this"

Maybe something like `compute_hybrid_results`? or `perform_hybrid_retrieval` or something along those lines?

lintool:
This is something that @ronakice requested? I'm missing the context here, in the sense that `-m search` already does hybrid?

lintool (approved):

ronakice (approved):
LGTM!

PR #447, opened 2021-04-01
Add tokenization tests

lintool:
Sanity checks our doc2query model.

With this, I want to revisit this issue: https://github.com/castorini/anserini/issues/1123 - let's try to make sure everything is in place, from the ground up.


rodrigonogueira4:
Sorry, why do we need tests for the tokenizer if HF's already does them?

rodrigonogueira4 (approved):

lintool:
Discussed over DM.

PR #453, opened 2021-04-02
script to convert json file to have contents tokenized

stephaniewhoo:
do you want it to be called as a module or just leave it as a script for now?


pyserini/convert_to_tokenized_json.py
```diff
@@ -0,0 +1,45 @@
+#
```
lintool, line 1: how about naming this `tokenize_json_collection`?

pyserini/convert_to_tokenized_json.py
```diff
@@ -0,0 +1,45 @@
...
+            contents = fdict['contents']
+            tok = tokenizer.tokenize(contents)
+            fdict['contents']=tok
+            out_f.write(json.dumps(fdict) + '\n')
+            if (i%10000==0):
+                print(f'Converted {i:,} docs, writing into file {args.output_json}')
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--input_json", type=str, help='Input Json file', required=True)
```
lintool, line 39: Can we just have `--input` and `--output`? Make it smart - if input is a file, just convert a single file. If input is a directory, converts all json files in that direction.

pyserini/convert_to_tokenized_json.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+#
+
+from transformers import BertTokenizer, T5Tokenizer
+import argparse
+import json
+
+def main(args):
+    if (args.tokenizer == 'bert'):
+        tokenizer = BertTokenizer.from_pretrained('bert-base-uncased')
```
lintool, line 23: okay to assume user specifies fully-qualified name, e.g.,
```
tokenizer = BertTokenizer.from_pretrained(args.tokenizer)
```

But make arg optional, e.g., default to bert-base-uncased?

lintool (changes requested):
Also try actually tokenizing MS MARCO passage... I want to have a sense of how long it'll take?

pyserini/convert_to_tokenized_json.py
stephaniewhoo: Yes, I will make it optional. 
But if we want t5 tokenizer, it is called from `**T5Tokenizer.**from_pretrained('castorini/doc2query-t5-base-msmarco')`. still cannot get rid of if statement?

lintool: ah, I see... maybe like a comprise and search for a substring of {bert, t5} in the if?

lintool:
@stephaniewhoo can we also have test cases, for both the per-file case and the entire-directory case?

stephaniewhoo:
> Also try actually tokenizing MS MARCO passage... I want to have a sense of how long it'll take?

It takes 7997.244991540909s so **~2h** to tokenize msmarco passage collection_jsonl on orca.
I did not do multi-thread. Could save time by doing this, but one thread is like 2 hours

lintool (approved):

PR #455, opened 2021-04-04
update integrate tests

MXueguang:


lintool (approved):
You removed `test_tct_colbert_encoded.py` but we're going to at least do some sanity checking of encoded queries, right?

Like download them and make there there are the right number of queries?

MXueguang:
yeah, will add in next PR

Issue #456, opened 2021-04-04
merge all query encoding scripts into one


MXueguang:
now we have multiple scripts to create encoded queries.
but we can simply put them into one script and use the encoder classes.

MXueguang:
closed by https://github.com/castorini/pyserini/pull/457

PR #457, opened 2021-04-04
encode queries script

MXueguang:
re https://github.com/castorini/pyserini/issues/456


scripts/encode_queries.py
```diff
@@ -0,0 +1,48 @@
+import argparse
```
lintool, line 1: Add boilerplate.

lintool (approved):

PR #465, opened 2021-04-07
CLEF HEALTH

spacemanidol:


docs/experiments-clefhealth-doc.md
```diff
@@ -0,0 +1,98 @@
...
+This guide contains instructions for running BM25 baselines on the [CLEF eHealth Lab Series Adhoc Information Retrieval](https://clefehealth.imag.fr/?page_id=610).
+
+## Data Prep
+
+The guide requires the [development installation](https://github.com/castorini/pyserini/#development-installation) for additional resource that are not shipped with the Python module; for the (more limited) runs that directly work from the Python module installed via `pip`, see [this guide](pypi-reproduction.md).
+
+We're going to use the repository's root directory as the working directory.
+First, we need to download and extract the CLEF document dataset:
+
+```
```
lintool, line 12: bash

docs/experiments-clefhealth-doc.md
```diff
@@ -0,0 +1,98 @@
...
+mkdir collections/clef_health
+wget -c https://www.dropbox.com/s/ixnqt33u5xeelth/clef2018collection.tar.gz -P collections/clef_health
+```
+
+There's is a need to decompress the file as the corpus is not in proper answerini format. Its worth noting that the full corpus is over 500gb and needs to be convered to answerini format so make sure your machine has at least 1tb of free space.
```
lintool, line 17: answerini typo

docs/experiments-clefhealth-doc.md
```diff
@@ -0,0 +1,98 @@
...
+```bash
+
+```
+Since the CLEF dataset is not in Answerini format you need to conver it using [a simple conversion script](https://github.com/spacemanidol/pyserini/blob/master/scripts/convert_clef_to_pyserini.py) which is run like this:
```
lintool, line 21: empty command here?

lintool (changes requested):
Is this a reasonable baseline? What are other effectiveness values reported by participants, for example - would be nice to have some context...

Issue #382, opened 2021-02-18
Integrate ANCE into pyserini


MXueguang:
paper: https://arxiv.org/pdf/2007.00808.pdf
https://github.com/microsoft/ANCE provided encoder checkpoints

MXueguang:
we will integrate both msmarco and NQ from ANCE (ance-msmarco, ance-nq-trivia).

For ANCE-NQ-Trivia, I converted their checkpoint into hugging face model, which can fit into our current DPR encoder directly.
For ANCE-Msmarco, we will create a new class with some hack on their code.

They didn't provide prebuilt indexes. We will encode the corpus by ourself, (writing encoding script).

MXueguang:
replicated ANCE for NQ by 
1. converting their ckpt into hugging face DPR encoder.
2. run our own passage embedding script to generate embedding
3. run our current dpr pipeline with ANCE-NQ ckpt.
Got result on nq-test as below:
```
Top20	accuracy: 0.8224376731301939
Top100	accuracy: 0.8786703601108034
```

Expected:
Top20:   82.1
Top100: 87.9

converted model ckpt:
ance-dpr-question-encoder-multi: https://www.dropbox.com/s/pps5rzzn4ynh3x3/ance-dpr-question_encoder-multi.tar.gz
ance-dpr-context-encoder-multi: https://www.dropbox.com/s/diq5y8dd1bytje1/ance-dpr-context_encoder-multi.tar.gz

MXueguang:
replicated ANCE for MS MARCO passage by
1. create new class `AnceEncoder` on top of `transformers.RobertaModel`
2. convert their ckpt so that can directly load into `AnceEncoder`, i.e. `AnceEncoder.from_pretrained('ckpt_path')`
3. run our passage embedding script to generate embedding.
4. add AnceQueryEncoder our current msmarco-passage dense pipeline.

Got result on msmarco-passage as below:
```
#####################
MRR @10: 0.3301838017919672
QueriesRanked: 6980
#####################
```

expected: `MRR@10: 0.330`

converted model ckpt:
query_encoder = ctx_encoder: https://www.dropbox.com/s/u02glpszk3jv6ws/ance-msmarco-passage-encoder.tar.gz

MXueguang:
ANCE-Msmarco-Passage and ANCE-NQ/TriviaQA has been replicated. Updating docs.

ANCE for msmarco document ranking can be integrated as well. 
[...]

PR #473, opened 2021-04-11
Fix quotes in ignore_keys for ANCE Encoder

vrdn-23:
PR to fix the issue mentioned in #382 [comments](https://github.com/castorini/pyserini/issues/382#issuecomment-817376103)


MXueguang (approved):
LGTM, Thank you!

PR #488, opened 2021-04-20
fix encoded queries inconsistancy

MXueguang:
update to latest Anserini fat jar file and then regenerate encoded queries for failed tests


MXueguang:
I was using the Anserini version which read CSV lines by simply split the line string (i.e. the very first version of https://github.com/castorini/anserini/commit/745933ba6f6f30ba372bca1182dbd7cc493fef80#diff-528f2b8f9c3d4fa3c55782b2[…]b85f64b6a1d59d5eacef065fb7fR40).
Newer version of Anserini replaced it by CSVParser.

lintool (approved):

PR #493, opened 2021-04-22
DPR results reproduction

lintool:
I successfully reproduced all results on my mac, both otf and encoded queries.

Note there is one very minor discrepancy -
for WQ, DPR, top-20 - the table says 75.0 (as does the paper https://arxiv.org/pdf/2104.05740.pdf)

But the replication result is:

```
$ python -m pyserini.eval.evaluate_dpr_retrieval --retrieval runs/run.dpr.wq-test.multi.bf.json --topk 20 100
Top20  accuracy: 0.7505
Top100 accuracy: 0.8297
```

Which should get rounded to 75.1.

Let's maybe fix in next draft of paper?


MXueguang:
sure

MXueguang:
emm, has 0.7505 been rounded already?
If it was from 0.75049xxxx we should still round it as 0.750?

lintool:
> emm, has 0.7505 been rounded already?
> If it was from 0.75049xxxx we should still round it as 0.750?

haha, yea, 0.75049 should be rounded to 0.750, but 0.7505 should be rounded to 0.751.

So, no change needed in paper, I guess...

MXueguang:
yeah I was 0.750492125984252

MXueguang (approved):
LGTM

PR #494, opened 2021-04-22
read topic from file if not exist

MXueguang:


pyserini/search/_base.py
```diff
@@ -142,6 +142,16 @@ def get_topics(collection_name):
         topics = JTopicReader.getTopicsWithStringIds(JTopics.DPR_SQUAD_TEST)
     elif collection_name == 'dpr-curated-test':
         topics = JTopicReader.getTopicsWithStringIds(JTopics.DPR_CURATED_TEST)
+    elif os.path.exists(collection_name):
```
lintool, line 145: hrm, the issue with this is that we break the qid order... can we do the thing with ms marco where we preserve the iteration order?

lintool (commented):
thoughts?

pyserini/search/_base.py
MXueguang: pass a new option, say `args.order` to use predefined query order?
https://github.com/castorini/pyserini/blob/df569681b70b08b15d37ae2e80b088c238cc3353/pyserini/search/__main__.py#L231

another choice maybe require the topic file name contains specific prefix or suffix

lintool: No need for option... I think it should just be the default impl.

I can't think of an instance where you'd _not_ want this... can you?

MXueguang: but how do we know which topic the topic file belongs to _by default_?
- msmarco-doc-dev
- msmarco-doc-test
- msmarco-passage-subset-dev
- msmarco-passage-subset-test

MXueguang: from file name?

lintool: yea, if the user specifies a file, we should respect the line ordering.

if it's a "known" collection, we do whatever the default is...

lintool (approved):

PR #504, opened 2021-04-25
fix typo in ance doc

MXueguang:
re https://github.com/castorini/pyserini/issues/503


lintool:
I will fix in my PR coming up soon.

PR #509, opened 2021-04-27
[WIP]add pretokenized option to Pyserini search

stephaniewhoo:


lintool (changes requested):
Hi @stephaniewhoo can we add a test case, something like what we have for anserini?
https://github.com/castorini/anserini/blob/master/.travis.yml#L20

The CACM collection is really small, so should be quick - both indexing and search, both "normal" tokenization and pre-tokenized.

Thanks!

pyserini/search/__main__.py
```diff
@@ -185,6 +187,16 @@ def define_search_args(parser):
     if args.dismax:
         query_generator = JDisjunctionMaxQueryGenerator(args.tiebreaker)
         print(f'Using dismax query generator with tiebreaker={args.tiebreaker}')
+    
+    if args.tokenizer != None:
+        analyzer = JWhiteSpaceAnalyzer()
+        searcher.set_analyzer(analyzer)
+        print(f'Using whitespace analyzer because of pretokenized topics')
+        if ('bert' in args.tokenizer):
+            tokenizer = BertTokenizer.from_pretrained('bert-base-uncased')
```
lintool, line 196: actually, let the arg be the actual name of tokenizer?

e.g., `--tokenizer bert-base-uncased`

lintool:
Just as an FYI, I'm reorganizing the integration tests per #515 - let me get that merged first (PR coming soon) and then you can build on top.

PR #510, opened 2021-04-27
MS MARCO doc/passage replication

saileshnankani:
Ran both experiments on MacOS. All numbers obtained were exactly the same as in the provided README. 

## Challenges 
- Downloads were quite slow using wget, so I had to manually download from the alternative mirror (dropbox), which was much easier. 
- I had to additionally install [sentence piece](https://github.com/google/sentencepiece#installation) which is not mentioned in the [README.md](https://github.com/castorini/pyserini/blob/master/README.md)


lintool:
Good call on sentencepiece - opened #512 to resolution.

lintool (approved):

PR #524, opened 2021-04-30
Update experiments-msmarco-doc.md

andrewyguo:


lintool:
Can you combine with #523 and give a few brief comments on the reproduction? I.e., I assume everything matched exactly? What OS, etc.?

PR #527, opened 2021-05-01
Refactoring and cleanup of sparse test cases

lintool:
Primarily to make sure you can run the test cases from inside an IDE (fixed relative paths).


stephaniewhoo (approved):
pretokenization part lgtm

PR #537, opened 2021-05-03
Add resort option to TrecRun

yuki617:


lintool (changes requested):
Change to just `--resort` as opposed to `--resort True`?

This should be an option that takes zero args.

lintool (approved):

PR #554, opened 2021-05-11
Move Trec covid unittest and reranker script into right place

yuki617:


lintool (changes requested):
Can we rename `scripts/classifier_prf/trec-covid-ranker.py` to `scripts/classifier_prf/rerank_trec_covid.py`?

In general, we've been consistent in using underscore for names, and naming scripts as verb phrases, i.e., "do_this_thing".

lintool (approved):

PR #561, opened 2021-05-12
Create experiments-tripclick-doc.md

olesota:
Reproduction of BM25 results on TripClick IR collection


lintool:
Thanks for your contribution! Let's get it merged in and I'll get someone from my end to reproduce.

lintool (approved):

Issue #560, opened 2021-05-12
Missing bindings for qrels


lintool:
There are missing bindings for qrels - for example, I can't get TREC 20 DL qrels. They're in Anserini, just not properly exposed in Pyserini.

We should add test cases:
https://github.com/castorini/pyserini/blob/master/tests/test_load_topics.py

Rename `test_load_topics.py` to `test_load_topics_and_qrels.py` and make sure each topics/qrels pair works...

MXueguang:
closed by https://github.com/castorini/pyserini/pull/562 https://github.com/castorini/pyserini/pull/594

Issue #559, opened 2021-05-12
get_topics and get_qrels fails silently


lintool:
```python
from pyserini.search import get_topics, get_qrels

topics = get_topics('blah')
qrels = get_qrels('blah')
```

Currently fails silently. It shouldn't - throw an exception or something?

MXueguang:
> fails silently
it returns empty dict right?
yes we should throw exceptions

MXueguang:
closed by https://github.com/castorini/pyserini/pull/562

PR #562, opened 2021-05-12
fix topic/qrels & update tests

MXueguang:
re https://github.com/castorini/pyserini/issues/560 and https://github.com/castorini/pyserini/issues/559


lintool (approved):
👍

PR #574, opened 2021-05-16
Add KILT downstream eval script and 100 word passage script

yuxuan-ji:
This PR adds the following:
- KILT downstream evaluation script (and also update retrieval evaluation script) 
- Script to convert KILT wiki into 100 words passages for indexing. This is to replicate their [BM25 run](https://arxiv.org/pdf/2101.00117.pdf) which they haven't released the index/script used
- `rouge>=1.0.0`  to `requirements.txt`. This is because their script has a dependency to this lib. (Note: should we be adding this to requirements.txt or asking users to install with `pip` prior to running eval? If we were to integrate other challenges we could run into dependency conflicts (imaginary e.g. KILT requires rouge=1.0.0 while OTHER_CHALLENGE requires rouge=2.0.0 -> for rouge it's probably fine because it seems stable at `1.0.0` for now)

Testing:
I created a dummy run to verify the downstream eval script works:
image

Note that we currently only generate retrieval runs, not downstream runs, which is why the scores are 0

### DIFF
--- a/pyserini/eval/evaluate_kilt_downstream.py
+++ b/pyserini/eval/evaluate_kilt_downstream.py
@@ -17,7 +17,7 @@ from pyserini.query_iterator import KiltQueryIterator
 # import kilt.eval_retrieval as retrieval_metrics
 # from kilt import kilt_utils
 # With the following import:
-import evaluate_kilt_retrieval as retrieval_metrics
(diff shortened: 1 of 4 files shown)


yuxuan-ji:
closing for now since we are only focusing on retrieval atm

PR #569, opened 2021-05-14
Documenting ColbertTCT reproduction

isoboroff:
I reproduced the Colbert results on a dual-Xeon (E5-2697), 4x NVIDIA Tesla V100, 128GB RAM, 150 GB local SSD storage.  This reproduction was done using PyPi pyserini v0.12.0.  I was able to confirm that enabling GPU allowed the query vectors to be encoded on the fly, but this did not actually save any time.

```
$ python --version
Python 3.8.5

$ pip freeze | egrep 'transform|torch|faiss'
faiss-cpu==1.7.0
faiss-gpu==1.7.0
sentence-transformers==1.1.1
torch==1.8.1
torchvision==0.9.1
transformers==4.4.0

$ python -m pyserini.dsearch --topics msmarco-passage-dev-subset \
                              --index msmarco-passage-tct_colbert-bf \
                              --encoded-queries tct_colbert-msmarco-passage-dev-subset \
                              --batch-size 36 \
                              --threads 12 \
                              --output runs/run.msmarco-passage.tct_colbert.bf.tsv \
                              --output-format msmarco
Using pre-defined topic order for msmarco-passage-dev-subset
Attempting to initialize pre-encoded queries tct_colbert-msmarco-passage-dev-subset.
Downloading index at https://github.com/castorini/pyserini-data/raw/main/encoded-queries/query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.tar.gz...
query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.tar.gz: 0.00B [00:query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.tar.gz:   0%| | 8.query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.tar.gz: 19.2MB [00:00, 26.0MB/s]
Extracting /local_home/soboroff/.cache/pyserini/queries/query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.tar.gz into /local_home/soboroff/.cache/pyserini/queries/query-embedding-tct_colbert-msmarco-passage-dev-subset-20210419-9323ec.b2fe6494241639153f26cc61acf3b39d...
Initializing tct_colbert-msmarco-passage-dev-subset...
Attempting to initialize pre-built index msmarco-passage-tct_colbert-bf.
[...]

PR #575, opened 2021-05-17
Update TCT ColBERT repro log

JMMackenzie:
Notes: Used Pyserini (0.12.0) via Pip. Only examined passages; CPU running over pre-built indexes; reproduced both brute force and HNSW runs.

Package version info:
```
faiss-cpu==1.7.0
torch==1.7.1
transformers==4.0.0
```


lintool:
Hey @JMMackenzie if you used PyPI version, would you mind explicitly saying so? Like in #569 

Can you also share your PyTorch, FAISS, and transformers versions? E.g., `$ pip freeze | egrep 'transform|torch|faiss'`

Thanks!

JMMackenzie:
@lintool, I have edited my comment above to capture that info, and modified the repro logs to explicitly point to the `0.12.0` release on PyPi. Let me know if there's anything else you need.

lintool (approved):

Issue #573, opened 2021-05-16
add detailed installation guide


MXueguang:
Now we have a general installation guide in README.md with all related info.
But we also need a more "step by step" setup guide for Windows (or Linux as well).

lintool:
Agreed, but we should have this linked off README to a separate page.

MXueguang:
yeah agreed!

lintool:
This is a reasonable start: https://github.com/castorini/pyserini/blob/master/docs/installation.md

Closing for now.

PR #586, opened 2021-05-19
Add doc for tuna

saileshnankani:
Adding some documentation to install pyserini on tuna/ocra  as part of #573

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,14 +73,7 @@ python -m unittest
 
 Assuming all tests pass, you should be ready to go!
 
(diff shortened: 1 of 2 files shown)


lintool (changes requested):
Hey @saileshnankani thanks for doing this! It probably should be part of #573 - can you create a separate file, maybe something like `docs/installation.md` and drop this note in there?

From main README, maybe something like "See additional detailed instructions for Waterloo development machines [here](...)"

lintool (approved):

Issue #547, opened 2021-05-06
init dense searcher using encoder_name string


MXueguang:
Currently we need pass dense searcher an encoder object when init searcher.
```python
encoder = TctColBertQueryEncoder('castorini/tct_colbert-msmarco')
searcher = SimpleDenseSearcher.from_prebuilt_index(
    'msmarco-passage-tct_colbert-hnsw',
    encoder
)
```

I think it is better we _also_ support passing model_name string when init searcher?
i.e. the code below should work too
```python
searcher = SimpleDenseSearcher.from_prebuilt_index(
    'msmarco-passage-tct_colbert-hnsw',
    'castorini/tct_colbert-msmarco'
)
```

lintool:
Agreed!

lintool:
@MXueguang I think this can be closed?

PR #596, opened 2021-05-20
init encoder from str

MXueguang:
re https://github.com/castorini/pyserini/issues/547


lintool (approved):
Might want to downcase somewhere so that both `ANCE` and `ance` will work?

Issue #615, opened 2021-05-26
Unit test temp_dir/ clean-up


lintool:
After running unit tests, `python -m unittest`, there seems to be `temp_dir/` that isn't cleaned up properly.

lintool:
Also, `test_run.msmarco-doc.passage.ance-maxp.otf.txt` seems to be left over from integration tests.

MXueguang:
closed by https://github.com/castorini/pyserini/pull/616

PR #616, opened 2021-05-26
fix temp dir

MXueguang:


tests/test_load_qrels.py
```diff
@@ -241,6 +241,7 @@ def test_trec2018_bl(self):
 
     def test_trec2019_bl(self):
         qrels_path = search.get_qrels_file('trec2019-bl')
+        print(qrels_path)
```
lintool, line 244: Probably don't need this debug in the final version?

lintool (approved):
Actually, it's not obvious to me the source of the issue... can you please explain?

MXueguang:
the issue is because other tests sets custom PYSERINI_CACHE as temp_dir but didn't unset.
`test_load_topics_qrels` is suppose to use default `~/.cache` as PYSERINI_CACHE

MXueguang:
fix https://github.com/castorini/pyserini/issues/615

PR #617, opened 2021-05-26
add tasb msmarco dev subset reproduce

jacklin64:
The reproduction basically follow distill bert KD. Currently, the score is a bit lower than the numbers reported in the hub (MRR@10 .344 vs .347) even when I use fp16 mixed precision. I guess the results are acceptable and we can wait until Sebastian has time to release his code to make a comparison? I've already had the pre-computed index and query embeddings in Tuna. Shall I upload to dropbox ?

### DIFF
--- a/docs/experiments-distilbert_tasb.md
+++ b/docs/experiments-distilbert_tasb.md
@@ -1,0 +1,50 @@
+# Pyserini: Reproducing DistilBERT KD TASB Results
+
+This guide provides instructions to reproduce the DistilBERT KD TASB dense retrieval model on the MS MARCO passage ranking task, described in the following paper:
+
+> Sebastian Hofstätter, Sheng-Chieh Lin, Jheng-Hong Yang, Jimmy Lin, Allan Hanbury. [Efficiently Teaching an Effective Dense Retriever with Balanced Topic Aware Sampling
+.](https://arxiv.org/abs/2104.06967) SIGIR2021. 
+
+Starting with v0.12.0, you can reproduce these results directly from the [Pyserini PyPI package](https://pypi.org/project/pyserini/).
+Since dense retrieval depends on neural networks, Pyserini requires a more complex set of dependencies to use this feature.
+See [package installation notes](../README.md#package-installation) for more details.
+
+Note that we have observed minor differences in scores between different computing environments (e.g., Linux vs. macOS).
+However, the differences usually appear in the fifth digit after the decimal point, and do not appear to be a cause for concern from a reproducibility perspective.
+Thus, while the scoring script provides results to much higher precision, we have intentionally rounded to four digits after the decimal point.
+
+Dense retrieval, with brute-force index:
+
+```bash
+$ python -m pyserini.dsearch --topics msmarco-passage-dev-subset \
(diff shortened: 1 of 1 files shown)


MXueguang:
@jacklin64 could you move the prebuild index/encoded-queries to a shared folder on tuna? I'll upload it to my dropbox, manage together with other indexes.

docs/experiments-distilbert_tasb.md
```diff
@@ -0,0 +1,50 @@
+# Pyserini: Reproducing DistilBERT KD TASB Results
+
+This guide provides instructions to reproduce the DistilBERT KD TASB dense retrieval model on the MS MARCO passage ranking task, described in the following paper:
+
+> Sebastian Hofstätter, Sheng-Chieh Lin, Jheng-Hong Yang, Jimmy Lin, Allan Hanbury. [Efficiently Teaching an Effective Dense Retriever with Balanced Topic Aware Sampling
+.](https://arxiv.org/abs/2104.06967) SIGIR2021.
```
lintool, line 6: move to previous line.

docs/experiments-distilbert_tasb.md
```diff
@@ -0,0 +1,50 @@
+# Pyserini: Reproducing DistilBERT KD TASB Results
+
+This guide provides instructions to reproduce the DistilBERT KD TASB dense retrieval model on the MS MARCO passage ranking task, described in the following paper:
+
+> Sebastian Hofstätter, Sheng-Chieh Lin, Jheng-Hong Yang, Jimmy Lin, Allan Hanbury. [Efficiently Teaching an Effective Dense Retriever with Balanced Topic Aware Sampling
+.](https://arxiv.org/abs/2104.06967) SIGIR2021. 
+
+Starting with v0.12.0, you can reproduce these results directly from the [Pyserini PyPI package](https://pypi.org/project/pyserini/).
```
lintool, line 8: "Starting with v0.12.0"... not true, rewrite...

lintool (changes requested):
We need a corresponding integration test?

Also, really, we don't need to write any new actual code?

If really so, cool! Shows that @MXueguang designed the package well!

docs/experiments-distilbert_tasb.md
jacklin64: I don't understand what it means. The current version is v0.12.0 and I use this version to run the experiment.

lintool: Yes, but the pre-built index isn't part of v0.12.0, right? So we'll need to publish a new artifact before the pip version will work fully?

lintool: As in, you need a dev installation, right?

jacklin64: ah, understood. So I remove this part temporarily.

lintool: yea, we can update to v0.13.0 or something when next package gets published...

MXueguang (approved):
lgtm, I just uploaded encoded queries and prebuilt on dropbox.
we can merge this first, I'll reproduce based on this doc. 
@lintool

lintool (approved):
sgtm

PR #620, opened 2021-05-26
Tokenizer test cases for 9 African Languages

### LINKED ISSUE #578: Tokenization test cases for African languages
Hey @keleog - just like how @crystina-z added tokenization test cases for a variety of languages - can you do the same for the African languages you're working with? Here: https://github.com/castorini/pyserini/blob/master/tests/test_tokenization.py#L138

This will give us a sense of how to proceed with search.

keleog:
Added mBERT and XLM-R tokenizer test cases for the following African languages:

1. Amharic
2. Afaan Oromooo
3. Hausa
4. Igbo
5. Nigerian Pidgin
6. Somali
7. Swahili
8. Tigrinya
9. Yoruba

Fixes #578 @

### DIFF
--- a/tests/test_tokenization.py
+++ b/tests/test_tokenization.py
@@ -286,6 +286,294 @@ class TestTokenization(unittest.TestCase):
         # adversarial
         tokens = tokenizer.tokenize('প্রতিকূল')
         self.assertEqual(['প্রতি', '##ক', '##ূ', '##ল'], tokens)
+    
+    def test_bert_base_multilingual_am(self):
+        """
+        amharic
+        """
+        tokenizer = BertTokenizer.from_pretrained('bert-base-multilingual-uncased')
+
+        tokens = tokenizer.tokenize('የሽፋኑ ርዕሰ ጉዳይ የሞቱ ሰዎች ይነሳሉ')
+        self.assertEqual(['[UNK]', '[UNK]', '[UNK]', '[UNK]', '[UNK]', '[UNK]'], tokens)
+
+        tokens = tokenizer.tokenize('የሽፋኑ')
+        self.assertEqual(['[UNK]'], tokens)
+
+        tokenizer = BertTokenizer.from_pretrained('bert-base-multilingual-cased')
+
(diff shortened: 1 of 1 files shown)


keleog:
All methods are named with the ISO 639-1 language codes.

tests/test_tokenization.py
```diff
@@ -286,6 +286,294 @@ def test_bert_base_multilingual_bn(self):
...
+        tokens = tokenizer.tokenize('upasuaji')
+        self.assertEqual(['▁up', 'asu', 'aji'], tokens)
+
+    def test_bert_base_multilingual_ti(self):
+        """
+        Tigrinya
+        """
+        tokenizer = BertTokenizer.from_pretrained('bert-base-multilingual-uncased')
+
+        tokens = tokenizer.tokenize('ስርዓተ ቀብሪ ኢንጂነር ስመኘው በቀለ ትማሊ ተፈፂሙ')
```
lintool, line 520: @keleog all `[UNK]`? So what do we do here? Like, how is any downstream NLP task possible then...?

lintool:
This is good. Based on these results, we can tell if simple whitespace tokenization for some languages will give "okay-ish" results, right?

tests/test_tokenization.py
keleog: We could adapt the tokenizer to the language via various methods (eg. https://www.aclweb.org/anthology/2020.findings-emnlp.240/).

Otherwise, we cannot use it on any downstream task. E.g. Recall our Japanese NER problem, and mBERT scores 0 on Amharic NER https://arxiv.org/abs/2103.11811

lintool: Yup, good to know at least!

lintool (approved):

keleog:
> we can tell if simple whitespace tokenization for some la

Yeah, we can. We could also do some subword analysis to see how aggressively the words are split for lower-resourced languages. Ideally, we would not want too aggressive splits.

PR #621, opened 2021-05-26
add doc for creating dense index 

MXueguang:


lintool (commented):
@MXueguang Can we add some quick tests for this? E.g., let's have a small toy collection with 10 docs and test? Maybe also two shards, and then merge?

Not sure if this should go into normal unit tests or integration tests. If quick, maybe unit tests?

docs/create-dense-index.md
```diff
@@ -0,0 +1,58 @@
+## How do I index and search my own documents (Dense)?
```
lintool, line 1: to be consistent with our other docs, rename file to `usage-dense-indexes`?

scripts/merge_indexes.py
```diff
@@ -0,0 +1,30 @@
+import argparse
```
lintool, line 1: add boilerplate.

why isn't this just part of `pyserini/dindex`?

lintool (approved):

PR #635, opened 2021-06-02
fix typo in dindex guidance

MXueguang:


lintool (approved):
Can probably move args to one line if it's not too long...

Issue #646, opened 2021-06-07
better hybrid/evaluation script for dpr retrieval


MXueguang:
we currently compute "answer exists" at the very end of the pipeline, which makes hybrid tuning cost a lot of space. we can precompute it and do not store the raw text in the retrieval file.

MXueguang:
closed by https://github.com/castorini/pyserini/pull/649

PR #649, opened 2021-06-08
update dpr eval

MXueguang:
re https://github.com/castorini/pyserini/issues/646

will ready for review after testing.


ronakice (approved):
LGTM!

Issue #648, opened 2021-06-08
Issue with Hybrid search


jayparks:
Hello, I got this error when running hybrid retrieval.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/site-packages/pyserini/hsearch/__main__.py", line 159, in <module>
    results = hsearcher.batch_search(
  File "/usr/local/lib/python3.8/site-packages/pyserini/hsearch/_hybrid.py", line 46, in batch_search
    hybrid_result = {
  File "/usr/local/lib/python3.8/site-packages/pyserini/hsearch/_hybrid.py", line 47, in <dictcomp>
    key: self._hybrid_results(dense_result[key], sparse_result[key], alpha, k)
  File "/usr/local/lib/python3.8/site-packages/pyserini/hsearch/_hybrid.py", line 58, in _hybrid_results
    min_sparse_score = min(sparse_hits.values())
ValueError: min() arg is an empty sequence
```

The problem occurs when sparse retrieval does not generate any hit for some queries. The same can happen for dense retrieval.
https://github.com/castorini/pyserini/blob/c7b37d6073cda62685f64d6d0b99dc46f0718346/pyserini/hsearch/_hybrid.py#L60

For the moment, I'm trying to find a way to get around it.
Is there any way to force sparse retrieval to generate hits for all queries? Thanks in advance!

lintool:
Which dense model, what OS, package versions, etc.?

First debug would be to check versions of underlying dependencies (faiss, transformers, etc.):
https://github.com/castorini/pyserini/blob/master/docs/installation.md

jayparks:
MacOS: 10.15.7
Pyserini version: 0.12.0

The issue is not about dense retrieval but occurs on the sparse side.

lintool:
Are you able to replicate the dense retrieval results? On which model?

MXueguang:
I noticed this issue with @crystina-z at the end of last week.
If all words in queries are either stop-word or not in vocabulary dict, the search with this query will just return an empty list.
I think in the case the sparse result is empty, we can just use dense results for that query when hybrid scores.
I'll make the update today. @jayparks

jayparks:
[...]

PR #650, opened 2021-06-08
fix empty queries during hybrid

MXueguang:
re https://github.com/castorini/pyserini/issues/648

@jayparks I think this update can solve the problem?


lintool:
@crystina-z since you seem to be aware of the problem, can you take a look?

@jayparks wanna take a look?

jayparks (approved):
LGTM. thanks for the quick fix!

PR #657, opened 2021-06-09
DPR reproduction using dpr-nq-test index

ArthurChen189:
OS: Ubuntu 21.04
Top20	accuracy: 0.7800554016620499
Top100	accuracy: 0.8578947368421053


lintool:
I can't figure out which condition you're trying to reproduce... is it "Natural Questions (NQ) with DPR-Multi"? Your reported figures don't seem to match any of the numbers in the page...

ArthurChen189:
My apology for not clarifying. I tried wikipedia-dpr-multi-bf first but RAM OOM. @MXueguang recommended me to try the compressed version, which is [LINK](https://www.dropbox.com/s/t5mu2ii87rcnrmk/dindex-dpr-pq-2.tar.gz?dl=0).

lintool:
What's the hardware config you're trying to run on? Is this CC?

ArthurChen189:
This is not on CC. I am using my local machine which only has 16GB of RAM. Should I try this on CC?

lintool:
I think so... @MXueguang have we been able to get this to run on CC? IIRC, we had storage issues?

ArthurChen189:
I will try run this on CC again.

PR #663, opened 2021-06-15
Fetching doc text from dense indexes

### LINKED ISSUE #603: Fetching doc text itself from dense indexes
Currently, dense indexes don't have the `doc` feature to fetch the actual document text, since they don't store documents. The sparse indexes store the documents.

So, we could simple add a `doc` feature to the dense index API that just "routes" over to the corresponding sparse index. However, each dense index needs to know which corresponding sparse index has the documents... so we need to add a field here: https://github.com/castorini/pyserini/blob/master/pyserini/prebuilt_index_info.py

Something like: https://github.com/castorini/pyserini/blob/master/pyserini/prebuilt_index_info.py#L451

We add an additional field, e.g., "texts" and point to "msmarco-passage".

I think this would be a neat feature for the demo, so maybe @saileshnankani can take a look?

saileshnankani:
This PR adds a `doc(docid)` API to the `SimpleDenseSearcher`. Since dense searchers don't store documents, it routes over to the corresponding sparse index. 

To know which sparse index contains the corresponding document for a given dense index, a `texts` field is added in `prebuilt_index_info.py` for each dense index. This field is accessed inside the `doc()` api using the `prebuilt_index_name` attribute. 

Resolves #603


pyserini/dsearch/_dsearcher.py
```diff
@@ -366,6 +369,28 @@ def load_index(self, index_dir: str):
...
+
+        Returns
+        -------
+        Document
+            :class:`Document` corresponding to the ``docid``.
+        """
+        if self.prebuilt_index_name is None:
+            return None
+        sparse_index = get_sparse_index(self.prebuilt_index_name)
+        ssearcher = SimpleSearcher.from_prebuilt_index(sparse_index)
```
MXueguang, line 391: this ssearch initialization should be put into constructor?

saileshnankani: I wasn't sure if this would become an unnecessary overhead in the constructor, but i'll put it in

MXueguang (approved):

MXueguang:
LGTM!

PR #686, opened 2021-07-03
Re-organize encode module

MXueguang:
this will merge into
`pyserini.encode.sparse` when doing refactor


MXueguang:
we need a similar class define for DeepImpact. i.e. load DI model from huggingface model and do encoding to get same format

MXueguang:
@lintool 
Changes from current Pyserini:
- separate `pyserini.encode` from `index` and `search` as an independent module.
- `pyserini.encode` manages encoder models for {sparse, dense} * (query/ doc) encoder
- `pyserini.encode` can use CLI to do encoding from jsonl collection, write into jsonl or faiss index

two entry points.
- `pyserini/encode/sparse/__main__.py` encode to sparse weights. currently use uniCOIL but should support general sparse model
- `pyserini/encode/dense/__main__.py` encode to dense vectors. support general dense models

example:
we assume all fields are stored in contents and separated by `\n`
if each jsonl line has format:
`{"id": "[doc_id]#[seg_id]", "contents": "[url]\n[title]\n[seg_text]"}`
we will run following command to encode to dense embeddigns (stored in faiss)
```
python -m pyserini.encode.dense \
		--corpus <corpus dir> \
                --fields url title text \
                --encode-title \
		--encoder <encoder name or dir> \
		--embeddings <output dir> \
		--batch 16 \
		--device cuda:0 \
               --to-faiss
```

lintool:
Why do we need both `pyserini.encode.dense` and `pyserini.encode.sparse`? My concern is that down the road we'll develop  some model that's halfway between dense and sparse, and then we won't know where to put it.

I think the important parts are:
1. what model are we using?
2. properties of the input corpus.
3. properties of the output - if we are writing to `JsonVector` format, Faiss flat, etc.

So what about something like this?

```
python -m pyserini.encode input --input path/to/corpus
                                ... other properties of the corpus and input
                          output --output path/to/model
                                 --format JsonVector # or FaissIndexFlatIP ...
                          encoder --model dpr  # or unicoil, or deepimpact, etc.
                                ... other properties of the encoder
```

Yes, some encoders would be incompatible with certain output formats - e.g., I don't think we can take DeepImpact and output in  Faiss `IndexFlatIP`.

So maybe:

```
python -m pyserini.encode --input path/to/corpus
                          ... other properties of the corpus and input
                          --output path/to/model
                          --format JsonVector # or FaissIndexFlatIP ...
                          encoder --model dpr  # or unicoil, or deepimpact, etc.
                          ... other properties of the encoder
```

I can imagine there being encoder-specific options, so we have to somehow "delimit" with a subcommand...?

MXueguang:
```
python -m pyserini.encode input --input path/to/corpus
                                ... other properties of the corpus and input
                          output --output path/to/model
                                 --format JsonVector # or FaissIndexFlatIP ...
                          encoder --model dpr  # or unicoil, or deepimpact, etc.
                                ... other properties of the encoder
```
I think this makes sense.

should I merge `pyserini.encode.sparse._doc_encoders.py` and `pyserini.encode.dense._doc_encoders.py` into `pyserini.encode._doc_encoders.py` in this case? so we manage sparse/dense models together?

lintool:
> ```
> python -m pyserini.encode input --input path/to/corpus
>                                 ... other properties of the corpus and input
>                           output --output path/to/model
>                                  --format JsonVector # or FaissIndexFlatIP ...
>                           encoder --model dpr  # or unicoil, or deepimpact, etc.
>                                 ... other properties of the encoder
> ```

One detail here - some combinations of encoder and output either won't make sense, right? For example, if I specify DI as encoder and Faiss flat index, the results will be huge, right? I.e., |V| would be huge... Should we explicitly disallow this? Or just let it be impractical?

> I think this makes sense.
> 
> should I merge `pyserini.encode.sparse._doc_encoders.py` and `pyserini.encode.dense._doc_encoders.py` into `pyserini.encode._doc_encoders.py` in this case? so we manage sparse/dense models together?

Yes.

Building on this pattern, I would then do:

```
python -m pyserini.index.lucene # for building Lucene indexes - dispatch to current `IndexCollection`
python -m pyseirni.index.faiss # for building Faiss indexes ... e.g., we could specify something like --type hnsw
```

And then:

```
python -m pyserini.search.lucene
python -m pyserini.search.faiss
python -m pyserini.search.nmslib
...

python -m pyserini.search.fusion # for hybrid
```

So, `SimpleSearcher` would become something like `pyserini.search.lucene.LuceneSearcher`.

MXueguang:
thoughts about `pyserini.index`,
I think one of the shortage of creating `pyserini.index.lucene_index` `pyserini.index.faiss_index` `pyserini.index.nmslib_index` is that we are just writhing 3 scripts, the code would be duplicate and not usable beyond the CLI.
We should try to define objects to wrap these code.
I think in the retrieve stage, we will define `LuceneSearcher` `ImpactSearcher` `FaissSearcher` `NmslibSearcher` ? if so, I think we can just define a `Searcher.index_from_jsonl`  function in the corresponding searcher?
in this case, we can just run a single script for indexing. do something like
`python -m pyserini.index --type LuceneSearcher --collection <dir> --index <dir>`
and for search
`python -m pyserini.search --type LuceneSearcher --index <dir>, ...other parameters`

MXueguang:
```
python -m pyserini.encode input --corpus path/to/corpus
                               --fields 'url' 'title' 'text' 'expand'
                              --shard-id 0 --shard-num 4
                          output --embeddings path/to/output/dir
                                 --to-faiss
                          encoder --encoder unicoil
                                --fields 'title' 'text' 
                               --batch 32
```

pyserini/encode/_doc_encoders.py
```diff
@@ -0,0 +1,211 @@
...
+        if self.pooling == "mean":
+            embeddings = mean_pooling(outputs[0], inputs['attention_mask']).detach().cpu().numpy()
+        else:
+            embeddings = outputs[0][:, 0, :].detach().cpu().numpy()
+        if self.l2_norm:
+            sklearn.preprocessing.normalize(embeddings, axis=1)
+        return embeddings
+
+
+def mean_pooling(last_hidden_state, attention_mask):
```
lintool, line 140: move this either to the beginning or to the end?

pyserini/encode/_embedding_writer.py
```diff
@@ -0,0 +1,82 @@
...
+#
+
+import json
+import os
+
+import faiss
+import numpy as np
+
+
+class EmbeddingWriter:
```
lintool, line 24: Not sure I like the name of this class... We don't use the term "Embedding" anywhere else. We say that encoders generate representation vectors... but it's hard to make a good name based on this.

Options I can think of: `VectorSerializer`, `VectorWriter`, `RepresentationOutput`, `RepresentationWriter`?

Do you like any of these?

pyserini/encode/_models.py
```diff
@@ -0,0 +1,122 @@
+#
```
lintool, line 1: What's the difference between `_models` and `_doc_encoders`?

pyserini/encode/_qry_encoders.py
```diff
@@ -0,0 +1,224 @@
+#
```
lintool, line 1: Why aren't doc and query encoders grouped together?

Why not an organization where each family is put in a separate file?

E.g., `ance.py`, `dpr.py`, etc.?

lintool (commented):
Some points for discussion...

pyserini/encode/_qry_encoders.py
MXueguang: yeah it make sense

pyserini/encode/_models.py
MXueguang: _models is to define custom hgf models.

models are used by encoders.

but I'll group it into encoders as we discussed vvvv

pyserini/encode/_embedding_writer.py
MXueguang: RepresentationWriter ?

MXueguang:
We will try to mirror the sklean repo structure.
each group of models will be stored in an individual file. e.g. `_ance.py`, `_dpr.py`
utils obj will be placed into `_base.py` together with base classes.
The model class will be imported in `__init__.py`.
So eventually we call `from pyserini.encode import AnceDocumentEncoder`

pyserini/encode/_pseudo.py
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import json
+
+from pyserini.encode import QueryEncoder
+
+
+class PseudoQueryEncoder(QueryEncoder):
```
lintool, line 22: How about something like `CachedDataQueryEncoder`?

pyserini/encode/_tct_colbert.py
```diff
@@ -0,0 +1,70 @@
...
+            return_tensors='pt'
+        )
+        inputs.to(self.device)
+        outputs = self.model(**inputs)
+        embeddings = self._mean_pooling(outputs["last_hidden_state"][:, 4:, :], inputs['attention_mask'][:, 4:])
+        return embeddings.detach().cpu().numpy()
+
+
+class TctColBertQueryEncoder(QueryEncoder):
+
```
lintool, line 51: Be consistent throughout whether you want this empty line or not...

pyserini/encode/_pseudo.py
lintool: Actually, why not just `QueryEncoder.load_data_from_jsonl(...)` or something like that?

lintool (approved):
lgtm

We need test cases... I'm fine if it's in another PR, but please don't forget about it.

Maybe end-to-end integrations with CACM collection or something? I.e., encode the entire corpus, and then do retrieval. Shouldn't be too long because CACM collection isn't very big...

PR #697, opened 2021-07-13
onnx & fp16 for tct_colbert encoding

MXueguang:
```
original               2hours    /   1M texts
fast_tok               1.5hours /    1M texts
fast_tok_fp16          1.2hours /    1M texts
fast_tok_fp16_onnx      50 min   /   1M texts
```

To encode with TCT_ColBERT
```
python -m pyserini.encode input --corpus <corpus_jsonl_dir> \
                                --fields 'url' 'text' 'expand'  \
                          output --embeddings <output_faiss_dir>  \
                                 --to-faiss    \
                          encoder --encoder tct_colbert-msmarco.onnx  \
                                  --fields 'text'  \
                                  --batch 16

```


lintool (approved):
Do we want a better name than ._pseudo?

But otherwise, lgtm

MXueguang:
> Do we want a better name than ._pseudo?
> 

changed to CachedData
I forgot to remove, so removed in this PR

PR #701, opened 2021-07-14
reorganize index module

MXueguang:


pyserini/index/lucene.py
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+#
+
+from jnius import autoclass
+import sys
+import os
+
+args = sys.argv[1:]
+
+# argument check
```
lintool, line 23: I have a suggestion here: how about we go through the args and change double dashes back into single dashes. That way we can transparently fix the differences between Java and Python args.

lintool (commented):
I see that you haven't removed the old main... so until we do, there will be two parallel methods for building Lucene indexes?

MXueguang:
>I see that you haven't removed the old main... so until we do, there will be two parallel methods for building Lucene indexes?

yea, we can drop the old version until the new version is stable

lintool (approved):

PR #727, opened 2021-08-06
Reproduce doc, unicoil for msmarco-v2 0shot

MXueguang:


lintool (approved):

MXueguang:
@lintool I just pushed reproduce doc in this PR

lintool (approved):
lg, but please confirm that the new code doesn't screw up the old results...

(do we have a test case)?

PR #740, opened 2021-08-19
Update experiments-msmarco-doc.md

apokali:
Running environment: UW undergrad student Linux environment
Problems: when index gzipped files, possible errors would occur: CPU time limit exceed.
Possible solutions: run the command: "ulimit -t unlimited" before indexing files.


MXueguang:
Hi @apokali,  are the results you get the same as the numbers in the reproduce document?

apokali:
> Hi @apokali, are the results you get the same as the numbers in the reproduce document?

Hi Yes. I have successfully reproduced the same results. It did take a long time for me to fix the issues that I mentioned above, but the results are the same.

lintool:
There's something wrong with your commit id... it says "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." when you visit the URL.

Please use the commit id of _this_ repo at the point where you did the reproduction?

Thanks!

lintool (approved):

PR #751, opened 2021-09-04
Fix bpr encoded query

MXueguang:


lintool (approved):
Any idea why it was working for you previously?

MXueguang:
I was using otf encoding accidentally ...

PR #752, opened 2021-09-08
add uniCOIL TILDE expansion instructions for MS MARCO passage v1

ArvinZhuang:
Hi, 
I have added the instructions to reproduce results for uniCOIL with TILDE expansion on MS MARCO passage v1.

The anserini comparable guide has not been added yet. Will do it as soon as possible, same for MS MARCO v2 guide.

And feel free to change corpus or index downloading links if they do not make sense to you.


lintool (approved):
Awesome, thanks!

PR #765, opened 2021-09-18
Update uniCOIL + d2q passage

lintool:
+ Add ability to read compressed topics.
+ Add documentation for on-the-fly query encoding.

For on-the-fly query encoding, I get on my iMac Pro:

```
#####################
MRR @10: 0.35088535270841803
QueriesRanked: 6980
#####################
```

And on `orca`:

```
#####################
MRR @10: 0.3508734138354477
QueriesRanked: 6980
#####################
```

Both scores are a tiny bit lower, but still reasonably close pre-computed queries, which we get:

```
#####################
MRR @10: 0.35155222404147896
QueriesRanked: 6980
#####################
```


MXueguang:
yes, the slightly difference to pre-computed can be caused by quantization.
but reasonably close

MXueguang (approved):

PR #776, opened 2021-09-23
late import AnceEncoder in dindex.

t-k-:
This cuts the dindex dependence on javac.

Dependence chain:

pyserini/pyserini/dindex/_base.py
from pyserini.dsearch import AnceEncoder

pyserini/pyserini/dsearch/_dsearcher.py
from pyserini.search import SimpleSearcher, Document

/home/tk/Desktop/pyserini/pyserini/search/__init__.py
from ._base import Document, JDocument ...

### DIFF
--- a/pyserini/dindex/_base.py
+++ b/pyserini/dindex/_base.py
@@ -19,8 +19,6 @@ import torch
 from transformers import AutoModel, AutoTokenizer, BertModel, BertTokenizer, DPRContextEncoder, \
     DPRContextEncoderTokenizer, RobertaTokenizer
 
-from pyserini.dsearch import AnceEncoder
-
 
 class DocumentEncoder:
     def encode(self, texts, titles=None):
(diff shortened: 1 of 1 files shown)


lintool:
Hi @t-k- what's the rationale for this? Pyserini depends Java for many features, so it's unclear why you'd want this?

t-k-:
@lintool so dindex is run as a command, and setting up the java env (using normal encoders) is just not necessary. I am ok to close this one if it doesn't bother other ones.

lintool:
I think what I'm trying to say is that this PR would make sense to support a "partially installed" Pyserini... but at the cost of a PEP 8 violation: https://www.python.org/dev/peps/pep-0008/#imports

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

IMO, not a worthwhile tradeoff...?

t-k-:
yea, it makes sense and I will close the issue. You want me to mention PEP8 on readme for future potential code contributor? @lintool

Issue #792, opened 2021-10-01
KeyError when reproducing deepimpact on MSMARCO


Qiaoyf96:
Hi,

I am trying to reproduce the deepimpact result on MSMARCO following https://github.com/castorini/pyserini/blob/master/docs/experiments-deepimpact.md, but received the following error:

```
$ python -m pyserini.search --topics ../msmarco-deepimpact-pyserini/collections/topics.msmarco-passage.dev-subset.deep-impact.tsv --index ../msmarco-deepimpact-pyserini/indexes/lucene-index.msmarco-passage-deepimpact-b8 --output ../msmarco-deepimpact-pyserini/runs/run.msmarco-passage-deepimpact-b8.tsv --impact --hits 1000 --batch 36 --threads 12 --output-format msmarco
Running ../msmarco-deepimpact-pyserini/collections/topics.msmarco-passage.dev-subset.deep-impact.tsv topics, saving to ../msmarco-deepimpact-pyserini/runs/run.msmarco-passage-deepimpact-b8.tsv...
  1%|▉                                                                                                                                                                                    | 35/6980 [00:00<00:01, 3849.20it/s]
Traceback (most recent call last):
  File "/home/yifan/anaconda3/envs/pyserini/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/yifan/anaconda3/envs/pyserini/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/yifan/Downloads/pyserini/pyserini/search/__main__.py", line 242, in <module>
    batch_topics, batch_topic_ids, args.hits, args.threads, fields=fields
  File "/home/yifan/Downloads/pyserini/pyserini/search/_impact_searcher.py", line 156, in batch_search
    if self.idf[token] > self.min_idf:
KeyError: 'cyclopeanbinocularmonoculartrichromatic'
```

Could you please take another look? Thank you in advance for your help.

MXueguang:
@Qiaoyf96  It seems like a bug caused by a recent change will fix it soon.

Qiaoyf96:
Got it. Thanks!

MXueguang:
should be fix in this PR https://github.com/castorini/pyserini/pull/793

lintool:
[...]

PR #793, opened 2021-10-01
fix bug in impact searcher 

MXueguang:
queries tokens probably not in idf list


lintool (approved):

MXueguang:
to fix https://github.com/castorini/pyserini/issues/792

Issue #770, opened 2021-09-21
Incorrect encoding on Windows


stekiri:
When using pyserini under Windows, it seems that the encoding of strings is breaking when passed to the JNI via the [pyjnius](https://github.com/kivy/pyjnius) package.

It happens when a string is encoded as UTF-8 like this `JString(my_str.encode('utf-8'))` (e.g., https://github.com/castorini/pyserini/blob/master/pyserini/search/_searcher.py#L114). It only occurs under Windows as it must collide with the default Windows encoding `CP-1252`.

I [discussed this issue](https://discord.com/channels/423249981340778496/670954391641128960/889782470403293185) with the maintainers of `pyjnius` and it seems that to make it work independently from the platform, the `.encode('utf-8')` could simply be dropped.

Was there a reason why this manual encoding was used in `pyserini`?

I created a branch with the [changes](https://github.com/castorini/pyserini/compare/master...stekiri:encoding-fix?expand=1), I could do a PR if you wish.

stekiri:
In fact, it can also be fixed by setting `set _JAVA_OPTIONS=-Dfile.encoding=UTF-8` as mentioned [here](https://github.com/castorini/pyserini/blob/master/docs/installation.md#troubleshooting-tips). The Java option is indeed a bit confusing as it is not only about files, but also about other internals apparently 😞

I am wondering what the best solution for this one is 🤔

If dropping the `.encode('utf-8')` works well under all platforms, it has the advantage that no platform-specific setup is required and, therefore, it will be easier to use (no one will forget to set this param). If the encoding is indeed required, I strongly suggest to make this requirement for Windows user a lot more visible. Maybe in an extra section of the [package installation](https://github.com/castorini/pyserini#package-installation)?

lintool:
Hi @stekiri - thanks for raising this issue!

My memory is a bit fuzzy here, but `.encode('utf-8')` was in response to an earlier underlying issue with pyjnius. If pyjnius has address this issue in a platform independent manner, then, yes, we should probably remove in pyserini.

Would you mind sending a PR to that effect? It'd be great if you also included a test case that would catch this issue in the future.

I'll run our suite of regression tests and see if anything is broken... there are queries with non-ASIIC encoding in, for example, MS MARCO - we need to confirm that this doesn't break anything...

Thanks in advance!

stekiri:
I looked a bit into the PR/commit history to better understand this 🕵️ 

Please, correct me if I got it wrong 😉 

[...]

PR #795, opened 2021-10-04
Remove unnecessary utf8 encoding causing problems under Windows

stekiri:
## Issue Description

When using pyjnius in a version prior 1.3.0, creating a JString with non-ASCII characters resulted in a failure. To work around this issue, the strings where manually encoded to utf8 (`.encode('utf-8')`), however, this is not necessary anymore with newer pyjnius versions.

The manual `.encode('utf-8')` caused problems under Windows when the environment variable `_JAVA_OPTIONS=-Dfile.encoding=UTF-8` was not set.

See the full discussion here: https://github.com/castorini/pyserini/issues/770

## Fix

Remove `.encode('utf-8')` everywhere in the code and bump pyjnius version to `>=1.3.0`. With this fix, it is not required to use an extra environment variable under Windows anymore.

## Bonus

Added a test that ensures that a newer, fixed version of pyjnius is being used.

### DIFF
--- a/pyserini/analysis/_base.py
+++ b/pyserini/analysis/_base.py
@@ -151,7 +151,7 @@ class Analyzer:
         List[str]
             List of tokens corresponding to the output of the analyzer.
         """
-        results = JAnalyzerUtils.analyze(self.analyzer, JString(text.encode('utf-8')))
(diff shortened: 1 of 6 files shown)


lintool (approved):

lintool:
Thanks so much for your contribution!

Issue #704, opened 2021-07-14
End-to-end integration tests for encode


lintool:
The CACM collection is small enough that this might be doable on CPU?

lintool:
@MXueguang we can close this issue right?

MXueguang:
closed by https://github.com/castorini/pyserini/pull/796

PR #796, opened 2021-10-04
add unittest and integration test for encoder

MXueguang:
- unittest trying to make sure encoder output are correct.
- integration test trying to make sure the full pipeline work


lintool:
`integrations/resources/sample_collection_cacm/cacm.json` is fairly big?

I usually wget otf from here: https://github.com/castorini/anserini-data/tree/master/CACM

Let's not bloat the repo if there are easy alternatives?

lintool (changes requested):
See comment.

lintool:
Ref #704

MXueguang:
2.4M, not that big. but yes, we can use `wget` to download

lintool:
> 2.4M, not that big. but yes, we can use `wget` to download

It's all relative, I suppose ;)

2.4M doesn't fit on one of these, for example: https://en.wikipedia.org/wiki/Floppy_disk

MXueguang:
tested on my ubuntu machine.

lintool (approved):

Issue #659, opened 2021-06-11
Reorganization of packages for dense retrieval?


lintool:
Hey @MXueguang - since `dindex` is really about encoding the corpus, and we're using `IndexFlatIP` mainly as a data interchange format (as opposed to TF records), does it make sense to reorg the package structure for the next major release?

E.g., something like:

+ `pyserini.encode` - primarily about encoding (bf index is really just a "side effect")
+ `pyserini.findex` - for faiss indexes (e.g., faiss hnsw)
+ `pyserini.nindex` - for nmslib indexes

Or shall we be explicit? e.g., `pyserini.lucene_index`, `pyserini.faiss_index`, `pyserini.nmslib_index`.

And `pyserini.faiss_search` and `pyserini.nmslib_search`.

So, for example, methods in `pyserini.nmslib_index` can either take `IndexFlatIP` as input or a JSON collection with explicit vectors?

Thoughts?

lintool:
Noting apparent code duplication I just discovered:

https://github.com/castorini/pyserini/blob/master/pyserini/dindex/_base.py#L54
https://github.com/castorini/pyserini/blob/master/pyserini/encode/_dpr.py#L22

MXueguang:
@lintool 
The first stage of code reorganization targeting at following current modules:
- `pyserini.dindex`: encoding raw collections in `jsonl` format, store as Faiss flat index. This module currently contains all document encoder class for dense retrieval.
- `pyserini.index`: build lucene index
- `pyserini.dsearch`: do dense search on faiss index. This module currently contains all query encoder class for dense retrieval.

In refactoring, we are trying to do:
- remove `pyserini.dindex`, replace it with `pyserini.encode`: `pyserini.encode` will contains all document encoder and query encoder class for both dense (e.g. dpr) and sparse (e.g. unicoil) model.
- remove the query encoders defined in current `pyserini.dsearch` by class defined in `pyserini.encode`
- `pyserini.encode` will do: a) encode raw collection (in `jsonl` format) into `jsonl` format with vectors (for sparse/dense), b) encode raw collection and store as Faiss Flat index (for dense only).
- will change current `pyserini.index` into `pyserini.index.lucene` `pyserini.index.faiss` and `pyserini.index.nmslib`
- `pyserini.index.lucene` can a). index raw text collection in `jsonl` format into lucene bm25 index. b). index collection with sparse vector (generated by `pyserini.encode` using sparse model, e.g. unicoil) into lucene impact index.
[...]

PR #797, opened 2021-10-04
Add pca/pq experiments for dpr

MXueguang:


lintool (approved):
I think this is fine for now, but we're going to continue bumping into #659 - in this PR we're starting to introduce Faiss-specific options into `dindex`, which is supposed to be "generic". What happens when we want to compare nmslib with Faiss?

MXueguang:
right now we have  `pyserini.index.faiss` and `pyserini.index.nmslib`
if we want to compare with nmslib, we can use `pyserini.index.nmslib` to convert faiss index or raw vector into hnsw.

PR #802, opened 2021-10-05
remove duplicate code for dindex

MXueguang:
remove duplicate code in original `pyserini.dindex`.
`pyserini.dindex` module will be replaced by `pyserini.encode`

see details in https://github.com/castorini/pyserini/issues/659
>remove pyserini.dindex, replace it with pyserini.encode: pyserini.encode will contains all document encoder and query encoder class for both dense (e.g. dpr) and sparse (e.g. unicoil) model.
remove the query encoders defined in current pyserini.dsearch by class defined in pyserini.encode
pyserini.encode will do: a) encode raw collection (in jsonl format) into jsonl format with vectors (for sparse/dense), b) encode raw collection and store as Faiss Flat index (for dense only).


integrations/dense/test_create_index.py
```diff
@@ -1,90 +0,0 @@
-#
```
lintool, line 1: Why are we removing this test case?

lintool (commented):
?

integrations/dense/test_create_index.py
MXueguang: duplicate with `test_encode.py`

lintool: ah, okay!

lintool (approved):

PR #819, opened 2021-10-12
update name for msmarco v2

MXueguang:


MXueguang:
changed all msmarco-passage/doc-v2 to msmarco-v2-passage/doc in parallel with anserini

lintool:
nah... go ahead and merge. I'll send another PR once paths have been fixed.

PR #828, opened 2021-10-16
Fix one path of index module depending on search.

w32zhong:
When writing ColBert encoder, I found any complex or compositional
`encoder` has a good chance to depend on `index` module (because such an index
needs more complex RepresentationWriter), and if `index` module depends
on search module, it has a good chance to depend on the `encode` module
again, thus a loop of dependency is formed.

We can fix it by cutting the very counter-intuitive "index --> search"
dependency link because it violates our common abstraction.


t-k-:
An upcoming PR to support Colbert needs this change: https://github.com/t-k-/pyserini/tree/patch-colbert

Issue #566, opened 2021-05-13
faiss-gpu not supported


isoboroff:
The README demos for dense and hybrid search work with faiss-cpu, but if I only have faiss-gpu installed, I get:
```
Attempting to initialize pre-built index msmarco-passage-tct_colbert-hnsw.
/local_home/soboroff/.cache/pyserini/indexes/dindex-msmarco-passage-tct_colbert-hnsw-20210112-be7119.7e12ae728ea5f2ae6d1cfb88a8775ba8 already exists, skipping download.
Initializing msmarco-passage-tct_colbert-hnsw...
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-4-93ed650a7449> in <module>
      5 ssearcher = SimpleSearcher.from_prebuilt_index('msmarco-passage')
      6 encoder = TctColBertQueryEncoder('castorini/tct_colbert-msmarco')
----> 7 dsearcher = SimpleDenseSearcher.from_prebuilt_index(
      8     'msmarco-passage-tct_colbert-hnsw',
      9     encoder

~/pyserini-play/venv/lib/python3.8/site-packages/pyserini/dsearch/_dsearcher.py in from_prebuilt_index(cls, prebuilt_index_name, query_encoder)
    260 
    261         print(f'Initializing {prebuilt_index_name}...')
--> 262         return cls(index_dir, query_encoder)
    263 
    264     @staticmethod

~/pyserini-play/venv/lib/python3.8/site-packages/pyserini/dsearch/_dsearcher.py in __init__(self, index_dir, query_encoder)
    229 
    230     def __init__(self, index_dir: str, query_encoder: QueryEncoder):
--> 231         requires_faiss(self)
    232         self.query_encoder = query_encoder
    233         self.index, self.docids = self.load_index(index_dir)

~/pyserini-play/venv/lib/python3.8/site-packages/transformers/file_utils.py in requires_faiss(obj)
    501     name = obj.__name__ if hasattr(obj, "__name__") else obj.__class__.__name__
    502     if not is_faiss_available():
--> 503         raise ImportError(FAISS_IMPORT_ERROR.format(name))
    504 
    505 

ImportError: 
SimpleDenseSearcher requires the faiss library but it was not found in your environment. Checkout the instructions on the
installation page of its repo: https://github.com/facebookresearch/faiss/blob/master/INSTALL.md and follow the ones
that match your environment.
```

isoboroff:
[...]

PR #829, opened 2021-10-17
ColBert implementation.

w32zhong:
Only containing code as for now, and roughly tested on a small **math dataset**. I've tried a few topics and looks like it is working by watching top-10 results.

Need help to get a trained text colbert to update the docs, as training colbert and encoding for any corpus more than 8 million docs are too much for me. I couldn't find any official one [here](https://huggingface.co/models?search=colbert) (please let me know if you have, encoded queries and corpus would be perfect). Otherwise, it is hard to verify and update the integration scripts. However, I can be responsible to test it and update docs once the encoded corpus/topics become available to me.

The purpose of creating this PR:
* once tested on test corpus, it will help me verify my colbert implementation and reproduce some results related to my research.
* it adds colbert to pyserini, closely following [the original implementation](https://github.com/stanford-futuredata/ColBERT), but with refactored code and comments, less-optimized code. The original one is less easy to follow.
* In addition, it contains reference implementation for [GPU faiss training](https://github.com/t-k-/pyserini/blob/aced3c21f7b710414b8847cf5275fb978cc07522/pyserini/index/faiss_gpu.py) which can be exported and used by other parts of the project (closely related to https://github.com/castorini/pyserini/issues/566).


t-k-:
The evaluation time using the current colbert code is estimated by tqdm:
5/6980 [02:17<48:59:30, 25.29s/it]
so about 20~30s per query. And it takes 2 days to generate dev results (actually only half of the results, each node on cc has 128GiB memory constraints, therefore I have run 2 jobs to get full results in parallel)

so far, from what I have got for the first 16 queries (searching only half of the index due to the main memory limit), the MRR is not bad, to say the least.

#####################
MRR @10: 0.3819444444444445
QueriesRanked: 13
#####################

t-k-:
Quick update.

current implementation using distilbert given by @justram produced an overall score that looks low:
#####################
MRR @10: 0.31478604913467884
QueriesRanked: 6980
#####################

However, when looking into different index shards, result from shards 0 to 10 looks extremely low: 
#####################
MRR @10: 0.023306369179796265
QueriesRanked: 6980
#####################

while result from shard 10 to 20 looks good to me:
#####################
MRR @10: 0.3511282060319803
QueriesRanked: 6980
#####################

still need some time to sort this out.

t-k-:
The reason that mrr results from the first half of the index is low, is because only ~10 percent of qrels are judging documents located in the first half of the index. In other words, "msmacro passage subset" judgments heavily target larger docIDs.

this can be verified using:
```py
mid_docid = 4420909
cnt = [0] * 2
with open('qrels.msmarco-passage.dev-subset.txt', 'r') as fh:
    for line in fh:
        line = line.rstrip()
        fields = line.split()
        docid = int(fields[2])
        cnt[docid <= mid_docid] += 1
print(cnt) # [6630, 807]
```
so `MRR@10 = 0.315` is a reasonable overall result for this version.

Possible reasons it is lower than reported:
* this distilbert is trained with punctuation, while inference is done with punctuation removal.
* query augmentation is not applied, i.e., it didn't append [MASK]s at the end of query tokens.
* encoder is not well trained
* bug in the searcher? (feel unlikely after a few times scanning of the code)

I will try to boil down and update more on this later.

t-k-:
this version mmr:
#####################
MRR @10: 0.3417674648656017
QueriesRanked: 6980
#####################

t-k-:
@justram and I have found a discrepancy in our code.

His generated vectors are fully padded (to the max length) while mine is not, however, it is not feasible to do e2e using fully padded because that is too costly.

So for reproducing the almost exact numbers from the paper (~0.36), we still need time to modify the code **on the training side**, to produce short vectors.

However, we have a version that was trained w/o punctuation masking but was doing inference w/ punctuation masking. This happens to be able to work in both of our codes. As the training vs inference inconsistency suggests, the score is lower, however, this e2e version can achieve better results than reranked version:

reranking MRR @10: 0.3352891481329868
e2e MRR @10: 0.34252251330331507

although we still need to verify once the less-padding encoder is available, this score probably indicates we are doing right.

t-k-:
@justram and I have done many attempts. We have trained a Colbert encoder that does not generate maximal padding, this time it should be completely consistent on both training and e2e retrieval, it turns out both the e2e score (using this code) and reranking score (using @justram pipelines) are very similar:

For reranking: MRR @10: 0.3409763155500974
For end2end: MRR @10: 0.34189759858097946

First, I don't think it is easy to apply original code also, they [load entire word vectors](https://github.com/stanford-futuredata/ColBERT/blob/master/colbert/ranking/index_part.py#L34) into memory and it is not easy to find a way around without major change on the original code. It is only feasible to evaluate on GPU cluster with memory over 400 GiB for msmarco-v1. In contrast, our code can split indexes and fit into any GPU in theory.

Second, I think it could be the best we can get now. Some arguments:
* the original colbert paper also report very similar scores for re-rank vs. e2e (36.4 and 36.7) on their homebrew eval set.
* as shown in the above comments, the DEV set is highly biased towards large docID, our e2e effectiveness can be shown very good if results are returned from high address and judgments are much less sparse, e.g., for the percentile 8 to 9, we can get MRR @10: 0.448! I've also sampled top results from other parts of the index, they are actually pretty good results except they are not judged. 
* Some visualization is also done to verify the e2e code is working (did many checks like this especially edge docIDs near the separation of index shards where I feel code logic may go wrong):
  image
  image
* the reranking obtained by the current encoder (that has less padding and is compatible with e2e) from @justram pipeline is the best score we can obtain so far (it is trained on 2 A6000 with 192 batch size), and there is no reproduced e2e score for high-score (~0.36) colbert as I am aware of, [leaderboard on msmarco](https://microsoft.github.io/MSMARCO-Passage-Ranking-Submissions/leaderboard/) has similar performance of ~0.34. However, if we could get a higher score on reranking, this e2e score can be better I think. 

@justram is training a colbert with query augmentation, that is the last thing we have not tried. will update on that one later.

t-k-:
the reported result has been reproduced (using a private checkpoint).

#####################
MRR @10: 0.3611133851821534
QueriesRanked: 6980
#####################

Thanks for all the very needed help from @justram @okhat and @lintool !
I am going to update docs for reproducing this result.

MXueguang (approved):
lgtm

lintool:
@t-k- Just to confirm - the current decision is _not_ to merge this to master for now. cc @MXueguang

t-k-:
Yes, sorry I did not notify @MXueguang earlier. Should I temporarily close this issue?

lintool:
I've converted it into a "draft" so it doesn't get merged accidentally. Thanks!

PR #844, opened 2021-10-29
LTR document exp + regression test

stephaniewhoo:
+ refactoring LTR to support document exp
+ add regression test for document
+ add documents for LTR document repro


yuki617 (approved):
lgtm.

PR #845, opened 2021-10-29
Add test for emoji document vector

yuki617:


tests/test_index_reader.py
```diff
@@ -47,6 +47,20 @@ def setUp(self):
...
+        self.emoji_corpus_path = 'integrations/resources/sample_collection_json_emoji'
+
+    def test_doc_vector_emoji_test(self):
+        index_dir = 'temp_index'
+        self.temp_folders.append(index_dir)
+        cmd1 = f'python -m pyserini.index -collection JsonCollection -generator DefaultLuceneDocumentGenerator -threads 1 -input {self.emoji_corpus_path} -index {index_dir} -storeDocvectors'
+        _ = os.system(cmd1)
+        tempt_index_reader = index.IndexReader(index_dir)
+        doc_vector = tempt_index_reader.get_document_vector('CACM-3807')
+        self.assertAlmostEqual(doc_vector['🙂'], 1)
```
lintool, line 60: why almost equal? this should be exact match, no?

integrations/resources/sample_collection_json_emoji/simple_cacm_corpus.json
```diff
@@ -0,0 +1 @@
+{"id": "CACM-3807", "contents": "Many emoji were pulled from the conventional depictions of emotions used in Japanese manga (Moschini, 2016) 😋, as evident in the nose bubble for sleep 🙂, shade on the forehead for dread , and the large sweat drop on the forehead for anxiety 😀 (Cohn, 2013a; Cohn & Ehly, 2016).\n"}
```
lintool, line 1: I think this should go in `tests/resources/`, not integrations.

lintool (changes requested):

tests/test_index_reader.py
yuki617: Sorry, it should be equal just fixed that

lintool (approved):

PR #860, opened 2021-11-03
Undo removal of encode.utf-8 

yuki617:
Fix the broken clrpf regression test;
Fix the unittest tests.test_index_reader.test_doc_vector_emoji_test failure


lintool:
hey @stekiri - it's looking like we have to basically revert #795 ... but we can add in documentation to clarify the Windows issues.

lintool:
I'm on pyjnius 1.4.0.

I can confirm that tests pass with this patch.

lintool:
#862 seems to be a better fix - that's been merged, pending additional testing.

Issue #841, opened 2021-10-27
Broken clprf integration tests


lintool:
At commit 56527181dc70226fd57194dbe3f322e56594269c:

```
$ python -m unittest integrations.clprf.test_clprf.TestSearchIntegration.test_core18_lr
Running core18 topics, saving to ./integrations/tmp4232/core18_lr.txt...
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 110/110 [00:02<00:00, 46.09it/s]
 53%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████                                                                                                                         | 533/1000 [00:11<00:10, 45.10it/s]
  0%|                                                                                                                                                                                                                                                                               | 0/50 [00:14<?, ?it/s]
Traceback (most recent call last):
  File "/opt/local/stow/Python3-3.6.4/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/local/stow/Python3-3.6.4/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/scratch2/jimmylin/pyserini/pyserini/search/__main__.py", line 260, in <module>
    scores, docids = ranker.rerank(docids, scores)
  File "/scratch2/jimmylin/pyserini/pyserini/search/reranker.py", line 85, in rerank
    return self._rerank_with_classifier(doc_ids, search_scores)
  File "/scratch2/jimmylin/pyserini/pyserini/search/reranker.py", line 68, in _rerank_with_classifier
    train_vecs, train_labels, test_vecs = self._get_prf_vectors(doc_ids)
  File "/scratch2/jimmylin/pyserini/pyserini/search/reranker.py", line 63, in _get_prf_vectors
    test_vecs = self.vectorizer.get_vectors(doc_ids)
  File "/scratch2/jimmylin/pyserini/pyserini/vectorizer/_base.py", line 127, in get_vectors
    tfidf = tf[term] * self.idf_[term]
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
F
======================================================================
FAIL: test_core18_lr (integrations.clprf.test_clprf.TestSearchIntegration)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #862, opened 2021-11-04
Drop JString wrapping

stekiri:
As discussed in https://github.com/castorini/pyserini/issues/770, converting Python strings to Java strings using `JString(...)` does not work correctly if the string contains an emoji. However, letting the `pyjnius` library perform the conversion automatically without the explicit `JString` conversion seems to work correctly.

Unit tests pass successfully. @lintool, you can check the regression tests on your side to make sure it doesn't break anything.

### DIFF
--- a/docs/usage-pyjnius.md
+++ b/docs/usage-pyjnius.md
@@ -13,10 +13,9 @@ Now `autoclass` can be used to provide direct access to Java classes:
 ```python
 from jnius import autoclass
 
-JString = autoclass('java.lang.String')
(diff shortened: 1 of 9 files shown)


lintool:
Seems to be a duplicate of #861 - haha, both you and @yuki617 sent a PR!

Okay, I will take a look and run more integration tests.

lintool:
Seems to fix #841 - I will merge and run more rigorous tests.

lintool:
Merged! Thanks @stekiri and @yuki617 both for your help so far!

I'll run our more extensive integration suite (takes a lot longer) and then close all issues if everything looks good.

PR #865, opened 2021-11-05
update requirements

MXueguang:


lintool (approved):
Are we changing `pyjnius>=1.3.0` to 1.4 also?

PR #887, opened 2021-11-19
fix bug in faiss index

MXueguang:


lintool:
Hey @MXueguang can you describe the bug briefly so I can actually do a proper CR?

MXueguang:
this part is not been commonly used before.
create faiss index from jsonl files. 
the bug was the `docid` files was overwritten in each loop.
The file should be opened outside of the loop. 
```
        with open(os.path.join(args.output, 'docid'), 'w') as f_out:

```

lintool (approved):
Please summarize issue in commit log.

PR #893, opened 2021-11-29
IBM model1 reranker

yuki617:


stephaniewhoo (approved):
I am able to repro with ltr ibm model 
`cat result-ibm-0.1.json
{"map": "0.2207", "ndcg": "0.2889"}`
This is a bit lower than Yuqi's score since we use different ibm models. But it looks good to me overall

scripts/rank_ibm.py
```diff
@@ -0,0 +1,311 @@
...
+    pred.reverse()
+    docs.reverse()
+    return pred, docs
+
+
+def get_ibm_score(arguments):
+    query_text_lst = arguments['query_text_lst']
+    test_doc = arguments['test_doc']
+    searcher = arguments['searcher']
+    fieldName = arguments['fieldName']
```
lintool, line 108: Please use Python variable naming conventions - `field_name` etc.

yuki617:
@stephaniewhoo  Yes, sorry I forgot to send you the new generated ibm model. The new model is using bert-tokenized raw word, the overall score will be higher. 0.2207 is the right one with old ibm model. Thanks!

scripts/rank_ibm.py
yuki617: Thanks, names are all updated now

stephaniewhoo:
update:
`cat pyserini/ibm/result-ibm-0.1.json 
{"map": "0.2293", "ndcg": "0.2992"}`
same result when using Bert tokenized raw model

lintool (approved):
This is good enough for now. Let's get this merged in.

PR #895, opened 2021-12-03
Update Reproduction Log of DeepImpact

namespace-Pt:
Reproduced DeepImpact Model from scratch, learnt a lot in implementing sparse retrieval pipeline. Above all, I think both `anserini` and `pyserini` are great toolkits. Maybe more python features such as classes and apis can be introduced in `pyserini`.


lintool (changes requested):
Thanks for the PR.

The commit id doesn't appear to be valid - if you navigate to the URL, it says: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

Can you please fix?

namespace-Pt:
I don't think I can access the commit in your `pyserini` master branch since I'm only able to commit changes in my forked repository. Or maybe could you please tell me how to find the right commit link.

lintool:
Use the commit id in the main repo that you did your reproduction from?

lintool:
hi @namespace-Pt were you able to find the right id? e.g., see #897

lintool:
Find the commit id here: https://github.com/castorini/anserini/commits/master or via `git log` in your shell.

namespace-Pt:
@lintool I opened a new PR. Please check if it's correct.

PR #935, opened 2022-01-12
requirements update for spacy and lightgbm

stephaniewhoo:
+ add spacy 3.0.6 and lightgbm 3.2.1 to the requirement

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,5 +9,5 @@ transformers>=4.6.0
 sentencepiece>=0.1.95
 nmslib>=2.1.1
 onnxruntime>=1.8.1
-spacy==3.0.6
-lightgbm==3.2.1
+lightgbm>=3.3.2
+spacy>=3.2.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,5 @@ transformers>=4.6.0
 sentencepiece>=0.1.95
 nmslib>=2.1.1
 onnxruntime>=1.8.1
+spacy==3.0.6
+lightgbm==3.2.1


requirements.txt
```diff
@@ -9,3 +9,5 @@ transformers>=4.6.0
 sentencepiece>=0.1.95
 nmslib>=2.1.1
 onnxruntime>=1.8.1
+spacy==3.0.6
```
lintool, line 12: @MXueguang @ronakice @yuki617 - what do you think? should it be pegged to a version?

`>=` instead? Or just unspecified?

FWIW, my system:

```
lightgbm           3.3.2
spacy              3.2.1
```

stephaniewhoo: ah yes, I think `>=` is a better idea
if using `>=`, should we use newer version instead of my old ones

MXueguang: @ronakice  does it affect the 2.1.6 restriction for d2q?

MXueguang: I see. yeah then >= 3.2.1 would be good

lintool: So @stephaniewhoo let's do

```
lightgbm>=3.3.2
spacy>=3.2.1
```

ronakice: yeah, for v1 d2q with the old stuff we can just leave it and mention in the repository clearly that we used that.

yuki617: Yea, I believe spacy>=3.2.1 should be good.

lintool (approved):

lintool:
🚢 it!

PR #943, opened 2022-01-14
Fix test case URL in integrations/clprf/test_trec_covid_r3.py

lintool:
URL location no longer exists due to @rodrigonogueira4 's reorg of his Google storage data. Changing to NIST URL.


rodrigonogueira4 (approved):
LGTM!

PR #946, opened 2022-01-15
Fix the Core17 clprf Regression Test Failure 

yuki617:
* The rank function is changed in pandas 1.3.5 from 1.1.4 which caused the Value Error in the fusion part of clprf regression tests. Fixed it by only ranking one column one now.

### DIFF
--- a/pyserini/trectools/_base.py
+++ b/pyserini/trectools/_base.py
@@ -110,7 +110,7 @@ class TrecRun:
         self.run_data = pd.read_csv(filepath, sep='\s+', names=TrecRun.columns, dtype={'docid': 'str'})
         if resort:
             self.run_data.sort_values(["topic", "score"], inplace=True, ascending=[True, False])
-            self.run_data["rank"] = self.run_data.groupby("topic")[["docid","score"]].rank(ascending=False,method='first')
+            self.run_data["rank"] = self.run_data.groupby("topic")["score"].rank(ascending=False,method='first')
 
     def topics(self) -> Set[str]:
         """Return a set with all topics."""


pyserini/trectools/_base.py
```diff
@@ -110,7 +110,7 @@ def read_run(self, filepath: str, resort: bool = False) -> None:
         self.run_data = pd.read_csv(filepath, sep='\s+', names=TrecRun.columns, dtype={'docid': 'str'})
         if resort:
             self.run_data.sort_values(["topic", "score"], inplace=True, ascending=[True, False])
-            self.run_data["rank"] = self.run_data.groupby("topic")[["docid","score"]].rank(ascending=False,method='first')
+            self.run_data["rank"] = self.run_data.groupby("topic")["score"].rank(ascending=False,method='first')
```
lintool, line 113: Why wouldn't we have the same number of docid/score? Is there a deeper bug elsewhere?

yuki617: We do have the same number of docid and score, just .rank() should return one column instead of two columns. In version 1.1.4, [["docid","score"]].rank() returns the rank of the sorted score column, but in newer version it also returns the docid string column. Thus, with pandas==1.3.5, assigning two columns into one caused the ValueError: Columns must be same length as key

lintool: Ah, gotcha, thanks for the explanation!

lintool (approved):

pyserini/trectools/_base.py
yuki617: Thanks a lot! I'm still running a full regression test on clprf experiments with new packages version. Will merge it after all tests are done.

PR #956, opened 2022-01-17
LTR+IBM reranker refactoring

stephaniewhoo:
refactor to `pyserini.search.lucene.ltr` and `pyserini.search.lucene.tpprob` draft
@lintool This is just a draft, just want to confirm if this is the way you want first?
In terms of tpprob, we just have all codes in one main files. @yuki617 is wondering if you want it like ltr to have separate search class and the main file? I think it may be a bit overkill?

for the current `SimpleSearcher`, do we want things like `Pyserini.search.lucene.bm25`?


lintool:
> for the current `SimpleSearcher`, do we want things like `Pyserini.search.lucene.bm25`?

No, "bm25" would be a parameter and not a package.

PR #958, opened 2022-01-18
Fix regression test failure: TestSearchIntegration

yuki617:
*Fixed regression test failure: TestSearchIntegration.test_cross_validate.  It supposed to find the row index of max value by column instead of by row. In the lower pandas version it automatically find column max when there was only one column. Updated in the new pandas version.


lintool (approved):
Confirmed it works!

PR #960, opened 2022-01-19
Refactoring the TransProbability Reranker with pep8 format

yuki617:


scripts/reranker_ibm_colbert.py
```diff
@@ -70,23 +69,28 @@ def get_topics_from_qrun(path: str) -> Set[str]:
...
+def evaluate(qrels_path: str, run_path: str, options: str = ''):
     curdir = os.getcwd()
     if curdir.endswith('scripts'):
-       anserini_root = '../../anserini'
+        anserini_root = '../../anserini'
     else:
-       anserini_root = '../anserini'
-    prefix = f"{anserini_root}/tools/eval/trec_eval.9.0.4/trec_eval -c -M1000 -m all_trec {qrels_path}"
+        anserini_root = '../anserini'
+    prefix = f"{anserini_root}/tools/eval/trec_eval.9.0.4/trec_eval \
```
lintool, line 84: Can we break this dependency on anserini for eval? `trec_eval` exists... we can do something like `python -m pyserini.eval.trec_eval ...`

lintool (changes requested):

scripts/reranker_ibm_colbert.py
yuki617: The eval step is updated now.

lintool (approved):

PR #966, opened 2022-01-23
Reproduced DPR results

manveertamber:
Reproduced DPR results on orca

- Ran into same issue as @vivianliu0: "For NQ with DPR-Single, using DPR retrieval with brute-force index, the top 100 accuracy result I recorded was 0.8609 rather than 0.8610.". I got Top100 accuracy of 0.8609418282548477 to be exact instead of 0.8610.


ronakice:
I vote to just change this here like done here in the PR @lintool @MXueguang? Seems like the tests don't check it anyway...

ronakice (approved):

MXueguang:
yeah agree

lintool:
Ref #964 - same results as @vivianliu0

Issue #975, opened 2022-01-30
Test failure on pandas 1.4.0


lintool:
Unit tests fail on pandas 1.4.0:

```
./Users/jimmylin/workspace/pyserini/pyserini/trectools/_base.py:300: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
  res.run_data = res.run_data.append([df for df in dfs], ignore_index=True)
.F/Users/jimmylin/workspace/pyserini/pyserini/trectools/_base.py:300: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
  res.run_data = res.run_data.append([df for df in dfs], ignore_index=True)
....
======================================================================
FAIL: test_normalize_scores (tests.test_trectools.TestTrecTools)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jimmylin/workspace/pyserini/tests/test_trectools.py", line 77, in test_normalize_scores
    self.assertTrue(filecmp.cmp(os.path.join(self.root, 'tests/resources/simple_trec_run_normalize_verify.txt'),
AssertionError: False is not true

----------------------------------------------------------------------
```

Works fine on 1.3.5.

@yuki617 or @stephaniewhoo can you take a look? I think you're most familiar with this code?

yuki617:
No problem. I found out that the issue is caused by the new pandas version. Should I send a PR fixing it?
1. In pandas 1.4.0, frame.append method should be updated to pandas.concat method instead
2. When assigning float value like 1.0, 2.0 to the series, the score automatically changed into integer

lintool:
@yuki617 can you implement a fix that is forward compatible? that works for both 1.3.5 and 1.4.0?

yuki617:
Yes, just tested it, the fix works for both versions

lintool:
Please send PR, and ref this issue in your PR please.

PR #976, opened 2022-01-30
Fix test failure on test_trectools on pandas 1.4.0

yuki617:


yuki617:
Fixed the test failure issue mentioned here #975

lintool (approved):

Issue #917, opened 2021-12-28
Clean up sys.path.append


lintool:
We have a bunch of these idioms scattered throughout the code base:

```python
import sys
sys.path.append('.')
```

Since our installation instructions suggest we do:

```bash
$ pip install -e .
```

These can be cleaned up?

HAKSOAT:
Hi @lintool can I work on this? I think it's a good first task.

lintool:
Sure! Any contributions welcome!

lintool:
hi @HAKSOAT - if you're interested in helping out, there are other small tasks that we could use help on...

HAKSOAT:
Hi @lintool Yes, I am very interested. I'd appreciate it if you fill me in.

lintool:
Please ping me on email?

HAKSOAT:
Your uwaterloo email right?

lintool:
Yes please.

HAKSOAT:
Done.

PR #988, opened 2022-02-06
Cleaned up path appends to current directory

HAKSOAT:
This pull request addresses #917

### DIFF
--- a/scripts/ltr_msmarco/convert_passage.py
+++ b/scripts/ltr_msmarco/convert_passage.py
@@ -16,7 +16,6 @@
 
 import multiprocessing
 from joblib import Parallel, delayed
-import sys
 import json
 import argparse
 from transformers import AutoTokenizer, AutoModel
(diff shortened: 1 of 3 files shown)


HAKSOAT:
@lintool I worked on the cleanups. However, I noticed [this line](https://github.com/castorini/pyserini/blob/e194140521ba5477e0f4663decbae523ee4f4e2a/scripts/ltr_msmarco/convert_queries.py#L23). I didn't clean that up because of the comment right above it. Should I leave it?

lintool:
> @lintool I worked on the cleanups. However, I noticed [this line](https://github.com/castorini/pyserini/blob/e194140521ba5477e0f4663decbae523ee4f4e2a/scripts/ltr_msmarco/convert_queries.py#L23). I didn't clean that up because of the comment right above it. Should I leave it?

No, I think we should remove this one also... thanks!

HAKSOAT:
Great. I've removed that as well.

lintool (approved):

PR #989, opened 2022-02-07
Initial Screening

jh8liang:


lintool (changes requested):
Please note your machine specs (OS, etc.). Also indicate if you were successful or if you had any issues.

jh8liang:
@lintool sorry about that. The OS i ran the experiments on is Linux with python version 3.8.12 and Anaconda version 4.10.3. I was able to successfully replicate the experiments for Pyserini. I did not encounter any issue related to the document except when i was trying to follow the development installation guide i had to do "source activate pyserini-dev" instead of "conda activate pyserini-dev" which might just be my computer issue

jh8liang:
@lintool I re-tried "conda activate pyserini-dev" after re-starting today and it works. Sorry I don't think there was any issues at all

lintool (approved):

PR #990, opened 2022-02-07
Update replication log for uniCOIL

yuki617:
I have successfully replicated the uniCOIL result on my laptop with macOS version 11.5.2, python 3.7,  pandas version 1.3.5 and lightgbm 3.2.1.  No error was encountered during the replication.


lintool (changes requested):
Doesn't seem like a valid URL?
https://github.com/castorini/pyserini/commit/e03e06880ad4f6d67a1666c1dd45ce4250adc95d3

Can you describe your setup (machine, OS, etc.)? Did you have any issues reproducing the results?

lintool (approved):

PR #1013, opened 2022-02-18
Add script to convert DKRR embeddings to faiss IndexFlatIP form

manveertamber:
Script that takes embeddings output by https://github.com/facebookresearch/FiD/blob/main/generate_passage_embeddings.py and converts it to faiss IndexFlatIP form.


ronakice (approved):
LGTM!

PR #1023, opened 2022-02-20
Update command-line invocations for dense retrieval models

lintool:
+ Update indentation
+ General pass through docs, fixing issues here and there


lintool:
I think this is ready for review. Let's get this merged in before #1028 - and then do another PR to change `pyserini.dsearch` to `pyserin.search.faiss`.

I'd prefer multiple smaller changes over bigger ones.

MXueguang (approved):

Issue #1006, opened 2022-02-15
Reorg pyserini.dsearch -> pyserini.search.faiss


lintool:
Since `pyserini.search` has been moved to `pyserini.search.lucene`, `pyserini.dsearch` should become `pyserini.search.faiss` as well.

lintool:
`SimpleDenseSearcher` should become `FaissSearcher` as well.

lintool:
@crystina-z is going to take a stab at this.

crystina-z:
@lintool I guess this can be closed with #1028 merged?

PR #1028, opened 2022-02-21
dsearch -> search.faiss

crystina-z:
For issue #1006 

Now I've move the folder `dsearch` into `search.faiss` and renamed `SimpleDenseSearcher` into `FaissSearcher`. 
all unittests has passed, still waiting for the pytest results.

lmk if I missed any files / conditions that need to be updated. also I hvn't added the fakes class and interface to make old API call works. wonder for which classes we'll wanna add this?


lintool:
`integrations/dense/temp_cacm/cacm.json` shouldn't be part of the PR I think?

> also I hvn't added the fakes class and interface to make old API call works. wonder for which classes we'll wanna add this?

We should add facade classes to make the old `python -m pyserini.dsearch` driver work. Can you do that please? Like this:
https://github.com/castorini/pyserini/blob/master/pyserini/search/__main__.py

crystina-z:
thanks! fixed the two above.
also wonder which classes should be included in the "dsearch/_deprecated.py" like this file - https://github.com/castorini/pyserini/blob/master/pyserini/search/_deprecated.py? e.g. do we need only the `xxSearcher` classes [here](https://github.com/castorini/pyserini/blob/master/pyserini/dsearch/__init__.py#L22-L23), or also the Encoders and other classes?

lintool:
> thanks! fixed the two above. also wonder which classes should be included in the "dsearch/_deprecated.py" like this file - https://github.com/castorini/pyserini/blob/master/pyserini/search/_deprecated.py? e.g. do we need only the `xxSearcher` classes [here](https://github.com/castorini/pyserini/blob/master/pyserini/dsearch/__init__.py#L22-L23), or also the Encoders and other classes?

Yes, let's mirror @MXueguang 's PR as closely as possible - add stubs for the deprecated?

crystina-z:
facade classes added and unittest passed

lintool (approved):

MXueguang (approved):
lgtm

PR #1033, opened 2022-02-23
Add to repro log of DKRR

lintool:
I rank through repro docs, tweaking them along the way.


docs/experiments-dkrr.md
```diff
@@ -1,79 +1,107 @@
...
-    --topk 5 20 100 500 1000
+  --retrieval runs/nq.dkrr.ans.test.json \
+  --topk 5 20 100 500 1000
+```
+
+The expected results are as follows, shown in the "ours" column:
+
+| Metric   |  ours | orig |
+|:---------|------:|-----:|
+| Top-5    | 84.27 |
```
manveertamber, line 48: Swapped Top-5 and Top-20 results

PR #1037, opened 2022-02-23
Mrtydi mbert prep and index

crystina-z:
The script to prepare the mbert tokenized Mrtydi, and the script to index and search in it.

`scripts/mrtydi/mbert-tokenize-mrtydi.py` is just for reference, `window-combined-corpus.sh` would download the prepared corpus and index on it. and `search.py` help to play around in the built index


lintool (approved):
quick look; looks reasonable. these are just one-off scripts, so no need to be CRed to the same level of quality.

🚢 it!

crystina-z:
one sec I think I missed one file, gonna update it quick

PR #1044, opened 2022-02-24
Refactor vsearch

MXueguang:


pyserini/search/nmslib/__init__.py
```diff
@@ -14,6 +14,6 @@
 # limitations under the License.
 #
 
-from ._vsearcher import SearchResult, SimpleVectorSearcher
+from ._searcher import SearchResult, NmslibSearcher
 
-__all__ = ['SearchResult', 'SimpleVectorSearcher']
+__all__ = ['SearchResult', 'NmslibSearcher']
```
lintool, line 19: `NmslibSearcher` is correct, but looks _ugly_!

pyserini/vsearch/index.py
```diff
@@ -1,101 +0,0 @@
-#
```
lintool, line 1: where did this go?

lintool:
Should we add a test case like https://github.com/castorini/pyserini/blob/master/integrations/dense/test_encode.py ?

We're going to need to use `nmslib` at some later point in time... so let's not accrue technical debt?

pyserini/search/nmslib/__init__.py
MXueguang: hrm... have to match Lucene and Faiss i guess

pyserini/vsearch/index.py
MXueguang: In pyserini index nmslib already https://github.com/castorini/pyserini/blob/master/pyserini/index/nmslib.py

MXueguang:
> Should we add a test case like https://github.com/castorini/pyserini/blob/master/integrations/dense/test_encode.py ?
> 
> We're going to need to use `nmslib` at some later point in time... so let's not accrue technical debt?

I'll create an issue to track and deal with the test in seperate PR. it will take some time to add a proper integration tests

lintool (approved):

PR #1046, opened 2022-02-24
Clean up LTR and IRST integration test case

yuki617:
* Clean up LTR and IRST test case with PEP8 compliant
* Rename tprob into IRST
* Add additional evaluation step for IRST integration test.


yuki617:
@stephaniewhoo Could you take a look and run the three integration tests on your side when you get a chance?

integrations/sparse/test_lucenesearcher_check_irst.py
```diff
@@ -0,0 +1,94 @@
...
+        os.system(f'wget {ibm_model_url} -P irst_test/')
+        os.system(f'tar -xzvf irst_test/{ibm_model_tar_name} -C irst_test')
+        # queries process
+        os.system('python scripts/ltr_msmarco/convert_queries.py \
+            --input tools/topics-and-qrels/topics.msmarco-passage.dev-subset.txt \
+            --output irst_test/queries.dev.small.json')
+        # qrel
+        self.qrels_path = f'{self.pyserini_root}/tools/topics-and-qrels/qrels.msmarco-passage.dev-subset.txt'
+
+    def test_ibm_reranking(self):
```
lintool, line 44: I think we called it sum and max aggregation?

integrations/sparse/test_ltr_msmarco_document.py
```diff
@@ -32,33 +29,53 @@ def test_reranking(self):
         inp = 'run.msmarco-pass-doc.bm25.txt'
         outp = 'run.ltr.msmarco-pass-doc.test.trec'
         outp_tsv = 'run.ltr.msmarco-pass-doc.test.tsv'
-        #Download prebuilt index
-        #retrieve candidate
+        # Download prebuilt index
```
yuki617, line 32: @stephaniewhoo Should we download the index somewhere or change the index below into pre-built index name?

pyserini/search/lucene/irst/__main__.py
```diff
@@ -17,7 +17,7 @@
 import json
 import os
 import subprocess
-from pyserini.search.lucene.tprob import TranslationProbabilitySearcher
+from pyserini.search.lucene.irst import TranslationProbabilitySearcher
```
lintool, line 20: Should we call it `LuceneIrstSearcher` to be more accurate?

Below, probably makes sense to call:
`pyserini/search/lucene/irst/_search_irst.py` -> `_searcher.py`? I think this makes it consistent with lucene package?

lintool (changes requested):

integrations/sparse/test_ltr_msmarco_document.py
lintool: use pre-built.

integrations/sparse/test_lucenesearcher_check_irst.py
```diff
@@ -19,7 +19,7 @@
 import unittest
 
 
-class TestLtrMsmarcoPassageIbm(unittest.TestCase):
+class TestLtrMsmarcoPassageIRST(unittest.TestCase):
```
lintool, line 22: I think this is misnamed? There's no LTR here right?

`TestMsmarcoPassageIrst`?

Note `Irst` not `IRST` per standard naming rules.

lintool (approved):

integrations/sparse/test_lucenesearcher_check_irst.py
yuki617: yes, sorry, that's a typo

Issue #1039, opened 2022-02-24
Update DKRR documentation


lintool:
Add instructions for running on dev set and comparisons to dev set numbers in the paper.

lintool:
Closed by #1066, further updated in #1069

PR #1066, opened 2022-03-06
Update DKRR doc

manveertamber:
- Added comparison scores from original paper
- Added support for encoded-queries
- Replaced pyserini.dsearch with pyserini.search.faiss


lintool (approved):

lintool:
Closes #1039

PR #1071, opened 2022-03-07
Dense encode doc

crystina-z:
Doc refactor as mentioned in #1063 

The original index.md contents are merged into encode.sh (sharding and searching).
Examples to convert embedding into faiss-index are added to index.md

PR #1077, opened 2022-03-10
Readme testcase 

crystina-z:
- unit test for the added command line in readme (#1071)
- moved the `integration/resources` to `tests`


lintool (changes requested):
There are references to `integrations/resources/` in README.md. Please change?

lintool:
@crystina-z additional things I noticed:

Screen Shot 2022-03-10 at 7 08 50 AM

Why is 4 even an option? It's a no-op? cc @MXueguang 

1. Should we just kill the code path?
2. Can you add simple test cases for options 1-3?

Also, there's a `dsearch` reference that needs to be replaced.

tests/test_index.py
```diff
@@ -0,0 +1,130 @@
...
+                        --fields text \
+                        --batch 1 \
+                        --device cpu'
+        status = os.system(cmd)
+        self.assertEqual(status, 0)
+        self.assertIsFile(os.path.join(encoded_corpus_dir, 'docid'))
+        self.assertIsFile(os.path.join(encoded_corpus_dir, 'index'))
+        return encoded_corpus_dir
+
+    # def test_faiss_hnsw_pq(self):
```
crystina-z, line 60: still working on this one, seems it's having problem with train() using only 3 data points.

crystina-z:
> Should we just kill the code path?

I'm fine with both, tho seems that allow users to save encoded collections in `.jsonl` format doesn't sound bad either? given that they _can_ choose to save into FaissFlat directly

tests/test_index.py
lintool: no need to try too hard... if it doesn't work, just axe it.

lintool:
re: "to faiss" code path - exporting back out to jsonl is useful, but isn't "indexing" per se. Shouldn't it be like a separate "export" program? @MXueguang what do you think?

MXueguang:
Since we have the option to store dense embedding as `jsonl` format during encoding, we need a way to convert the `jsonl` dense embeddings to the flat index that we can conduct search. It seems to be an 'indexing' process. 
Converting faissflat to faissflat is no-op, so I guess we can just update the readme for 4, saying that is for indexing `jsonl` to faissflat?

lintool:
sgtm

lintool (approved):

PR #1091, opened 2022-03-24
add test for nmslib sparse hnsw index

MXueguang:


integrations/sparse/test_nmslib.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # Hard-code larger values for internal servers
+        if socket.gethostname().startswith('damiano') or socket.gethostname().startswith('orca'):
+            self.threads = 36
+            self.batch_size = 144
+
+    def test_msmarco_passage_deepimpact_nmslib_hnsw(self):
+        output_file = 'test_run.msmarco-passage.deepimpact.nmslib.tsv'
+        self.temp_files.append(output_file)
+        cmd = 'wget https://www.dropbox.com/s/7rp56owflmfpi7s/deepimpact_msmarco_passage_dev_topic.jsonl'
```
lintool, line 40: let's not use Dropbox? Let's stash with the other queries?

integrations/sparse/test_nmslib.py
```diff
@@ -0,0 +1,66 @@
...
+        status = os.system(cmd)
+        cmd = 'wget https://www.dropbox.com/s/d7flii2tglf3l9m/nmslib-index.msmarco-passage.deepimpact.20211012.58d286.tar.gz'
```
lintool, line 42: Tell me where the tarball is on `orca` and I'll lift into S3 for you?

lintool (changes requested):
Let's make the URLs more stable?

lintool (approved):

Issue #1067, opened 2022-03-06
DPR accuracy has too many figures of precision


lintool:
Right now we have something like:

```
python -m pyserini.eval.evaluate_dpr_retrieval \
  --retrieval runs/run.dpr.nq-test.single.bf.bm25.json \
  --topk 20 100
```

Which gives:

```
Top20	accuracy: 0.8288088642659279
Top100	accuracy: 0.8836565096952909
```

There's no need to have that many figures of precision. Four digits is sufficient.

PR #1092, opened 2022-03-26
Reduced DPR acccuracy's figures of precision

HAKSOAT:
Fixes issue #1067 

I was not able to test this fix. I tried to download the index needed to get `runs/run.dpr.nq-test.single.bf.bm25.trec` and its DPR format before running the command:

> python -m pyserini.eval.evaluate_dpr_retrieval \
  --retrieval runs/run.dpr.nq-test.single.bf.bm25.json \
  --topk 20 100

However, the index was quite large (55GB). I believe the fix should work though as it's not a big change.

Are there much smaller indexes available that make it easier to test out functionalities and fixes quickly?


lintool (approved):
I've verified, works as expected.

PR #1098, opened 2022-03-28
Improved error message for RM3 expansion doc vector issue

### LINKED ISSUE #695: Better error message when index doesn't store doc vectors to support RM3 expansion
This allows us to use RM3 query expansion.

HAKSOAT:
Fixes #695 

Raises an error when an attempt is made at setting RM3 for indexes without document vectors.

### DIFF
--- a/pyserini/search/lucene/_searcher.py
+++ b/pyserini/search/lucene/_searcher.py
@@ -228,7 +228,10 @@ class LuceneSearcher:
         rm3_filter_terms: bool
             Whether to remove non-English terms.
         """
-        self.object.setRM3(fb_terms, fb_docs, original_query_weight, rm3_output_query, rm3_filter_terms)
+        if self.object.reader.getTermVectors(0):
+            self.object.setRM3(fb_terms, fb_docs, original_query_weight, rm3_output_query, rm3_filter_terms)
+        else:
+            raise TypeError("RM3 is not supported for indexes without document vectors.")
 
     def unset_rm3(self):
         """Disable RM3 query expansion."""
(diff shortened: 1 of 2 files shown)


lintool (commented):
Would it be possible to get a test case? We have existing test cases that create stub indexes - you can hook into that?

HAKSOAT:
@lintool I have not used Lucene before, but I made the assumption here that all indexes will have a doc id of 0 (as the first document), is this assumption correct?

Also, I thought of possibly adding a test case here: https://github.com/castorini/pyserini/blob/b71dbc6dca04b214f95cf609390c3294976f8e8b/tests/test_search.py#L192

But:

1. I wasn't so sure if the index loaded in that test suite has vectors (I should actually confirm myself)
2. If it has vectors, how can I possibly delete them before running the test?

HAKSOAT:
> Would it be possible to get a test case? We have existing test cases that create stub indexes - you can hook into that?

Aha. Answers my question already. I'll take a look :-)

lintool:
Take a look at `tests/test_index_reader.py` - it builds an index as part of the test. You can modify it to build two versions - one storing doc vectors, one not...

Thanks!

HAKSOAT:
Added a test. Any improvements in mind?

lintool (approved):

lintool:
Hi @HAKSOAT I just realized:

> The [setUp()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) and [tearDown()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown) methods allow you to define instructions that will be executed before and after each test method.

https://docs.python.org/3/library/unittest.html

So your implementation needlessly builds the index multiple times... which actually means that the original test case was poorly designed also...

I think the right approach is this:

> https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

Would you mind fixing?

HAKSOAT:
True. I'll do that.

Issue #1112, opened 2022-04-07
pyserini.encode


joaocp98662:
I'm trying to build dense indexes on my document collections but it raises an exception. The command I'm using is:

python -m pyserini.encode \
input   --corpus ... 
output  --embeddings ...
        --to-faiss \
encoder --encoder sentence-transformers/msmarco-bert-base-dot-v5 \

And it outputs:

375580it [00:10, 36280.28it/s]
  0% 0/375580 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.7/dist-packages/pyserini/encode/__main__.py", line 103, in <module>
    embeddings = encoder.encode(**kwargs)
  File "/usr/local/lib/python3.7/dist-packages/pyserini/encode/_auto.py", line 52, in encode
    sklearn.preprocessing.normalize(embeddings, axis=1)
AttributeError: module 'sklearn' has no attribute 'preprocessing'

crystina-z:
Hi @joaocp98662 thanks for your interest in our repo! It seems like the import of sklearn has been changed, and the PR [here](https://github.com/castorini/pyserini/pull/1114) is supposed to fix it. Could u try it and let us know if that solves the issue for you?

joaocp98662:
Hi @crystina-z thank you for your help. I'm using Colab and tried to upgrade Pyserini using pip. Then I tried to run Pyserini.encode again and the error persists. Maybe I'm doing something wrong, I don't know how to push your changes in the environment I'm using in other way that doesn't use a package manager.

lintool:
hi @joaocp98662 you'll have to build a dev installation, following instructions here:
https://github.com/castorini/pyserini/blob/master/docs/installation.md#development-installation

I'm going to close the issue for now... please reopen if you have additional questions/issues.

PR #1114, opened 2022-04-07
sklearn.preprocessing.normalize -> from sklearn.preprocessing import …

crystina-z:
For issue #1112

### DIFF
--- a/pyserini/encode/_auto.py
+++ b/pyserini/encode/_auto.py
@@ -14,8 +14,8 @@
 # limitations under the License.
 #
 
-import sklearn
 import numpy as np
+from sklearn.preprocessing import normalize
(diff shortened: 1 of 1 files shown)


lintool (approved):

lintool:
hey @crystina-z any idea why this wasn't caught by our unit tests? Is it a sklearn version issue?

crystina-z:
@lintool not really a sklearn version issue I think. the line is only triggered when the `self.l2_norm` is set to True, which only happens when using SentenceTransformer - I assume we don't have a test case for that?

where the SentenceTransformer calls this function - https://github.com/castorini/pyserini/blob/master/pyserini/encode/__main__.py#L33

PR #1118, opened 2022-04-18
Update irst result table

stephaniewhoo:
+ modify irst table passage result
+ waiting @yuki617  on doc seg results


lintool (approved):
Matches what I have!

PR #1126, opened 2022-04-25
Reproduce MS MARCO passage and document ranking experiments results and fix unit tests failure

jx3yang:
## Description

Reproduced the MS MARCO passage and document ranking experiments results as part of [initial screening](https://github.com/lintool/guide/blob/master/ura.md#initial-screening).

## Environment
- Java version 11.0.15
- Maven version 3.8.5
- Python version 3.8.13 with Conda version 4.11.0
- MacOS Monterey version 12.3.1 (21E258)

## Issue Encountered
When running the unit tests with `python -m unittest` for the first time, the following tests failed:
- `test_tct_colbert_v2_encoder_cmd`
- `test_tct_colbert_v2_encoder_cmd_shard`
- `test_faiss_hnsw`
- `test_faiss_pq`

This was due to the cache directories not being properly created, particularly `~/.cache/pyserini/temp_index` and `~/.cache/pyserini/temp_index-0`. The tests use `os.mkdir` to create the directories, which succeeds only when all the intermediate directories exist. In this case, the directory `~/.cache/pyserini` is likely to not exist the first time the tests are run, resulting in the failure of the tests when they try to access a nonexistent directory.

The fix is to change `os.mkdir` to `os.makedirs`, the latter being equivalent to `mkdir -p`, which creates the intermediate directories if they are not present.


lintool (approved):
Thanks for the catch!

PR #1130, opened 2022-04-30
Refactor IRST script

yuki617:
* Change bm25 search into batch_search
* Auto downloading ibm model/ term frequency table to ~/.cache/pyserini/translation_model


integrations/sparse/test_lucenesearcher_check_irst.py
```diff
@@ -67,18 +59,16 @@ def test_sum_aggregation_dl19_passage(self):
         self.assertEqual(ndcg_score, 0.5260)
 
     def test_sum_aggregation_dl20_passage(self):
-        #dl20 passage
+        #d l20 passage sum
```
lintool, line 62: typo here.

pyserini/search/lucene/irst/_searcher.py
```diff
@@ -87,6 +89,55 @@ def from_prebuilt_index(cls, prebuilt_index_name: str):
         print(f'Initializing {prebuilt_index_name}...')
         return cls(index_dir)
 
+    def download_and_unpack_translation_model(self):
```
lintool, line 92: there seems to be a lot of code duplication here with `pyserini/util.py` - is it really necessary? why can't you just reuse the existing methods?

lintool (changes requested):

pyserini/search/lucene/irst/_searcher.py
yuki617: All CR fixed.

lintool (approved):

PR #1139, opened 2022-05-03
Fix bug: parsing results of trec_eval

lintool:
Ignore the first few lines of output until we get to 'Results'; then start parsing.


yuki617 (approved):
Lgtm! Confirmed that irst unittest still pass after this change.

PR #1146, opened 2022-05-04
Add to repro log for GAR-T5 experiments

manveertamber:
There were some differences in the top 300 scores. The other scores matched perfectly. I have updated the doc accordingly.


lintool:
Fix nits and merge. I don't need to see it again.

manveertamber:
> Fix nits and merge. I don't need to see it again.

I don't think I have write access to merge

PR #1148, opened 2022-05-04
not removing trailing \n in contents when encoding, as it is probably…

crystina-z:
… the delimiter before an empty field; add field length checking

To solve the issue that @NThakur20 met when encoding SciFact documents

### DIFF
--- a/pyserini/encode/_base.py
+++ b/pyserini/encode/_base.py
@@ -88,6 +88,25 @@ class JsonlCollectionIterator:
                 to_yield[key] = self.all_info[key][idx: min(idx + self.batch_size, end_idx)]
             yield to_yield
 
+    def _parse_fields_from_contents(self, contents):
+        """
+        :params contents: str, the text containing all fields as speicifed in self.fields
+        This function will parse the input contents into each fields based the self.delimiter
+        return: List, each corresponds to the value of self.fields
+        """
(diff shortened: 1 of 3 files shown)


lintool:
@crystina-z can we write a test case?

NThakur20:
Tried out locally, works for SCIDOCS and FiQA.
This helps to encode datasets such as SCIDOCS or FIQA, where the title might be present but the text empty. 

Kind Regards,
Nandan Thakur

pyserini/encode/_base.py
```diff
@@ -88,6 +88,25 @@ def __iter__(self):
                 to_yield[key] = self.all_info[key][idx: min(idx + self.batch_size, end_idx)]
             yield to_yield
 
+    def _parse_fields_from_contents(self, contents):
```
crystina-z, line 91: adding a separate function to process the contents, deciding if we should remove the trailing \n. 
As it seems CACM has a trailing \n for all documents

crystina-z:
and added the test case

lintool:
hey @NThakur20 can you sign off on this?

NThakur20 (approved):
Looks good to me!

Kind Regards,
Nandan Thakur

crystina-z:
`python -m unittest` passed

PR #1149, opened 2022-05-04
Fixing judge testcase run's path

vjeronymo2:


lintool:
Ooops, I found and fixed it in #1150 ... closing.

Issue #1158, opened 2022-05-09
Check output of trec_eval compute judged@k


lintool:
`trec_eval` inside pyserini has recently been updated to include computation of judged@k - we should cross reference to see if they produce the same results as: https://github.com/castorini/anserini-tools/blob/master/eval/measure_judged.py

they should... but we need to verify.

lintool:
Closed by #1184.

PR #1159, opened 2022-05-09
Fix broken tests related to trec_eval computing unjudged

lintool:


rodrigonogueira4 (approved):

rodrigonogueira4:
Hey @lintool, now that we removed the judged@10/20 from the output, `def test_undjudged_keep()` is not testing the judge calculation anymore, right?

lintool:
yes, it's just a generic "baseline" test

rodrigonogueira4:
But shouldn't we also test the judged calculations here?

lintool:
Yea, I was going to circle around and look at it in detail later: noted as part of #1158

I was planning on doing some refactoring, to make consistent with other metrics, something like `-m judged.10,20`

rodrigonogueira4:
Ok!

PR #1164, opened 2022-05-10
Fix broken clprf regressions

lintool:
+ use internal trec_eval
+ change to consistent use of f-strings


yuki617 (approved):
Looks good to me.

PR #1165, opened 2022-05-10
Fix broken regressions for ANCE

lintool:
+ Add explicit parameter for ANCE-prf
+ Add different methods for parsing output of different eval utils

### DIFF
--- a/integrations/utils.py
+++ b/integrations/utils.py
@@ -42,6 +42,7 @@ def run_command(cmd, echo=False):
     return stdout, stderr
 
 
+# Function for parsing the output from pyserini.eval.trec_eval
 def parse_score(output, metric, digits=4):
     lines = output.split('\n')
     # The output begins with a bunch of debug information, get rid of lines until we get to 'Results'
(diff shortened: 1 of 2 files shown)


integrations/utils.py
```diff
@@ -53,3 +54,22 @@ def parse_score(output, metric, digits=4):
...
+    return None
+
+
+# Function for parsing the output from MS MARCO eval scripts
+# (Currently, parses the same way as parse_score_qa above, but keeping separate in case they diverge in the future.)
+def parse_score_msmarco(output, metric, digits=4):
+    for line in output.split('\n'):
+        if metric in line:
+            score = float(line.split()[-1])
+            return round(score, digits)
```
rodrigonogueira4, line 74: nit: why do we need rounding here?

rodrigonogueira4 (approved):

integrations/utils.py
lintool: Due to floating point precision issues, Python might parse string `0.2000` and internally it becomes 0.19999999. We make sure this doesn't happen...

rodrigonogueira4: Got it, thanks!

PR #1167, opened 2022-05-12
Add support for quantizing weights in BM25 vectors dump

manveertamber:


lintool:
hey @manveertamber - I don't think this approach will scale if the corpus becomes too big? It looks like you're storing the vectors in memory to locate the max? You'll run out of memory...

lintool:
Thinking about this a bit, I think quantization should be a separate program... so that it can quantize _any_ float vectors...

So we run two programs: generate BM25 vectors, and then run quanitzation.

Thoughts @manveertamber ?

manveertamber:
I think that makes sense. So where would the quantization code go?

manveertamber:
> hey @manveertamber - I don't think this approach will scale if the corpus becomes too big? It looks like you're storing the vectors in memory to locate the max? You'll run out of memory...

Ok that should be fixed then. The existing code even before the quantization saves the vectors in a list to dump to the json at the end for easy formatting. @lintool

lintool:
something like `python -m pyserini.index.quantize_vectors`?

lintool:
> > hey @manveertamber - I don't think this approach will scale if the corpus becomes too big? It looks like you're storing the vectors in memory to locate the max? You'll run out of memory...
> 
> Ok that should be fixed then. The existing code even before the quantization saves the vectors in a list to dump to the json at the end for easy formatting. @lintool

Yea, I must have missed in my previous CR. My bad.

pyserini/index/lucene/_base.py
```diff
@@ -539,22 +540,68 @@ def stats(self) -> Dict[str, int]:
...
+            for term in self.get_document_vector(docid):
+                bm25_vector[term] = self.compute_bm25_term_weight(docid, term, analyzer=None, k1=k1, b=b)
+
+            # vectors are written line by line to avoid running out of memory
+            f.write(json.dumps({'id': docid, 'vector': bm25_vector}) + "\n")
+
+        f.close()
+
+    def quantize_weights(self, input_file_path, output_file_path, bits = 8):
+        """Takes vectors of weights in Pyserini's JSONL vector format and quantizes them
```
lintool, line 570:
```suggestion
        """Takes vectors of weights in Pyserini's JSONL vector format and quantizes them.
```

pyserini/index/lucene/_base.py
```diff
@@ -539,22 +540,68 @@ def stats(self) -> Dict[str, int]:
         return index_stats_dict
 
     def dump_documents_BM25(self, file_path, k1=0.9, b=0.4):
-        """Dumps out all the document vectors with BM25 weights in Pyserini's JSON vector format 
+        """Dumps out all the document vectors with BM25 weights in Pyserini's JSONL vector format
```
lintool, line 543:
```suggestion
        """Dumps out all the document vectors with BM25 weights in Pyserini's JSONL vector format.
```

lintool (approved):

PR #1172, opened 2022-05-13
mrtydi test (for untied-nq)

crystina-z:
Add test case for mrtydi dense results. as in slack discussion

### DIFF
--- a/integrations/dense/test_mrtydi.py
+++ b/integrations/dense/test_mrtydi.py
@@ -1,0 +1,66 @@
+#
+# Pyserini: Reproducible IR research with sparse and dense representations
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Integration tests for Mr TyDi results using pre-encoded queries."""
+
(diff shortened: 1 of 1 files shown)


integrations/dense/test_mrtydi.py
```diff
@@ -0,0 +1,66 @@
...
+                                --threads {self.threads} \
+                                --encoder castorini/mdpr-question-nq'
+            cmd2 = f'python -m pyserini.eval.trec_eval -c -M 100 -m recip_rank -c -m recall.100 mrtydi-v1.1-{lang}-test {output_file}'
+ 
+            status1 = os.system(cmd1)
+            stdout, stderr = run_command(cmd2)
+            score = parse_score(stdout, 'recip_rank')
+ 
+            self.assertEqual(status1, 0)
+            self.assertAlmostEqual(score, expected_score, delta=0.01)
```
lintool, line 66: smaller delta? Start with 0.001.

If it becomes too tight, we can loosen it.

lintool (changes requested):
You want a teardown method as well:

```
    def tearDown(self):
        clean_files(self.temp_files)


if __name__ == '__main__':
    unittest.main()
```

integrations/dense/test_mrtydi.py
crystina-z: ah for some language the difference is somehow slightly over 0.005, I'll see what's the tightest bound we can have now

crystina-z:
updated, using delta=0.0012 as the largest gap observed is on telugu; (0.1099 v.s 0.111)

lintool:
> updated, using delta=0.0012 as the largest gap observed is on telugu; (0.1099 v.s 0.111)

I think you're missing a zero?

crystina-z:
oh I meant the previous score was 0.111 on Telugu(after rounding), and now we get 0.1099,  and the delta is slightly over 0.0011 so was setting it as 0.0012 here

lintool (approved):

Issue #370, opened 2021-02-11
Multi-thread scripts/dpr/convert_trec_run_to_retrieval_json.py


lintool:
Takes a non-trivial amount of time to run for standard query sets... seems like we can benefit from multi-threading here?

MXueguang:
yeah

MXueguang:
`convert_trec_run_to_retrieval_json.py` currently fetch raw texts from sparse-index by `pyserini.search.SimpleSearcher.doc()`. 
We have to construct multiple `searcher` in this case (if we want multi-threading)?

MXueguang:
another solution maybe we load all doc in the collections and create a docid-rawtext dictionary. And then mutithread access the dictionary.

lintool:
I see, the issue is that `.doc` has no "batch" version, huh? This will need to be done on the Java side... a bit more involved.

We should punt on this for now?

MXueguang:
yeah, improve the java end makes more sense.
we should punt it now and do this for future release.

lintool:
hi @manveertamber I believe you were going to take this on?

lintool:
More details: https://github.com/castorini/onboarding/blob/master/docs/cc-guide.md

The key is `searcher.doc`. As part of https://github.com/castorini/anserini/issues/1778 we now have a more efficient batch method. So, replace `.doc` with that more efficient method.

The result: QA eval should be more efficient (faster), but the results should be exactly the same.

PR #1178, opened 2022-05-18
Add multi-thread support for scripts/dpr/convert_trec_run_to_retrieval_json.py

manveertamber:
For https://github.com/castorini/pyserini/issues/370

Controlling for everything else it seems using searcher.batch_doc is slower than using searcher.doc. That is to say, I have found using more than one thread leads to this script running slower. However, increasing the number of threads beyond 2 does seem to help.

### DIFF
--- a/pyserini/demo/miracl.py
+++ b/pyserini/demo/miracl.py
@@ -15,111 +15,135 @@
 #
 
 """
-This script provides a web interface demo for retrieval on the MIRACL dataset.
+This script provides an interactive web interface demo for retrieval on the MIRACL dataset.
 It requires `flask` (`pip install flask~=2.2.0`).
-An example command looks like `python -m pyserini.demo.miracl --language en` that starts up a server on port 8080.
+An example command looks like `python -m pyserini.demo.miracl` that starts up a server on port 8080.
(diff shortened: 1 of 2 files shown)


manveertamber:
> @manveertamber can you run some concrete performance measurements? E.g.,
> 
>     * previous impl
> 
>     * new impl with 1, 2, 4, 8, ... threads
> 
> 
> non-linear scaling is to be expected, I think, since there is inevitably contention for the underlying data structures...

Running this on orca:
```
nohup python -m pyserini.eval.convert_trec_run_to_dpr_retrieval_run \
    --index wikipedia-dpr \
    --topics dpr-nq-test \
    --input runs/run.dpr.nq-test.bm25.trec \
    --output runs/run.dpr.nq-test.bm25.json \
```
* previous impl: 
19 minutes 13 seconds

* new impl: 
19:16 (1 thread, uses searcher.doc)
52:11 (2 threads)
43:54 (4 threads)
37:42 (8 threads)

lintool:
Ack. Let's put this issue on pause.

@HAKSOAT has confirmed that the slowdown happens on the Java end also. We're trying to diagnose why this is the case.

PR #1191, opened 2022-05-29
Update doc for MS MARCO uniCOIL experiments

lintool:


lintool:
@ronakice @rodrigonogueira4 Ready for review.

Primary change is to bring pages into alignment with repro matrix and results presented in SIGIR 2022 paper.

rodrigonogueira4 (approved):

PR #1205, opened 2022-06-09
Expose Rocchio feedback in Pyserini

yuki617:


lintool (changes requested):
Please add tests? https://github.com/castorini/pyserini/blob/master/tests/test_search.py

yuki617:
> Please add tests? https://github.com/castorini/pyserini/blob/master/tests/test_search.py

No problem, will add some tests for both anserini and pyserini env.

lintool (approved):

PR #1207, opened 2022-06-14
add es scripts for unicoil

MXueguang:
add elasticsearch script for unicoil/bm25.

### DIFF
--- a/scripts/es/check_status.py
+++ b/scripts/es/check_status.py
@@ -1,3 +1,18 @@
+#
+# Pyserini: Reproducible IR research with sparse and dense representations
+#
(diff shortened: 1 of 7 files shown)


scripts/es/check_status.py
```diff
@@ -0,0 +1,14 @@
+from elasticsearch import Elasticsearch
+
+# Create the client instance
+client = Elasticsearch("http://localhost:9200")
+
+resp = client.search(index="example_index", query={"bool": {
```
lintool, line 6: Let's actually give the index a "proper" name - e.g., `msmacro-v1-passage`?

lintool (changes requested):

PR #1218, opened 2022-06-22
Add negative feedback option to Rocchio 

yuki617:
* Add negative label to Rocchio reranker
* Add tests for Rocchio with negative feedback


pyserini/search/lucene/__main__.py
```diff
@@ -88,6 +88,7 @@ def define_search_args(parser):
 
     parser.add_argument('--rm3', action='store_true', help="Use RM3")
     parser.add_argument('--rocchio', action='store_true', help="Use Rocchio")
+    parser.add_argument('--usenegative', action='store_true', help="Use nonrelevant labels in Rocchio")
```
lintool, line 91: change to `--use-negative` please.

pyserini/search/lucene/__main__.py
```diff
@@ -182,7 +183,10 @@ def define_search_args(parser):
     
     if args.rocchio:
         search_rankers.append('rocchio')
-        searcher.set_rocchio()
+        if args.usenegative:
+            searcher.set_rocchio(rocchio_gamma=0.15, rocchio_use_negative=True)
```
lintool, line 187: Do we need `rocchio_` prefix here? Method already disambiguates... right?

lintool (changes requested):
Minor comments? Thoughts?

pyserini/search/lucene/__main__.py
yuki617: Will --rocchio-use-negative be better? To be more specific for rocchio reranking.

PR #1227, opened 2022-06-28
Update tct_colbert-v2 documentation

yuki617:
* Replicate the results for tct_colbert-v2 on msmarco-doc 
* Add scores for DL20-doc topic


lintool:
@MXueguang can you look at this please?

docs/experiments-tct_colbert-v2.md
```diff
@@ -274,7 +274,7 @@ $ python scripts/tct_colbert/merge_indexes.py \
 Step4: search (with on-the-fly query encoding)
```
MXueguang, line 274: I think we can remove step0 to step3 as we are using prebuilt index now

MXueguang (changes requested):

docs/experiments-tct_colbert-v2.md
yuki617: No problem. I've removed those steps. Also, I found out the score I got is a little bit different from the one in documentation. Maybe it's because one is by pre-build index another is self-build one? Here is the one I got:

$ python -m pyserini.eval.msmarco_doc_eval --judgments msmarco-doc-dev --run runs/run.msmarco-doc.passage.tct_colbert-v2-hnp-maxp.txt
#####################
MRR @100: 0.35079861563936404
QueriesRanked: 5193
#####################

$ python -m pyserini.eval.trec_eval -c -mrecall.100 -mmap -mndcg_cut.10 msmarco-doc-dev 
map                   	all	0.3508
recall_100            	all	0.8908
ndcg_cut_10           	all	0.4121

MXueguang (approved):

PR #1233, opened 2022-07-11
Reporoduction Log Update

aivan6842:
The reproduction was done on Ubuntu 18.04, 32gb RAM and 1Tb HDD. The process went smoothly, only notable difference was in the indexing time which took a little longer than README suggested. I suspect this is because of my HDD drive.


lintool (changes requested):
When I visit https://github.com/castorini/pyserini/commit/cbf933b15d6add94269962a138401626d1c5e923 I get the issue:

"This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

You should use the commit id from the `master` branch you did the repo?

lintool (approved):

PR #1248, opened 2022-08-11
Add unique_term to prebuilt msmarco-document-segment-ltr

yuki617:
All other stats are verified, just unique_term is missing.


lintool (approved):

lintool:
Yea... it appears you forgot to build index with `--optimize`... but it's okay

yuki617:
> Yea... it appears you forgot to build index with `--optimize`... but it's okay

I see. That's why the unique_term is -1...

PR #1258, opened 2022-08-30
Update MANIFEST.in to include requirements.txt to fix broken sdist

shaileshj2803:


lintool:
Here's the context of why we need this: https://packaging.python.org/en/latest/guides/using-manifest-in/
And related: https://stackoverflow.com/questions/38533669/include-requirements-txt-file-in-python-wheel

lintool (approved):

PR #1259, opened 2022-09-05
Update Installation.md

yogeswarl:
Hi, 
The documentation requires a slight change to the installation.md file. The sparse retrieval still uses msmarco-passage. 
I also encountered a runtime error for which I have appended the solution to the troubleshooting tip section.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -69,7 +69,7 @@ To confirm that bag-of-words retrieval is working correctly, you can run the BM2
 ```bash
 $ python -m pyserini.search \
     --topics msmarco-passage-dev-subset \
-    --index msmarco-passage \
+    --index msmarco-v1-passage\
     --output run.msmarco-passage.txt \
     --output-format msmarco \
     --bm25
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -175,6 +175,7 @@ On a Linux system, the correct path might look something like `/usr/lib/jvm/java
 Unfortunately, we are unable to offer more concrete advice since the actual path depends on your OS, which JDK you're using, and a host of other factors.
 + Windows uses GBK character encoding by default, which makes resource file reading in Anserini inconsistent with that in Linux and macOS.
 To fix, manually set environment variable `set _JAVA_OPTIONS=-Dfile.encoding=UTF-8` to use `UTF-8` encoding.
++ When installing with Windows, there are times you will encounter the error. `RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd`. The solution to this is to check the version of your numpy. At the time of this writing, the latest numpy version is 1.23.2 which is incompatible with the API. We consider downgrading to `1.21.1`.This way the other dependant libraries also are compatible with the API version.
```
lintool, line 178: Can you offer a bit more details on what's happening here? Is this a Windows only bug?

> numpy version is 1.23.2 which is incompatible with the API.

Which API?

lintool (commented):
Hi @yogeswarl thanks for your contribution and apologies for the delay in getting back to you.  A few clarification questions for you.

docs/installation.md
yogeswarl: This is because of the faiss-cpu library, I believe. I have noticed this in windows only in the latest 0.17.0 build. I have a personal Linux machine in which it worked fine.  I also checked it with macOS Monterrey.

lintool (approved):

PR #1268, opened 2022-09-17
Dr doc format

crystina-z:
Allow the document format of dense encoder to be not only 
```
{'id': ..., 'content', ...}
```
but also any key-value pair, as long as the keys matches the given `--fields`, for example:
```
{'docid': ..., 'title': ..., 'text': ...}
```

`python -m unittest` passed. scores looks reasonable on MIRACL arabic test


lintool (approved):

crystina-z:
@lintool sorry could you take another quick look? just added test case

lintool (approved):

lintool:
lgtm!

PR #1270, opened 2022-09-21
add dimension arg to dense encoding

crystina-z:
this is for models like miniLM and xlrm-large, where the hidden dimension isn't 768


MXueguang (approved):
lgtm

PR #1280, opened 2022-09-28
update reproduction log

Jasonwu-0803:
URA screening


lintool:
Can you confirm that everything works as expected? Briefly describe your computing environment.

Jasonwu-0803:
Yes, every thing works as expected. 
I ran it on Ubuntu with 8G ram.

lintool (approved):

PR #1281, opened 2022-09-28
Mrtydi lucene9

crystina-z:


lintool (approved):

lintool:
hi @crystina-z either do it in this PR or another PR is fine -

but please update: https://github.com/castorini/pyserini/blob/master/pyserini/resources/mrtydi.yaml

Issue #1278, opened 2022-09-27
Update Installation.md re: Windows issues


lintool:
Perhaps we should organize and create a separate section for windows installation tips?

PR #1284, opened 2022-10-03
Update installation.md

lintool:
In particular, we don't need to pin the version of PyTorch anymore - I've verified in a recent `conda` fresh install.


crystina-z (approved):

lintool:
Closes #1278

PR #1296, opened 2022-10-12
Miracl index (bm25)

crystina-z:


crystina-z:
side update: the topics and qrels for train and dev set are hosted [here](https://huggingface.co/datasets/miracl/miracl) (private mode). will continue fill in the dataset card

lintool (approved):

PR #1303, opened 2022-10-15
Update reproduction log

limelody:
Update Pyserini experiments reproduction log


lintool:
Were you able to overcome https://github.com/castorini/pyserini/discussions/1299 and reproduce the results? Did they match the documentation?

limelody:
Yes, I ran it remotely through a Google Colab notebook and the results matched (I have a Mac with better configs that I will have access to again in two-ish weeks)

lintool (approved):

PR #1312, opened 2022-10-20
MIRACL demo

ehsk:
An interactive web interface to test queries over MIRACL corpora. The demo is implemented via a lightweight web framework, `flask`.


lintool:
Hey @ToluClassics can you please take a look?

ToluClassics:
Hi @ehsk , I'm wondering if we can put  all the languages in a drop down so we can toggle between languages easily without having to restart the server

ToluClassics (approved):

PR #1313, opened 2022-10-21
MIRACL demo: language selection option added

ehsk:


pyserini/demo/miracl.py
```diff
@@ -70,20 +79,45 @@ def search():
             "miracl.html", search_results=search_results, query=query, lang=lang, retriever=retriever
         )
 
+    @app.route("/lang", methods=["GET"])
```
lintool, line 82: hi @ehsk - I just noticed - we use single quotes throughout the repo. Would you mind changing throughout?

lintool (changes requested):
double quotes to single quotes please.

pyserini/demo/miracl.py
ehsk: Fixed!

lintool (approved):

PR #1350, opened 2022-11-17
Fix all of int()-->isdigit() in  "_base.py"

ChuanMeng:
I still found other "int()" operations in the script _base.py. I've fixed all of them.

### DIFF
--- a/pyserini/search/_base.py
+++ b/pyserini/search/_base.py
@@ -439,11 +439,13 @@ def get_topics(collection_name):
 
     t = {}
     for topic in topics.keySet().toArray():
-        # Try and parse the keys into integers
-        if topic.isdigit()
+        
+        if topic.isdigit():
+            # parse the keys into integers
             topic_key = int(topic)
         else:
             topic_key = topic
+            
         t[topic_key] = {}
         for key in topics.get(topic).keySet().toArray():
             t[topic_key][key] = topics.get(topic).get(key)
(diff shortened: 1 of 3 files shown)


lintool:
Haha, would you be willing to write a test case?

lintool:
Here, for example: https://github.com/castorini/pyserini/blob/master/tests/test_load_topics.py

ChuanMeng:
> Here, for example: https://github.com/castorini/pyserini/blob/master/tests/test_load_topics.py

Hi Jimmy,
I already tested the modified script ("_base.py") on the server of my university. My modified version runs without any errors and the topic keys remain unchanged finally (see the following screenshot).
image

Thank you for contributing Pyserini, which is easy to use :-)
I also really hope, in the future, we can finally support the pre-built indexes on some conversational search datasets, such as Trec cast 19, 20, 21, 22, as well as OR-QuAC.

Cheers,
Chuan

lintool:
@ChuanMeng - what I meant was, can you write a test case to go with your code? ;)

ChuanMeng:
I think you can test using the case in the attachment and use the following function.
  
 def test_trec_topicreader(self):
        path = 'test.tsv'
        self.assertTrue(os.path.exists(path))
        topics = search.get_topics_with_reader('io.anserini.search.topicreader.TrecTopicReader', path)
        self.assertEqual(len(topics), 3)

[test.tsv.zip](https://github.com/castorini/pyserini/files/10035779/test.tsv.zip)

Cheers,
Chuan

lintool:
We prefer to adopt software engineering best practices in my group. Can you please add your test case to the PR? Then make sure `python -m unittest` passes?

ChuanMeng:
I'm sorry that I don't have much experience in software engineering. I have no idea of how to use "python -m unittest". I’d like to leave the standard verification to your group.

lintool (approved):

PR #1384, opened 2022-12-14
Miracl all pft

crystina-z:


crystina-z:
image

2cr looks like this now.

crystina-z:
I just realize on Enlgish w/ ft-all drops a lot, which isn't the case on mrtydi dev set. lmk double check before we merge this?

same for all non-mrtydi languages as well, tho it's more expected since the model might be more biased towards other seen languages.

lintool (approved):
lgtm. as long as the effectiveness differences aren't bugs, then I think we're good. okay to merge.

Issue #1375, opened 2022-12-07
Importing LuceneSearcher relies on FAISS and Torch


cakiki:
Currently, importing `LuceneSearcher` fails if faiss and torch aren't installed. (They aren't installed by design because they're platform-specific, see: https://github.com/castorini/pyserini#installation)

This is likely caused by the imports in the following init file: https://github.com/castorini/pyserini/blob/master/pyserini/search/__init__.py#L23-L26

A fix would need to modularize those imports.

If no one gets to it before me, I will attempt to send a PR to fix this.

MXueguang:
yeah, PR is welcomed!

lintool:
See #2009 - this has been resolved by #1997

PR #1387, opened 2022-12-16
Refactor Dependencies 

ToluClassics:
Initial PR Based on https://github.com/castorini/pyserini/issues/1375

Modularize imports so that LuceneSearcher does not rely on Faiss, torch, and transformers

### DIFF
--- a/pyserini/search/lucene/__init__.py
+++ b/pyserini/search/lucene/__init__.py
@@ -14,10 +14,11 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


lintool:
hi @ToluClassics I'm not familiar enough with Python dependencies to meaningfully comment... 

@cakiki can you comment?
@rodrigonogueira4 mentioned this issue before, I think.

I think we want to keep `requirements.txt` since feeding it into `pip` is a common pattern?

Issue #1290, opened 2022-10-06
Extend Trec to DPR Run Converter to use Custom Topics ?


ToluClassics:
The file [convert_trec_run_to_dpr_retrieval_run.py](https://github.com/castorini/pyserini/blob/master/pyserini/eval/convert_trec_run_to_dpr_retrieval_run.py) only allows the conversion of topics currently checked into anserini. 

I guess we can open this up to use custom query files also?

https://github.com/castorini/pyserini/blob/2673031f6b202941fe0f9953c9b876e6d4f1e653/pyserini/eval/convert_trec_run_to_dpr_retrieval_run.py#L26-L37

I can see it's possible using this function:

https://github.com/castorini/pyserini/blob/2673031f6b202941fe0f9953c9b876e6d4f1e653/pyserini/search/_base.py#L391-L407

ToluClassics:
@MXueguang

MXueguang:
yeah, that's a good idea. I guess we can first check if the "name" is a path, and if it is, load it using `get_topics_with_reader`, otherwise, assume it is self contain topic file

PR #1389, opened 2022-12-17
Extend Trec to DPR Run Converter to use Custom Topics 

ToluClassics:
PR based on https://github.com/castorini/pyserini/issues/1290

### DIFF
--- a/integrations/dense/test_dpr.py
+++ b/integrations/dense/test_dpr.py
@@ -17,6 +17,7 @@
 """Integration tests for DPR model using pre-encoded queries."""
 
 import os
+import json
 import socket
 import unittest
 
@@ -57,6 +58,7 @@ class TestSearchIntegration(unittest.TestCase):
         self.assertEqual(status1, 0)
         self.assertEqual(status2, 0)
         self.assertAlmostEqual(score, 0.7947, places=4)
+
 
     def test_dpr_nq_test_bf_bm25_hybrid_otf(self):
         output_file = 'test_run.dpr.nq-test.multi.bf.otf.bm25.trec'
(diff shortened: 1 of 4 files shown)


lintool:
Instead of overloading `--topics`, what if we implement _two_ options, `--topics-file` and `--topics-reader` to make more general?

@MXueguang what do you think?

Also, simple test case please.

ToluClassics:
I'd also like to extend this to load custom DPR format kind of QA topics with `DprNqTopicReader` but I'm not sure how to restructure the if-else conditions 🤔. any ideas?

https://github.com/castorini/pyserini/blob/3d8c4731507c3f30e6a88243e26443c681a5c826/pyserini/query_iterator.py#L76-L105

MXueguang:
> Instead of overloading `--topics`, what if we implement _two_ options, `--topics-file` and `--topics-reader` to make more general?

yeah, `--topics-file` and `--topics-reader` looks good.

integrations/dense/test_dpr.py
```diff
@@ -301,9 +303,38 @@ def test_dpr_squad_test_encoded_queries(self):
...
+        self.assertEqual(topic_data["0"]["answers"], ['text'])
+        self.assertEqual(topic_data["0"]["question"], "what is in document three")
+        self.assertEqual(topic_data["1"]["answers"], ['contents'])
+        self.assertEqual(topic_data["1"]["question"], "what is document two")
+
+
     def tearDown(self):
         clean_files(self.temp_files)
 
-
```
lintool, line 307: I think PEP8 wants two empty lines?

lintool (commented):
Minor comment from me - but this is @MXueguang 's code I think? so I'll let him take a look also?

MXueguang (approved):
changes looks good to me

PR #1400, opened 2022-12-27
Indexing and Search with GTR Models in Pytorch

ToluClassics:


ToluClassics:
PR based off https://github.com/lintool/fourdrinier/issues/636

pyserini/search/faiss/_searcher.py
```diff
@@ -168,10 +168,11 @@ def _mean_pooling(model_output, attention_mask):
         sum_mask = torch.clamp(input_mask_expanded.sum(1), min=1e-9)
         return sum_embeddings / sum_mask
 
-    def encode(self, query: str):
+    def encode(self, query: str, max_length: int = 64):
```
lintool, line 171: is this change going to affect all our other models?

ToluClassics: No , it won't

cramraj8:
Any reason why ? Does the GTR integration works with Pyserini ?

cramraj8:
Any reason why GTR models were not integrated with Pyserini ? Are the current changes run without any errors and correctly reproduce the reported results?

ToluClassics:
I'm not sure why we didn't end up merging this but I think the branch works, maybe you can pull that branch and try it out?

Alternatively is there a particular use case you need this for?

Issue #1402, opened 2022-12-29
Formatting issues with ODQA repro matrix


lintool:
(Minor) display issues:

<img width="1383" alt="Screen Shot 2022-12-29 at 6 42 27 PM" src="https://user-images.githubusercontent.com/313837/210021622-31e8c9e9-60f9-4740-a312-90670be2faab.png">

In above screenshot, top row rule doesn't span all the way over to the right...

---

<img width="1430" alt="Screen Shot 2022-12-29 at 6 42 38 PM" src="https://user-images.githubusercontent.com/313837/210021656-5fe00cd7-6e87-4e2d-99f7-b1971d3baa92.png">

When you expand a row, see how the column widths change? They shouldn't... the auto column resizing causes visual inconsistencies and is a bit jarring.

lintool:
Also, I'm not sure we need top-1000 for repro... it takes a long time to run, esp the part that involves fetching Wikipedia passages. Perhaps repro top-100 is sufficient? Users can easily adjust to 1000 if they wish?

From the repro perspective, it is unlikely that a test passes at top-100 but somehow fails at top-1000?

lintool:
Closed by #1403

PR #1403, opened 2022-12-30
ODQA frontend + repro topk issues resolved 

ArthurChen189:
issue #1402 resolved


scripts/repro_matrix/run_all_odqa.py
```diff
@@ -44,6 +44,8 @@ def print_results(metric, topics):
                         default=False, help='Skip running trec_eval.')
     parser.add_argument('--topics', choices=['triviaqa', 'naturalquestion'],
                         help='Topics to be run [triviaqa, naturalquestion]', required=True)
+    parser.add_argument('--full_topk', action='store_true',
```
lintool, line 47: Change `--full_topk` to `--full-topk`?

More consistent with rest of code base and more readable.

lintool (changes requested):
You're changing the eval... but where are the changing the part where you only retrieval top-k  from Lucene?

scripts/repro_matrix/run_all_odqa.py
ArthurChen189: Interesting. I thought that was a parsing issue. The dash turns out to be parsed as an underscore.

ArthurChen189:
Oh, I thought we only want topk100 on the eval end since evaluating topk1000 requires re-retrieving.

lintool:
> Oh, I thought we only want topk100 on the eval end since evaluating topk1000 requires re-retrieving.

Yes, correct - but that means we only need to fetch top 100 from Lucene also, right? That will make the Lucene part faster also.

ArthurChen189:
Fixed
Notes:
- web page uses default hits = 100

lintool (approved):

PR #1420, opened 2023-01-25
miracl yo, de pft-then-ft-all; hybrid 2cr for all

crystina-z:
current 2cr looks like this:
image


lintool (approved):
This is good for now. I will run again on my end. Diff is too hard to process in this form.

Thanks!

PR #1465, opened 2023-03-16
Bug in AutoDocumentEncoder where `l2_norm` argument not being applied.

sieu-n:


sieu-n:
```
from sklearn.preprocessing import normalize
import numpy as np
v = np.array([[0.5, 0.5]])
normalize(v), v

Out: 
(array([[0.70710678, 0.70710678]]), array([[0.5, 0.5]]))
```

`sklearn.preprocessing.normalize` is not inplace operation.

sieu-n:
@lintool @MXueguang

MXueguang (approved):

PR #1467, opened 2023-03-19
Add instructions of doc() method when searching their own documents

flora0110:
The code for fetching a document in the readme did not work properly when searching their own documents, as it returned NoneType instead of the expected document. So I Added instructions on how to use the doc() method to fetch document under "How do I index and search my own documents?" subsection of the readme. These instructions specify the format of the document ID, which should include the 'doc' prefix. 

I apologize for the extra files that were created in my previous PR.I'm not sure what caused the creation of multiple extra files in the contributors/ directory during my previous PR submission., I hope it wouldn't happen again, but if it happend, please tell me how to fixed it.

Thank you for considering my contribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ A sparse (Lucene) index can be configured to include the raw document text, in w
 from pyserini.search.lucene import LuceneSearcher
 
 searcher = LuceneSearcher.from_prebuilt_index('msmarco-v1-passage')
-doc = searcher.doc('doc7157715')
+doc = searcher.doc('7157715')
 ```
 
 <details>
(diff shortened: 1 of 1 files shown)


lintool:
Hi @flora0110 - thanks for the PR, but the description you added is not accurate.

If you look at the few lines above your addition, `doc2` (as a string) is the name of the document, which is the input arg to `doc()`.

Note that in this code snippet:

```
for i in range(len(hits)):
    print(f'{i+1:2} {hits[i].docid:4} {hits[i].score:.5f}')
```

We're printing `hits[i].docid` - this docid (as a string) is the arg you pass back to `doc()` to fetch the doc.

PR #1475, opened 2023-04-04
Add Aggretriever into pyserini

jacklin64:
In this PR, we add Aggretriever query and document encoders into pyserini code base. Using the encoders, we built the msmarco v1 passage index and dev query embeddings from Aggretriever-DistilBERT and Aggretriever-coCondenser for easily reproducing. The huggingface models, pre-built indexes and queriers are now in orca.


lintool (approved):
I think this is fine as an initial PR to merge, but we additionally need:
(1) on-the-fly query inference
(2) checking in actual pre-encoded queries
(3) uploading pre-built indexes

alexlimh (approved):
LGTM

agimbrac:
> I think this is fine as an initial PR to merge, but we additionally need:
> 
> (1) on-the-fly query inference
> 
> (2) checking in actual pre-encoded queries
> 
> (3) uploading pre-built indexes

[]()

PR #1490, opened 2023-04-08
Reorganize info for pre-built indexes into different files

lintool:
Also removed deprecated indexes.

What do you think @MXueguang @crystina-z ? Worth doing?


lintool:
Closed in favor of #1493

PR #1498, opened 2023-04-13
Rename Faiss indexes to be consistent with current naming scheme

lintool:


jacklin64:
It seems some dense models still use the same naming rule like adore and aggretriever. Do we have to change them also? or just change all the main model in pyserini this time?

lintool:
> It seems some dense models still use the same naming rule like adore and aggretriever. Do we have to change them also? or just change all the main model in pyserini this time?

Yea, I'll circle back and do another round of cleanup.

jacklin64 (approved):
LGTM

PR #1529, opened 2023-05-21
Add ONNX otf options into msmarco-v1-passage 2cr

ArthurChen189:
Add ONNX otf entries:
- UniCoil (no-exp)
- UniCoil (w/doc2query-T5)
- SPLADE-pp-ed
- SPLADE-pp-sd


lintool:
Can we change: "uniCOIL (w/ doc2query-T5): on-the-fly query inference"
to "uniCOIL (w/ doc2query-T5): on-the-fly query inference with PyTorch" (for all existing otf cases?)
and then we can say "... on-the-fly query inference with ONNX"

Also, let's use the full canonical name of the models:
+ SPLADE++ CoCondenser-EnsembleDistil
+ SPLADE++ CoCondenser-SelfDistil

lintool (changes requested):
See comments.

pyserini/2cr/msmarco-v1-passage.yaml
```diff
@@ -495,10 +539,56 @@ conditions:
...
+  - name: splade-pp-ed-otf-onnx
+    display: "SPLADE++ CoCondenser-EnsembleDistil with ONNX Runtime"
+    display-html: "SPLADE++ CoCondenser-EnsembleDistil with ONNX Runtime"
+    display-row: "[2]"
+    command: python -m pyserini.search.lucene --threads 16 --batch-size 128 --index msmarco-v1-passage-splade-pp-ed --topics $topics --onnx-encoder SpladePlusPlusEnsembleDistil --output $output --hits 1000 --impact
+    topics:
+      - topic_key: msmarco-passage-dev-subset
+        eval_key: msmarco-passage-dev-subset
+        scores:
+          - MRR@10: 0.3830
```
lintool, line 551: hey @ArthurChen189 why don't the numbers match those from Anserini? e.g., https://github.com/castorini/anserini/blob/master/docs/regressions-msmarco-passage-splade-pp-ed-onnx.md

I spot checked a few, and they seemed to be the pre-encoded queries?

lintool:
hey @ArthurChen189 as you're working on this, can we regroup the conditions differently?

Currently they are:
+ method A (pre-encoded)
+ method B (pre-encoded)
+ ...
+ method A (otf)
+ method B (otf)

Would it make more sense to do?

+ method A (pre-encoded)
+ method A (otf PyTorch)
+ method A (otf ONNX)
+ ...
+ method B (pre-encoded)
+ method B (otf PyTorch)
+ method B (otf ONNX)

thoughts?

pyserini/2cr/msmarco-v1-passage.yaml
ArthurChen189: I've re-run experiments and put checkpoint Pyserini did call Anserini's encoder functions and with the ONNX model. Results are still the same as the ones in this pr.

lintool (approved):

PR #1531, opened 2023-05-24
Fix bug with ImpactSearcher initialization

lintool:
hey @alexlimh as part of #1445 - you accidentally removed the branch when we specify pre-encoded queries, i.e., `args.encoder` isn't set. Adding back in.


alexlimh (approved):
lgtm

PR #1532, opened 2023-05-25
Allow users to specify whether to use inner product or L2 for buildin…

jonhilgart22:
…g faiss vectors

### DIFF
--- a/pyserini/index/faiss.py
+++ b/pyserini/index/faiss.py
@@ -35,6 +35,7 @@ if __name__ == '__main__':
     parser.add_argument('--pq-m', type=int, default=192, required=False)
     parser.add_argument('--pq-nbits', type=int, default=8, required=False)
     parser.add_argument('--threads', type=int, default=12, required=False)
+    parser.add_argument('--flatmetric', type=string, default="inner", required=False)
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


pyserini/index/faiss.py
```diff
@@ -35,6 +35,7 @@
     parser.add_argument('--pq-m', type=int, default=192, required=False)
     parser.add_argument('--pq-nbits', type=int, default=8, required=False)
     parser.add_argument('--threads', type=int, default=12, required=False)
+    parser.add_argument('--flatmetric', type=string, default="inner", required=False)
```
lintool, line 38: why `flatmetric` and not just `metric`?

lintool (commented):
Hi, thanks for your contribution!

pyserini/index/faiss.py
jonhilgart22: both are specific for the flat index, but, likely, you should pick a metric and apply it to either the flat index or one of the other index builds.

lintool: So we should rename to `--metric` then? @MXueguang @crystina-z thoughts?

lintool:
@MXueguang what do we do with this PR?

MXueguang (approved):

PR #1535, opened 2023-05-27
Minor tweak for Mr.TyDi 2CR

lintool:
cc @jasper-xian

### DIFF
--- a/docs/2cr/mrtydi.html
+++ b/docs/2cr/mrtydi.html
@@ -3083,4 +3083,3 @@ clipboard.on('error', function(e) {
 
 </body>
 </html>
-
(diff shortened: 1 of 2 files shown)


jasper-xian:
Ah yes, lgtm apologies about the typo

lintool:
@jasper-xian btw, I don't think the code will work when we publish as a PyPI package... I'll need to circle back and refactor: https://github.com/castorini/pyserini/blob/master/pyserini/2cr/msmarco.py#L415

```
yaml_file = pkg_resources.resource_filename(__name__, f'{args.collection}.yaml')
```

crystina-z (approved):

PR #1539, opened 2023-05-31
Update miracl/mrtydi docs with 2CR instructions

jasper-xian:


lintool (changes requested):
Hi @jasper-xian I think you need to modify the template and then run the command to regenerate the report?

jasper-xian:
Ah that makes more sense, changes made

lintool (approved):

PR #1542, opened 2023-06-04
Update working-with-acl-anthology.md

billcui57:
## Changes

Added the following line: "Note that you should be using the already created `pyserini_acl` conda environment rather than making a new one that was instructed [here](https://github.com/castorini/pyserini/blob/master/docs/installation.md#:~:text=conda%20create%20%2Dn%20pyserini%20python%3D3.8). "

Added a section for Reproduction Log

### DIFF
--- a/docs/working-with-acl-anthology.md
+++ b/docs/working-with-acl-anthology.md
@@ -43,7 +43,9 @@ Generated ACL files can now be found in `acl-anthology/build/data/`
 
 ## Indexing Data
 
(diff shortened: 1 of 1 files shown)


lintool:
We have this convention of one sentence on each line, because it makes the diffs easier to read. Can you please modify throughout?

lintool:
Also, add a section for Reproduction Log, just like all the other pages?

lintool (approved):

PR #1552, opened 2023-06-13
Added acl lucene demo

billcui57:
### What was done

- Added a web server that demonstrates retrieval from the `ACL anthology` dataset
image

### How it was done

- Followed through steps to generate `ACL` index from [working-with-acl-anthology.md](https://github.com/billcui57/pyserini/blob/master/docs/working-with-acl-anthology.md)
- Copy pasted the generated index into `Pyserini` `/indexes` directory
- Created web server that points `Lucene` at that index

### Notes

- Only sparse search was used. I.e this demo is only backed by `LuceneSearcher`
- Multiple languages not supported

### DIFF
--- a/pyserini/demo/acl.py
+++ b/pyserini/demo/acl.py
@@ -1,0 +1,124 @@
+#
+# Pyserini: Reproducible IR research with sparse and dense representations
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


lintool:
hey @ehsk @mobehbooei can you take a look and try it out?

ehsk:
I followed the instructions and was able to run the demo.
The content column is shown as HTML, and it'd be great if it is rendered accordingly.

billcui57:
> The content column is shown as HTML, and it'd be great if it is rendered accordingly. 
Could you please elaborate on this? What should the content column render instead? @ehsk

ehsk:
The content of the returned documents is sometimes an HTML code. I was wondering if they can be rendered instead of being shown as raw text, but just realized that not all documents contain HTML code, so I think no need to do anything.

image

lintool (approved):

PR #1555, opened 2023-06-16
Update pooling options

jonhilgart22:
For some models, such as e5 ([see here ](https://huggingface.co/intfloat/e5-large-v2)), we need the option to specify mean pooling instead of just cls pooling.


lintool:
@crystina-z or @MXueguang can either of you take a look at this?

jonhilgart22:
@crystina-z or @MXueguang Let me know what you think. This is helpful for us running the `https://huggingface.co/intfloat/e5-large-v2` models

MXueguang:
yes, this change looks good to me.

MXueguang:
I think e5 also needs adding prefix "passage: ", easy to adding to this PR also?

jonhilgart22:
> I think e5 also needs adding prefix "passage: ", easy to adding to this PR also?

I don't think we should tie this to `e5`; I was only giving that as an example. However, happy to add if you'd like!

MXueguang:
yeah, lets deal with that in other PR if needed, current change should be good to merge

MXueguang (approved):

PR #1556, opened 2023-06-19
Added demo to talk to ACL with GPT

billcui57:
# What was done
Created a demo that connects a flask server running Lucene sparse index searcher with GPT. 

User queries are first sent to GPT to rewrite into "absolute" questions, meaning questions that can stand on their own without additional context

Absolute questions are then proxied to the Lucene searcher, the results are used by GPT to summarize and produce meaningful answers back to the user

## Demo

https://github.com/castorini/pyserini/assets/30924631/4e0ca725-41e8-42dd-8df9-d806b8412ba6


lintool:
Considering that this dates from 2023, we were ahead of our time :)

But MCP supersedes this...

PR #1558, opened 2023-06-19
Add SLIM 2CR on MSMARCO-v2-passage

alexlimh:


lintool:
hey @alexlimh I don't think you built index with `--optimize`. Can you please do so? (You can tell by the number of file segments in the index directory).

Also, the README doesn't actually exist...

lintool (changes requested):
Noted in comments.

Issue #1563, opened 2023-07-04
Pre-built indexes throws value error.


mrdrozdov:
When calling:

```
LuceneSearcher.list_prebuilt_indexes()
```

I get this ValueError:

```
File ~/miniconda3/envs/torch-latest/lib/python3.9/site-packages/pyserini/search/lucene/_searcher.py:93, in LuceneSearcher.list_prebuilt_indexes()
     90 @staticmethod
     91 def list_prebuilt_indexes():
     92     """Display information about available prebuilt indexes."""
---> 93     get_sparse_indexes_info()

File ~/miniconda3/envs/torch-latest/lib/python3.9/site-packages/pyserini/util.py:193, in get_sparse_indexes_info()
    190 for index in df.keys():
    191     df[index]['downloaded'] = check_downloaded(index)
--> 193 with pd.option_context('display.max_rows', None, 'display.max_columns',
    194                        None, 'display.max_colwidth', -1, 'display.colheader_justify', 'left'):
    195     print(df)

File ~/miniconda3/envs/torch-latest/lib/python3.9/site-packages/pandas/_config/config.py:444, in option_context.__enter__(self)
    441 self.undo = [(pat, _get_option(pat, silent=True)) for pat, val in self.ops]
    443 for pat, val in self.ops:
--> 444     _set_option(pat, val, silent=True)

File ~/miniconda3/envs/torch-latest/lib/python3.9/site-packages/pandas/_config/config.py:160, in _set_option(*args, **kwargs)
    158 o = _get_registered_option(key)
    159 if o and o.validator:
--> 160     o.validator(v)
    162 # walk the nested dict
    163 root, k = _get_root(key)

File ~/miniconda3/envs/torch-latest/lib/python3.9/site-packages/pandas/_config/config.py:882, in is_nonnegative_int(value)
    879         return
    881 msg = "Value must be a nonnegative integer or None"
--> 882 raise ValueError(msg)

ValueError: Value must be a nonnegative integer or None
```

Was able to fix locally by changing -1 in pd.option_context to None.

lintool:
@mrdrozdov thanks for your report. However, I wasn't able to reproduce the bug:

```python
from pyserini.search.lucene import LuceneSearcher
LuceneSearcher.list_prebuilt_indexes()

from pyserini.index.lucene import IndexReader
IndexReader.list_prebuilt_indexes()
```

These all work fine for me.

```
% pip list | grep pandas
pandas             1.5.2
```

[...]

PR #1567, opened 2023-07-07
Add support for pandas v2.0.0+.

mrdrozdov:
After pandas 2.0.0, using -1 for display.max_colwidth was removed in favor of None. Using -1 throws a ValueError. 

Reference from pandas: https://github.com/pandas-dev/pandas/pull/49295/files

See additional discussion here: https://github.com/castorini/pyserini/issues/1563


lintool (approved):
I confirm that this fix is backwards compatible, i.e., works with `panda` 1.5.2. Merging.

PR #1572, opened 2023-07-20
custom docid field added to pyserini.encode

ehsk:
- `--docid-field` argument added for custom docid field names.
- `_id` added to default names as well.

### DIFF
--- a/pyserini/encode/__main__.py
+++ b/pyserini/encode/__main__.py
@@ -130,7 +130,7 @@ if __name__ == '__main__':
         embedding_writer = FaissRepresentationWriter(args.output.embeddings, dimension=args.encoder.dimension)
     else:
         embedding_writer = JsonlRepresentationWriter(args.output.embeddings)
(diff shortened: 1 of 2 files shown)


pyserini/encode/__main__.py
```diff
@@ -91,6 +91,9 @@ def parse_args(parser, commands):
                               required=True)
     input_parser.add_argument('--fields', help='fields that contents in jsonl has (in order)',
                               nargs='+', default=['text'], required=False)
+    input_parser.add_argument('--docid-field',
+                              help='name of id field name. `default` would be either of `id`, `_id`, `docid`',
+                              default=None, required=False)
```
thakur-nandan, line 96: @ehsk, Overall looks good to me. Maybe here you can add by default it is `id`. This way someone is not confused in the future with the `--docid-field`, i.e. why does it work with `None`.

Also in the helper we can mention that the `doc_id` can be any key (few common options are `id`, `_id` or `docid` .

thakur-nandan (approved):
I added my review to one line. 

Overall looks good to me.

PR #1574, opened 2023-07-21
Refactor + augment onboarding docs for MS MARCO passage

lintool:
Big addition is the interactive retrieval section.


ronakice (approved):
LGTM!

PR #1577, opened 2023-07-23
Update experiments-msmarco-passage.md with qrel conversion

sahel-sh:
This cl:
- Adds qrel conversion commands 
- Adds `sahel-sh` to repo log for ` experiments-msmarco-passage.md`

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -45,10 +45,7 @@ wget https://msmarco.blob.core.windows.net/msmarcoranking/collectionandqueries.t
 tar xvfz collections/msmarco-passage/collectionandqueries.tar.gz -C collections/msmarco-passage
 ```
 
-To confirm, `collectionandqueries.tar.gz` should have MD5 checksum of `31644046b18952c1386cd4564ba2ae69`:
-```bash
-md5sum collections/msmarco-passage/collectionandqueries.tar.gz
-```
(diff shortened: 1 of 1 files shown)


docs/experiments-msmarco-passage.md
```diff
@@ -45,7 +45,10 @@ wget https://msmarco.blob.core.windows.net/msmarcoranking/collectionandqueries.t
 tar xvfz collections/msmarco-passage/collectionandqueries.tar.gz -C collections/msmarco-passage
 ```
 
-To confirm, `collectionandqueries.tar.gz` should have MD5 checksum of `31644046b18952c1386cd4564ba2ae69`.
+To confirm, `collectionandqueries.tar.gz` should have MD5 checksum of `31644046b18952c1386cd4564ba2ae69`:
```
lintool, line 48: Hrm, I'm thinking that perhaps we should remove the checksum command here?

(1) our target audience is CS ugrads, and I think a CS ugrad should already know how to verify checksums?
(2) furthermore, on Mac the command is `md5`, which is different... and may end up causing more confusion?

I missed it, but maybe be consistent and remove here?

https://github.com/castorini/anserini/blob/master/docs/start-here.md

What do you think?

sahel-sh: removed for Pyserini in this PR, will create a PR for Anserini to also remove it there

lintool (approved):
lgtm

PR #1584, opened 2023-08-01
Add script to convert AToMiC raw numpy embeddings to pyserini format

jasper-xian:
Converts raw numpy embeddings (`embeddings.npy` and `ids.txt`) files to pyserini-format topics (`topic.json`) and encoded queries (`embedding.pkl`) to be able to use `pyserini.search.faiss` to run baselines on the AToMiC dataset


justram (approved):
LGTM

PR #1586, opened 2023-08-03
Fix typo in searchFields method name

AmrMKayid:
Fix typo when searching with fields. Rename method `searchFields` to correct one `search_fields`

`AttributeError: 'io.anserini.search.SimpleSearcher' object has no attribute 'searchFields'`


lintool:
Hi @AmrMKayid - thanks for the PR, but this has already been fixed!

PR #1592, opened 2023-08-08
Add ciral bm25 baselines and experiment doc

Mofetoluwa:
This cl includes:
- CIRAL's bm25 baselines for all 4 languages: Hausa, Somali, Swahili, Yoruba.

Comments/Issue:
- Just `recall@1000` is evaluated currently, as `ndcg@20` and `recall@100` for Swahili and Hausa are 0.

### DIFF
--- a/docs/experiments-ciral-v1.0.md
+++ b/docs/experiments-ciral-v1.0.md
@@ -1,14 +1,14 @@
 # Pyserini Regressions: BM25 Baselines for CIRAL Version 1.0
 
-This page documents BM25 regression experiments for CIRAL (v1.0)
+This page documents BM25 regression experiments for CIRAL (v1.0).
 
 ## Corpus Download
 
 ### 1. Manual Download
 
-The CIRAL Corpus can be downloaded from the [Hugging Face Repo](https://huggingface.co/datasets/CIRAL/ciral-corpus)
+The CIRAL Corpus can be downloaded from the [Hugging Face Repo](https://huggingface.co/datasets/CIRAL/ciral-corpus).
 
-We can now index the documents for each language as a `MrTyDiCollection`  using Anserini bindings from Pyserini
+We can now index the documents for each language as a `MrTyDiCollection`  using Anserini bindings from Pyserini.
 
 ```bash
 python -m pyserini.index.lucene --collection MrTyDiCollection \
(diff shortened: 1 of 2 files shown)


lintool (commented):
hi @Mofetoluwa - the participant _either_ downloads the pre-built indexes or indexes from scratch, right? Doesn't need to do _both_?

Perhaps this can be reflected in the docs?

Mofetoluwa:
Hi Jimmy, yeah that's true, I was kinda following the format for the current Pyserini regressions documents.

I'll remove the prebuilt indexes downloads, or do I just leave a comment about it in the doc?

docs/experiments-ciral-v1.0.md
```diff
@@ -0,0 +1,139 @@
...
+python -m pyserini.eval.trec_eval -c -m recall.1000 tools/topics-and-qrels/qrels.ciral-v1.0-sw-train.tsv runs/run.ciral-v1.0-sw.bm25.topics.ciral-v1.0-sw.train.txt
+python -m pyserini.eval.trec_eval -c -m recall.1000 tools/topics-and-qrels/qrels.ciral-v1.0-yo-train.tsv runs/run.ciral-v1.0-yo.bm25.topics.ciral-v1.0-yo.train.txt
+```
+
+## Effectiveness
+
+### Yoruba
+
+With the above commands, you should be able to reproduce the following results:
+| **recall@1000**                                                                   | **BM2 (default)**|
```
lintool, line 109: Should be BM25.
Why not evaluate also with MRR or nDCG?

Mofetoluwa: Hi Jimmy,
Do you think we should hold on about this based on the results?

docs/experiments-ciral-v1.0.md
```diff
@@ -0,0 +1,139 @@
...
+cp -r ciral/*/*/* $PYSERINI_PATH/tools/topics-and-qrels/
+```
+
+
+
+## Retrieval: Dev Topics
+
+```bash
+python -m pyserini.search.lucene --index ciral-v1.0-ha \
+    --topics tools/topics-and-qrels/topics.ciral-v1.0-ha-train \
```
crystina-z, line 71: we hvn't add the topics to the tools/ yet?

crystina-z: also could we name it `dev` instead of `train`? that'd be more accurate

lintool: our just call it "sample"?

crystina-z (approved):

PR #1606, opened 2023-08-28
Reproduction of Pyserini Onboarding

Andrwyl:
Environment:

Ubuntu 22.04 on student environment.

Issues:

Found there were issues with the dense retrieval on 11.5 gb of RAM, so linux virtual machine on a windows laptop with 16gb RAM could not work. Good news is the entire onboarding to this point works on the student environment for any future Waterloo student that doesn't have more than 12 gb RAM on their local computer. When running dense retrieval, CPU times out by default so you actually have to use the command `ulimit -t unlimited` to get it going.


lintool:
Hi @Andrwyl - thanks for working on this. We're in the middle of a release cycle right now - I'm cutting a release soon, and then I'll circle back and merge these. Thanks!

lintool (approved):

Issue #1616, opened 2023-09-02
Resolve unittest issues


ArthurChen189:
inconsistent method names with `test_encoder.py`, due to refactoring the ONNX query encoder in Anserini

lintool:
Unit tests for me... did you update your branch?

ArthurChen189:
Hrmm weird. And yeah, my branch is up-to-date with the castorini upstream

ArthurChen189:
I ran the unittest with anserini 0.22.0's fatjar on both my local end and ORCA....

SumonKantiDey:
How much time does it take to complete the unit test? I am trying the unit test with ```anserini-0.22.1-SNAPSHOT-fatjar.jar``` but somehow it takes a significant amount of time.

lintool:
```
Ran 308 tests in 321.069s

OK
```

On my main dev machine:

<img width="270" alt="Screen Shot 2023-09-04 at 5 31 20 PM" src="https://github.com/castorini/pyserini/assets/313837/516ea0ba-badb-4c33-8160-0db131e23412">

lintool:
Closing issue since #1617 has been merged.

PR #1617, opened 2023-09-02
Fix unittest errors and failures

ArthurChen189:
Related to issue #1616


lintool:
@ArthurChen189 can you fix your conflicts and then see what happens?

lintool (approved):

PR #1619, opened 2023-09-05
Update _impact_searcher.py to use renamed java functions.

sahel-sh:
https://github.com/castorini/anserini/commit/e475cc4e885f5fbb223533d75a2da7abaa618c6c renames `document/documentByField` functions to `doc/doc_by_field`. This cl changes the function calls in _impact_searcher.py to match the new names to fix the following error:
````
File ".../lib/python3.10/site-packages/pyserini/search/lucene/_impact_searcher.py", line 248, in doc
    lucene_document = self.object.document(docid)
AttributeError: 'io.anserini.search.SimpleImpactSearcher' object has no attribute 'document'
```

### DIFF
--- a/pyserini/search/lucene/_impact_searcher.py
+++ b/pyserini/search/lucene/_impact_searcher.py
@@ -243,7 +243,7 @@ class LuceneImpactSearcher:
         Document
             :class:`Document` corresponding to the ``docid``.
         """
-        lucene_document = self.object.document(docid)
+        lucene_document = self.object.doc(docid)
         if lucene_document is None:
             return None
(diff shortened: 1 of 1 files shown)


lintool (commented):
Can you add a quick test case so these name mismatch issues don't creep back in?

sahel-sh:
> Can you add a quick test case so these name mismatch issues don't creep back in?

Created an issue for the test coverage per our chat: https://github.com/castorini/pyserini/issues/1621

lintool (approved):
We'll circle back to add test case, but this is good for now!

Issue #1629, opened 2023-09-11
Integration failures on linux.cs


lintool:
At d8dc5b3a1f32fd5d0cebeb711ba148ea967fadbe setting new env on `linux.cs`:

```
deactivate

cd ~/virtualenv/
python -m venv pyserini-dev
source ~/virtualenv/pyserini-dev/bin/activate

export JAVA_HOME=/usr/lib/jvm/java-1.11.0-openjdk-amd64

pip install torch faiss-cpu
pip install -e .
```

Running integrations:

```
python -m unittest discover -s integrations/sparse
python -m unittest discover -s integrations/dense
python -m unittest discover -s integrations/clprf
python -m unittest discover -s integrations/papers
```

`dense` and `papers` pass fine; getting failures for `sparse` and `clprf`.

Failures for `sparse`:

For `sparse`, these were the failures:

```
======================================================================
ERROR: test_reranking (test_lucenesearcher_check_ltr_msmarco_document.TestLtrMsmarcoDocument)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/u4/jimmylin/pyserini/integrations/sparse/test_lucenesearcher_check_ltr_msmarco_document.py", line 50, in test_reranking
    result = subprocess.check_output(f'python tools/scripts/msmarco/msmarco_doc_eval.py --judgments tools/topics-and-qrels/qrels.msmarco-doc.dev.txt --run ltr_test/{outp}', shell=True).decode(sys.stdout.encoding)
  File "/usr/lib/python3.10/subprocess.py", line 421, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'python tools/scripts/msmarco/msmarco_doc_eval.py --judgments tools/topics-and-qrels/qrels.msmarco-doc.dev.txt --run ltr_test/run.ltr.msmarco-pass-doc.test.trec' returned non-zero exit status 1.

======================================================================
ERROR: test_reranking (test_lucenesearcher_check_ltr_msmarco_passage.TestLtrMsmarcoPassage)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/u4/jimmylin/pyserini/integrations/sparse/test_lucenesearcher_check_ltr_msmarco_passage.py", line 50, in test_reranking
[...]

PR #1638, opened 2023-09-15
Fix integration tests as described in #1629

lintool:
Ref #1629


lintool:
hang on... I'm going to layer more factoring on top of this PR.

lintool:
Confirmed that the following passes on `linux.cs` and `tuna`:

```
nohup python -m unittest discover -s integrations/sparse >& logs/log.sparse &
nohup python -m unittest discover -s integrations/dense >& logs/log.dense &
nohup python -m unittest discover -s integrations/clprf >& logs/log.clprf &
nohup python -m unittest discover -s integrations/papers >& logs/log.papers &
```

PR #1640, opened 2023-09-17
Add flask app and template to demo bm25 search on atomic

dlrudwo1269:
Added a Flask app and html template to demo BM25 search on AToMiC. Uses pre-built indexes from https://github.com/castorini/pyserini/pull/1536.

Screenshots
Screenshot 2023-09-17 at 6 42 51 PM
Screenshot 2023-09-17 at 6 43 13 PM


lintool:
I'll let @justram handle it... but do I get to see images?

dlrudwo1269:
For the BM25 baselines we only used the caption (generated by concatenating the values under the fields reference, alt-text, and attribution) to represent the images.

I believe we would have to reindex the documents to include the image url if we did want to display them on the page. Currently the indexed docs only have "id" and "contents" information.

lintool:
Aren't those images available on the web already?

dlrudwo1269:
That's true, I think I can use the huggingface api to get the images from the ids. I will make the change.

lintool:
Aren't these images already on Wikimedia site? Just link to their original URLs?

dlrudwo1269:
The thing is we throw away the url info before we index the documents (we only keep the id and the caption contents), so the urls are not available directly.

Instead, I thought we could use the ids to look up the url info in the dataset hosted on HuggingFace. However, looking through the documentation, I could not find an API to efficiently search only a few documents by id without preprocessing the dataset. Even if there were an efficient way, I believe the user would have to download the dataset beforehand, which is > 100GB. (I don't have extensive experience using HF datasets, so there may be a method I am unaware of; if there is a good method, I would be happy to know about it)

So if we do want to show the images I think the easiest way would be to reindex the docs to include the image url info?

lintool:
> So if we do want to show the images I think the easiest way would be to reindex the docs to include the image url info?

Yes, I think that's the right approach.

lintool:
Hi @dlrudwo1269 any updates here?

dlrudwo1269:
Hi, sorry I've been struggling from COVID these past few days and wasn't able to work on this :(

I have a preliminary change after reindexing the docs that looks like
Screenshot 2023-09-24 at 1 37 46 AM


I will talk with @justram if there are any additional changes we want to make and push the changes soon!

lintool:
@dlrudwo1269 Oh no, sorry to hear! Yes, please take care and coordinate with @justram

dlrudwo1269:
Hi @lintool, sorry for the delay here... I was getting different retrieval scores when including the image urls in the indexes so had to do some troubleshooting. It is fixed now and I updated all the index information. We can now show the images like the following screenshot

image

dlrudwo1269:
As for the next step, I can coordinate with @jasper-xian to incorporate the dense search to this page as well

docs/experiments-atomic-bm25.md
dlrudwo1269, line 1: Also, I think we may want to rename the qrels and topics here (and in other relevant files) to avoid confusion, e.g. `topics.atomic.validation.text.jsonl` to something like `topics.atomic.validation.text.lucene.jsonl`

justram (approved):

Issue #1438, opened 2023-02-02
Move 2CR scripts into pyserini itself


lintool:
Currently for 2CR, to generate the entire matrix, we need a dev installation, e.g.,

```bash
python scripts/repro_matrix/run_all_msmarco.py --collection v1-passage
```

Because the scripts aren't part of the package itself.

What if we moved the scripts themselves into `pyserini` module, so then we can do something like:

```bash
python -m pyserini.2cr.run_msmarco --all --collection v1-passage
```

and even something like:

```bash
python -m pyserini.2cr.run_msmarco --all --generate-report --collection v1-passage
```

to generate the actual HTML docs.

And even to generate each condition individually:

```bash
python -m pyserini.2cr.run_msmarco --collection v1-passage --condition blah
```

I can try to hack something up.

lintool:
hey @jasper-xian do you have cycles to wrap this up?

Currently, `miracl` and `mrtydi` have been moved over, so this works:

```
nohup python -m pyserini.2cr.miracl --all --directory runs --display-commands >& logs/log.miracl &
nohup python -m pyserini.2cr.mrtydi --all --directory runs --display-commands >& logs/log.mrtydi &
```

But we still have:

```
nohup python scripts/repro_matrix/run_all_beir.py >& logs/log.beir &
nohup python scripts/repro_matrix/run_all_odqa.py --topics nq >& logs/log.odqa.nq &
nohup python scripts/repro_matrix/run_all_odqa.py --topics tqa >& logs/log.odqa.tqa &
```

Can we convert the above 3 to `python -m pyserini.2cr...`?

lintool:
Done a while ago... closing.

PR #1641, opened 2023-09-18
Move beir and odqa 2cr into pyserini

jasper-xian:
WIP (still testing), addressing #1438


lintool:
hi @jasper-xian can we please keep the old odqa scripts? i.e., so we'll have two copies, the original in `scripts/` and the new 2CR. For BEIR we can remove the original in `scripts/`.

Rationale is that we have in our ECIR paper the repro commands for odqa in `scripts/`... so we want to still make sure it works.

jasper-xian:
Sure, we only need to keep `scripts/run_all_odqa.py` and `scripts/generate_odqa_html.py` right? So I can edit them to point to the correct html templates/yaml files?

jasper-xian:
Also, there's a very slight difference to the report generated from this script than as previous for odqa, but this is due to an inconsistency in definitions: `HITS_1K = set(['GarT5-RRF', 'DPR-DKRR', 'DPR-Hybrid'])` in the file to run everything but `HITS_1K = set(['GarT5-RRF', 'DPR-DKRR'])` in the html generation file.

lintool:
Yes, I think so. See: https://link.springer.com/chapter/10.1007/978-3-031-28241-6_10

This is what we have -
Screen Shot 2023-09-18 at 7 21 12 PM

So let's not break that command - although that command can be refactored itself and point to new yaml files, etc. Does that make sense?

lintool:
> Also, there's a very slight difference to the report generated from this script than as previous for odqa, but this is due to an inconsistency in definitions: `HITS_1K = set(['GarT5-RRF', 'DPR-DKRR', 'DPR-Hybrid'])` in the file to run everything but `HITS_1K = set(['GarT5-RRF', 'DPR-DKRR'])` in the html generation file.

@manveertamber @ronakice any comments there? I suppose we should generate everything?

jasper-xian:
Yeah everything is being generated but on the report we have `DPR-Hybrid` getting `--hits 100` but in the run_all we get `--hits 1000`

lintool:
Ah yes, in that case make consistent. Let's go with 1000?

jasper-xian:
@lintool good for review, I've played around with it a bit but didn't test _everything everything_, let me know if you want me to run them all

jasper-xian:
moved the yaml files for odqa back to `pyserini/resources/` fix the merge conflict, will shift them to `pyserini/2cr/` in another PR

lintool (approved):

lintool:
Hi @jasper-xian I'm fine with merging this. I'll run on my end and we can patch up additional issues as we discover them.

lintool:
And forgot to mention... thanks for all your hard work on this!

jasper-xian:
no problem! yeah happy to merge and tweak later, I'll send in a short PR moving the yaml files afterwards

PR #1642, opened 2023-09-21
Move integration tests into 2CR

lintool:
Some integration tests currently actually belong as 2CRs. Moving them over.


lintool:
I've verified on `tuna`, my Mac Studio, and my iMac Pro.

jasper-xian (approved):
lgtm!

lintool:
Here goes! 🤞

Issue #1658, opened 2023-09-28
Failure with BEIR regressions


lintool:
At Pyserini commit [`142c77`](https://github.com/castorini/pyserini/commit/142c774a303c906ee245913bc7e714b165074b77) (2023/09/25).
Minimal command that exposes the bug:

```
$ python -m pyserini.2cr.beir --dataset climate-fever --condition bm25-flat --display-commands
condition bm25-flat:
  - dataset: climate-fever

python -m pyserini.search.lucene \
  --index beir-v1.0.0-climate-fever.flat \
  --topics beir-v1.0.0-climate-fever-test \
  --output run.beir.bm25-flat.climate-fever.txt \
  --output-format trec \
  --batch 36 --threads 12 \
  --hits 1000 --bm25 --remove-query

      nDCG@10: 0.0000 [FAIL] expected 0.1651
      R@100  : 0.0000 [FAIL] expected 0.4249
      R@1000 : 0.0000 [FAIL] expected 0.6324
```

Bug is in preforming the evaluation:

```bash
$ python -m pyserini.eval.trec_eval -c -m ndcg_cut.10 beir-v1.0.0-climate-fever-test run.beir.bm25-flat.climate-fever.txt
Downloading https://search.maven.org/remotecontent?filepath=uk/ac/gla/dcs/terrierteam/jtreceval/0.0.5/jtreceval-0.0.5-jar-with-dependencies.jar to /home/jimmylin/.cache/pyserini/eval/jtreceval-0.0.5-jar-with-dependencies.jar...
/home/jimmylin/.cache/pyserini/eval/jtreceval-0.0.5-jar-with-dependencies.jar already exists!
Skipping download.
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/tuna1/scratch/jimmylin/pyserini/pyserini/eval/trec_eval.py", line 70, in <module>
    run = pd.read_csv(args[-1], delim_whitespace=True, header=None)
  File "/tuna1/scratch/jimmylin/virtualenv/pyserini-dev3/lib/python3.10/site-packages/pandas/io/parsers/readers.py", line 948, in read_csv
    return _read(filepath_or_buffer, kwds)
[...]

PR #1660, opened 2023-09-29
Fix for BEIR regression failure described in #1658

lintool:
As described in #1658 - changing the csv read engine to 'python' seems to fix the issue. Rerunning all regressions to confirm.


lintool:
@jasper-xian I've confirmed that this patch fixes the issues on `linux.cs` and `tuna`. Regression passes without issue.

jasper-xian (approved):

PR #1663, opened 2023-09-30
add remove_dups argument for dense retrieval

yogeswarl:
Hi @lintool, Apologies for the previous issue with the tests. I have fixed the issue and here is the new pull request. 
I noticed the problem was with an indentation error. 
This pull request solves a problem where datasets with smaller document collections may return duplicate documents in dense search results. 
Here is a screenshot of the test cases passing correctly.
Screenshot 2023-09-30 at 11 05 41 AM


pyserini/search/faiss/_searcher.py
```diff
@@ -477,6 +479,14 @@ def search(self, query: Union[str, np.ndarray], k: int = 10, threads: int = 1, r
             distances, indexes = self.index.search(emb_q, k)
             distances = distances.flat
             indexes = indexes.flat
+            if remove_dups == True:
```
lintool, line 482:
```suggestion
            if remove_dups:
```

I think you can just do this?

Please re-verify tests.

lintool (changes requested):

pyserini/search/faiss/_searcher.py
yogeswarl: This works! Thanks for the advice

lintool (approved):

PR #1668, opened 2023-10-03
De-lint + minor refactoring for PEP8 compliance

lintool:
@jasper-xian I think a light review is sufficient. I've verified that the 2CRs still run from my end.

### DIFF
--- a/pyserini/2cr/beir.py
+++ b/pyserini/2cr/beir.py
@@ -265,7 +265,7 @@ def run_conditions(args):
     end = time.time()
 
     print('\n')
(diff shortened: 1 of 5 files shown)


lintool:
These worked fine on my iMac Pro:

```
python -m pyserini.2cr.odqa --all --directory runs/ --topic tqa --display-commands >& logs/log.odqa.tqa
python -m pyserini.2cr.odqa --all --directory runs/ --topic nq  --display-commands >& logs/log.odqa.nq
```

jasper-xian (approved):
the linting changes look good to me! I also ran some sanity 2cr checks on my end and checks out as well

PR #1670, opened 2023-10-06
Tweak regression numbers to pin to tuna results as ground truth

lintool:
After recent upgrade of `tuna` to Unbuntu 22, we're using the machine a "ground truth" for the regression numbers. This PR pins the results to `tuna` - thus small changes from before.


jasper-xian (approved):
This looks good to me, just want to ask that these small changes shouldn't affect anything in `orca` right?

lintool:
> This looks good to me, just want to ask that these small changes shouldn't affect anything in `orca` right?

`orca` is still on Ubuntu 18... so it needs to be upgraded... and once we do, I'll rerun and recalibrate...

PR #1674, opened 2023-10-08
2CR for dense AToMiC retrieval

jasper-xian:
AToMiC dense 2CR for model ViT-L-14.laion2b_s32b_b82k

Related PRs: https://github.com/castorini/anserini/pull/2219 https://github.com/castorini/anserini-tools/pull/57 https://github.com/castorini/pyserini-data/pull/6

### DIFF
--- a/pyserini/2cr/atomic_html.template
+++ b/pyserini/2cr/atomic_html.template
@@ -1,0 +1,242 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
+    <title>Pyserini Reproductions</title>
+    <!-- Font Awesome -->
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.11.2/css/all.css" />
+    <!-- Google Fonts Roboto -->
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap" />
+    <!-- MDB -->
+   <link href="https://cdnjs.cloudflare.com/ajax/libs/mdb-ui-kit/4.0.0/mdb.min.css" rel="stylesheet" />
+
+    <style>
+tr.hide-table-padding td {
+  padding: 0;
+}
+
+.expand-button {
+  position: relative;
+}
+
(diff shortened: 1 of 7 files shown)


jasper-xian:
Hey @lintool, am looking to add some topic files to anserini so we can do a 2CR. These are json files that follow the [`DefaultQueryIterator`](https://github.com/castorini/pyserini/blob/a935a365329db5a78f7c573ddae74b948fb7424f/pyserini/query_iterator.py#L84) format, so basically:

```
{"<id>": {"title": "<some text>"}, ....}
```

Which anserini TopicReader would work for this sort of topic file? Let me know and I'll add a corresponding PR in anserini as well.

lintool:
Take a look here? https://github.com/castorini/anserini/tree/master/src/main/java/io/anserini/search/topicreader

Maybe this one? https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/topicreader/JsonStringTopicReader.java

pyserini/2cr/atomic_html.template
```diff
@@ -0,0 +1,242 @@
...
+  <li>Base: AToMiC base dataset</li>
+  <li>Large: AToMiC large dataset</li>
+</ul>
+
+<div class="table-responsive">
+  <table class="table">
+    <thead>
+      <tr>
+        <th class="headertop"></th>
+        <th class="headertop"></th>
```
justram, line 148: How about we add a description here to identify the query set?
E.g., val queries

justram (changes requested):
Looks good. Just need to change a few names to align the paper:
1. Eval settings: ~validation~ -> small
2. A column to show the query set we used in the table: val queries

pyserini/2cr/atomic_html.template
jasper-xian: not too sure what you mean by this, does it correspond to any columns in the [data spreadsheet](https://docs.google.com/spreadsheets/d/1wSi_79Qx3GA1WAirwvoapiWJ4m2bPRM_rtUWRZ2qRIo/edit#gid=0)?

justram: I believe it would be beneficial to clarify that the results pertain to validation queries, as demonstrated in the [Miracl HTML template](https://github.com/castorini/pyserini/blob/2cab9cbf53139a821495f8bd46c5a45b14fea333/pyserini/2cr/miracl_html_table.template#L6). The data spreadsheet currently displays scores exclusively for the validation just for reference, but there are also test queries available.

justram (approved):
LGTM

PR #1717, opened 2023-11-16
Update installation instructions

lintool:
Cleaned up a lot of outdated information...

A lot of the "tips and tricks" are outdated, but it's hard to determine remains relevant... so, I've killed most cruft. If the issues come back up again we can re-add to installation guide?


ronakice (approved):
LGTM!

PR #1721, opened 2023-11-19
Update AToMiC demo page to support dense search

dlrudwo1269:
Update AToMiC demo page to support the dense searchers.

One issue is that it is difficult to navigate the image ids when searching for pre-encoded queries for images. I was thinking of adding a index->topic file mapping in the server, and in `search_options()` we  search the topic file to get the image_urls of the matching options to pass to the client side. Did you have any other suggestions on how to better handle this @jasper-xian ?

Demo video: https://youtu.be/ukb4Bc0XzW4

### DIFF
--- a/pyserini/demo/atomic.py
+++ b/pyserini/demo/atomic.py
@@ -29,41 +29,134 @@ from functools import partial
 from typing import Callable, Optional, Tuple, Union
 
 from flask import Flask, render_template, request, flash, jsonify
-from pyserini.search import LuceneSearcher, FaissSearcher
+from pyserini.search import LuceneSearcher, FaissSearcher, QueryEncoder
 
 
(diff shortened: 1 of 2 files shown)


dlrudwo1269:
@jasper-xian @justram would appreciate a review here!

jasper-xian:
Not sure I'm the right person to comment on this, but the way we did pre-encoded image queries was something like this:

Topics: {'id': <image-id>, 'title': <image-id>}
Pre-encoded queries: {'id': <image-id>, 'text': <image-id>, 'embedding': <image-emb>}

So you could use the image-id to search for the right image. I'm not sure what this mapping stuff relates to in general though, but I think the app looks good! @justram can probably comment more about it

Issue #1720, opened 2023-11-18
Normalize embeddings when using a custom dense encoder?


orionw:
Hi there!

Thanks for the great work on Pyserini! I had a naive question that I can't seem to find the answer to.

I'd like to use E5 (https://huggingface.co/intfloat/e5-large but other models are similar) and they recommend normalizing the embeddings. I can't find an option for that in Pyserini to do that, when it's not a dense/sparse combination.  I'd like to do just dense encodings, but make sure the embeddings are normalized to properly use E5.

I've been using `pyserini.encode` so far but don't see any options in there for it. Does Pyserini support this?

MXueguang:
Hi @orionw,
The AutoDocumentEncoder has the argument `l2_norm` for initialisation.
https://github.com/castorini/pyserini/blob/b931e5293252aaf5cc00e9045b6aef4a70ca182d/pyserini/encode/_auto.py#L25
However, the option is not exposed in `pyserini.encode` as an CLI argument.
I'll create a pull request to add this.

orionw:
Thanks a bunch @MXueguang!

MXueguang:
https://github.com/castorini/pyserini/pull/1722

PR #1722, opened 2023-11-21
config pooling and l2norm for autoencoder

MXueguang:
https://github.com/castorini/pyserini/issues/1720

### DIFF
--- a/pyserini/encode/__main__.py
+++ b/pyserini/encode/__main__.py
@@ -33,9 +33,8 @@ encoder_class_map = {
     "openai-api": OpenAIDocumentEncoder,
     "auto": AutoDocumentEncoder,
 }
-ALLOWED_POOLING_OPTS = ["cls","mean"]
 
-def init_encoder(encoder, encoder_class, device):
+def init_encoder(encoder, encoder_class, device, pooling, l2_norm)
     _encoder_class = encoder_class
 
     # determine encoder_class
@@ -52,6 +51,7 @@ def init_encoder(encoder, encoder_class, device):
         # if none of the class keyword was matched,
         # use the AutoDocumentEncoder
         if encoder_class is None:
+            _encoder_class = "auto"
             encoder_class = AutoDocumentEncoder
 
     # prepare arguments to encoder class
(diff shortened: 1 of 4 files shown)


MXueguang:
@lintool  do we want to make `prefix` configurable for `AutoEncoder` as well? E.g. e5 is prepending "query: " and "document: " to the text.

lintool:
> @lintool do we want to make `prefix` configurable for `AutoEncoder` as well? E.g. e5 is prepending "query: " and "document: " to the text.
> 
> or it can be handled in raw data pre-processing.

I think yes?

@crystina-z @ronakice can you review?

MXueguang:
okay updated, ready for review

marcospiau:
Hi @MXueguang, I was looking into your PR because, coincidentally, I will also need to use e5 models as encoders. Is there any particular reason for hard coding these values for sentence-transformer models? `kwargs.update(dict(pooling='mean', l2_norm=True))` (here is the permalink: https://github.com/MXueguang/pyserini/blob/7d40eef1f9b3baf771c5e307265b8dba40719c84/pyserini/encode/__main__.py#L60C58-L60C58).

I noticed this because the embeddings when using `pyserini.encode` with `--prefix "passage:"` and `--encoder-class auto` were the same as if no prefix was used. Using `--encoder-class sentence-transformers` resulted in the expected values.

Also, a related question: Are there any conceptual differences between "auto" and "sentence-transformers" encoder classes?

MXueguang:
Hi @marcospiau,
the hard code sentence-transformer setting was originally to reproduce https://github.com/castorini/pyserini/blob/master/docs/experiments-sbert.md

I just realise they all share the AutoDocumentEncoder, so should be same?

> I noticed this because the embeddings when using pyserini.encode with --prefix "passage:" and --encoder-class auto were the same as if no prefix was used. 

this issue was found in this PR?

marcospiau:
@MXueguang Thanks, the hardcoded part now makes sense. Sorry, but by mistake I switched the `--encoder-class` part. Using `--encoder-class auto` and specifying a prefix works as expected; using `--encoder-class sentence-transformers`  and a prefix results in the same scores as not using a prefix.

Thanks!

lintool (approved):

PR #1753, opened 2023-12-27
Fixed broken tests 

lintool:
+ from Anserini refactoring (renamed args)
+ from pooling and l2norm config for autoencoder (https://github.com/castorini/pyserini/pull/1722/)


MXueguang (approved):
lgtm

PR #1801, opened 2024-03-01
Fix unittest warnings and failure

manveertamber:
Fixed failure in test_bm25_vectorizer_train where values were slightly off, perhaps due to an underlying update

Added Pyarrow requirement to avoid the error:

```
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
(to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
but was not found to be installed on your system.
If this would cause problems for you,
please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
```

Some other small changes as well to deal with warnings


lintool (approved):
thanks for cleaning these issues up!

PR #1807, opened 2024-03-05
Clean up MS MARCO V1 passage + BEIR regressions 

lintool:
Implemented much simpler heuristic for flaky tests where scores vary depending on OS: anything within 0.005, just output "OKish". Everything else, code up case-by-case exceptions.


sahel-sh (approved):
LGTM!

PR #1819, opened 2024-03-17
Refactor 2CR

lintool:
Recalibrate 2CR scores wrt new dependencies.


lintool:
Added blanket catch-all: if results are within 0.0005, just call it "OKish".
Deal with other differences on a case-by-case basis.

sahel-sh (approved):
LGTM!

PR #1847, opened 2024-03-30
Add to onboarding doc with updated scores

Lindaaa8:


lintool:
Can you describe your setup? OS, versions, etc.?

Odd that you're finding slight differences...

lintool:
Hi @Lindaaa8 let's not change the effectiveness numbers... might be due to your idiosyncratic setup because others haven't noticed... if we hear more from others we can circle back to tweak?

So please revert those changes?

lintool:
I think you're missing the final step in the in onboarding path?

Lindaaa8:
Added the missing final step.

lintool (approved):

PR #1854, opened 2024-04-07
Support multiple alpha settings in `pyserini/fusion`

hanhainebula:
Support multiple alpha settings in `pyserini/fusion`


lintool:
hi @hanhainebula can you confirm that `python -m unittest` still passes?

Also, please add a test case to https://github.com/castorini/pyserini/blob/master/tests/test_fusion.py

lintool:
Thanks in advance for your contribution!

hanhainebula:
Hi, I just commit a new PR: https://github.com/castorini/pyserini/pull/1858. I add a simple test case to `tests/test_fusion.py`, and I confirm that `python -m unittest` still passes.

PR #1863, opened 2024-04-14
Add OpenAI text-embedding-3-large msmarco 2cr

manveertamber:


lintool:
@manveertamber can you please add a README?

manveertamber:
> @manveertamber can you please add a README?

Where would the README go? I'm seeing anything similar for ada2 for example except for in docs/prebuilt-indexes.md

lintool:
README goes in `pyserini/resources/index-metadata`.

>  I'm seeing anything similar for ada2 for example except for in docs/prebuilt-indexes.md

Yea... unfortunately, we haven't been as good about documenting as we should have... but let's change that please.

manveertamber:
> README goes in `pyserini/resources/index-metadata`.
> 
> > I'm seeing anything similar for ada2 for example except for in docs/prebuilt-indexes.md
> 
> Yea... unfortunately, we haven't been as good about documenting as we should have... but let's change that please.

Sounds good

lintool (approved):

PR #1866, opened 2024-04-18
Add Colbert MLX

ToluClassics:
Could probably do with some redesign but here's a first pass at integrating MLX into Pyserini with ColBERT.

The tests and test outputs are similar to the same tests for the PyTorch hugging face implementation

### DIFF
--- a/pyserini/encode/__init__.py
+++ b/pyserini/encode/__init__.py
@@ -15,11 +15,13 @@
 #
 
 from ._base import DocumentEncoder, QueryEncoder, JsonlCollectionIterator,\
-    RepresentationWriter, FaissRepresentationWriter, JsonlRepresentationWriter, PcaEncoder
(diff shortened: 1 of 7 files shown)


ToluClassics:
Side Note: We have some duplicated code e.g. https://github.com/castorini/pyserini/blob/184a212e7d578fac453ead64f7f796bc2e44bcf2/pyserini/encode/_tct_colbert.py#L72

and 

https://github.com/castorini/pyserini/blob/184a212e7d578fac453ead64f7f796bc2e44bcf2/pyserini/search/faiss/_searcher.py#L147


Is there a particular reason we are keeping both?

lintool:
@jacklin64 @justram can either of you take a look at this?

ToluClassics:
Bump on this @jacklin64 @justram .

PR #1882, opened 2024-05-09
Onboarding tasks billycz8

billycz8:
System setup:

OS: macOS Ventura 13.4
Memory: 16GB
Chip: Apple M1 Pro
Python Version: 3.10.13
Java Version: 21.0.3
Maven: 3.9.2

Everything worked perfectly!

Suggestion: When I went through Pyserini: BGE-base Baseline for NFCorpus / Data Prep, I was curious why we added Q0 instead of 0 as the artifact in Anserini. It seems to be used in trec_eval.py, and I think explaining it would be helpful for others to understand the reason of the change. (I have added the change in the PR, we could reword or ignore that change if it is not useful.)

### DIFF
--- a/docs/experiments-nfcorpus.md
+++ b/docs/experiments-nfcorpus.md
@@ -71,7 +71,7 @@ This command-line invocation does the trick:
 tail -n +2 collections/nfcorpus/qrels/test.tsv | sed 's/\t/\tQ0\t/' > collections/nfcorpus/qrels/test.qrels
 ```
 
-Q0 instead of 0 (historical artifact mentioned in [Anserini: Start Here](https://github.com/castorini/anserini/blob/master/docs/start-here.md)) marks as trec format in pyserini to avoid extraneous conversion from msmarco in [trec_eval.py](https://github.com/castorini/pyserini/blob/c945c50c3e22e3c6ecae50a55aed48853617acc0/pyserini/eval/trec_eval.py#L67).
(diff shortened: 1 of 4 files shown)


docs/experiments-nfcorpus.md
```diff
@@ -71,6 +71,8 @@ This command-line invocation does the trick:
 tail -n +2 collections/nfcorpus/qrels/test.tsv | sed 's/\t/\tQ0\t/' > collections/nfcorpus/qrels/test.qrels
 ```
 
+Q0 instead of 0 (historical artifact mentioned in [Anserini: Start Here](https://github.com/castorini/anserini/blob/master/docs/start-here.md)) marks as trec format in pyserini to avoid extraneous conversion from msmarco in [trec_eval.py](https://github.com/castorini/pyserini/blob/master/pyserini/eval/trec_eval.py).
```
lintool, line 74: Thanks for noticing. Actually, this isn't technically correct... so let's just remove?

lintool (changes requested):

docs/experiments-nfcorpus.md
billycz8: Oh right, I read the code over and that Q0 is for checking the format of the search output. I'll remove the lines. Then, adding in the qrels the Q0 or 0 does not really matter here?

lintool (approved):

PR #1898, opened 2024-05-27
Update requirements nmslib compatible with late pybind11

GrigoriiTarasov:


lintool:
Our instructions specify use of conda to install nmslib: https://github.com/castorini/pyserini/blob/master/docs/installation.md

Have you tried that?

GrigoriiTarasov:
Yes, I done both Linux instruction and from source:
without pybind11 update they led to
https://github.com/pybind/pybind11/discussions/4333
and now installing from scratch the pybind11 in which above is solved is in conflict with too new nmslib.

lintool:
What Python version are you on?

lintool:
> python 3.11.9 h955ad1f_0

Hrm. Our entire group is still on Python 3.10... wondering if this makes a difference?

GrigoriiTarasov:
Oh I see, I started from
```
pip install torch faiss-cpu
pip install pyserini
```
as it was under Linux and thought

```
conda create -n pyserini python=3.10 -y
conda activate pyserini

```
related to Mac as it under Mac section, but this part has to be common for all the platforms.

lintool (approved):

lintool:
Shouldn't hurt... but also untested wrt other parts of Pyserini.

PR #1901, opened 2024-05-29
Bind anserini server application

16BitNarwhal:
Binded [anserini server and web application](https://github.com/castorini/anserini/blob/master/docs/fatjar-regressions/fatjar-regressions-v0.36.1.md#webapp-and-rest-api)

Run with `python -m pyserini.server.AnseriniApplication --server.port=8081`

### DIFF
--- a/pyserini/server/AnseriniApplication/__main__.py
+++ b/pyserini/server/AnseriniApplication/__main__.py
@@ -1,0 +1,29 @@
+#
+# Pyserini: Reproducible IR research with sparse and dense representations
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


pyserini/server/AnseriniApplication/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+from pyserini.pyclass import autoclass
+
+if __name__ == '__main__':
+    args = sys.argv[1:]
+
+    JServer = autoclass('io.anserini.server.Application')
+    JServer.main(args)
+
+    while True:
+        time.sleep(60)
```
lintool, line 29: Hrm... there's gotta be a better way to do this?

16BitNarwhal: I tried using signals, which should wait until it gracefully catches a terminating signal and stop the server. Hope that works better!

lintool (approved):

PR #1954, opened 2024-08-04
Refactor tests that verify pyserini output against anserini output

lintool:
Before, tests would assume that Anserini indexes existed in `../anserini/indexes` but this PR breaks the dependency, with #1953


UShivani3 (approved):

lintool:
Both have been verified to work:

```bash
nohup python -m unittest discover -s integrations/sparse >& logs/log.sparse &
nohup python -m unittest discover -s integrations/clprf >& logs/log.clprf &
```

Issue #1611, opened 2023-08-31
Refactor installation instructions


lintool:
@yilinjz @UShivani3 et al. recently had issues getting Pyserini installed... I think we should refactor the installation instructions?

+ I had a start here: https://github.com/castorini/pyserini/pull/1609 but I don't think it's enough.
+ M1/M2 issues documented here: https://github.com/castorini/pyserini/pull/1599

@sahel-sh since you did the onboarding recently, you might share experiences also?

cc @Andrwyl

lintool:
Also, something to think about - should we move some dependencies into "optional"?
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

lintool:
On Ubuntu 18.04.6. I did the following:

```
conda env list
conda create -n pyserini-pypi-test python=3.9
conda activate pyserini-pypi-test

conda install -c pytorch pytorch faiss-cpu

pip install pyserini
```

The important package versions:

```bash
$ pip list | egrep '(numpy|pyjnius|transformers|torch|sentencepiece|faiss|scikit-learn|lightgbm|spacy|pandas)\s'
faiss              1.7.4
lightgbm           4.0.0
numpy              1.25.2
pandas             2.1.0
pyjnius            1.5.0
scikit-learn       1.3.0
sentencepiece      0.1.99
spacy              3.6.1
torch              2.0.1
transformers       4.32.1
```

Above config, will get this error for `faiss`:

```
ImportError: libmkl_intel_lp64.so.1: cannot open shared object file: No such file or directory
```

Issue described here: https://github.com/facebookresearch/faiss/issues/2890

Fix is to install `mkl` separately:

```
conda install mkl=2021
```

@Andrwyl in the student cs env, did you have any issues?

Andrwyl:
For the student env, from the beginning all the way to the end there were no issues. Following the instructions exactly should work all the way to the end of onboarding. The only issue that you run into is there is a cpu limit on every student account which causes you to be kicked off while doing dense retrieval. `ulimit -t unlimited` gives you unlimited cpu time.

https://uwaterloo.ca/computer-science-computing-facility/teaching-hosts shows the specs of the student computers. The resources are shared so its possible that on high usage days you may end up failing (has not happened to me though).

To access, any uwaterloo student can just do `ssh userid@linux.student.cs.uwaterloo.ca`, but I'm pretty sure nearly everyone has used this for a 2nd year CS class

lintool:
[...]

PR #1962, opened 2024-08-16

emily-emily:
Successfully reproduced results for conceptual-framework, conceptual-framework2, experiments-msmarco-passage, experiments-nfcorpus

Reproduction Environment:
UW Linux server
Java Version: 21.0.3
Maven Version: 3.9.8
Python version: 3.10.12

Setup steps for the linux environment:
- The existing Java and Maven versions are too old. I installed the versions above and updated `PATH` and `JAVA_HOME`.
- Created a Python venv.
- Had to install faiss and torch manually: `pip install faiss-cpu torch`

I believe it was at the index creation step that the default allocated resources are not enough.
- `ulimit -n 4096`: increase number of file pointers from 1024 to 4096
- `ulimit -u 8192`: increase number of processes from 4096 to 8192
- `ulimit -t unlimited`: increase CPU time, as suggested in [this issue](https://github.com/castorini/pyserini/issues/1611#issuecomment-1703253700)


ronakice (approved):
LGTM! Jimmy is AFK, and I feel you're probably looking to do a URA? Reach out at `rpradeep@uwaterloo.ca` and we can figure out some tasks so you are not blocked

PR #1965, opened 2024-08-21

nicoella:
Successfully reproduced all results.

Reproduction environment:
Apple M1 Pro / macOS Monterey 12.1
Java 21.0.2
Maven 3.9.8
Python 3.11.8


docs/experiments-nfcorpus.md
```diff
@@ -26,10 +26,10 @@ How to actually train such models will be covered later.
```
lintool, line 26: Something weird happened here? Please revert and only add in the repro log entry?

lintool (changes requested):
Please also fix conflicts?

PR #1974, opened 2024-08-29
Seperate cqadupstack rows in BEIR 2CR

manveertamber:
Brought the cqadupstack rows in the BEIR 2CR into a separate table. Created an averaged row in the main table.


lintool (approved):
🚢 it!

PR #1986, opened 2024-09-14
Switch from using jtrec_eval to trec_eval

lintool:
trec_eval is already bundled with Anserini. Also, clean up trec_eval output - i.e., not so noisy.


jasper-xian (approved):
lgtm!

PR #1988, opened 2024-09-16
Update installation.md

andrewxucs:
- discourage other python versions
- reminder to use the newest version of Anserini
- troubleshooting with pip install

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,10 +1,10 @@
 # Pyserini: Detailed Installation Guide
 
-Pyserini is built on Python 3.10 (other versions might work, but YMMV).
+Pyserini is built on Python 3.10 (other versions might not work).
 At a high level, we try to keep our [`requirements.txt`](../requirements.txt) up to date.
 Pyserini has a number of important dependencies:
 
-+ For sparse retrieval, Pyserini depends on [Anserini](http://anserini.io/), which is built on Lucene.
++ For sparse retrieval, Pyserini depends on [Anserini](http://anserini.io/), which is built on Lucene. Make sure that Anserini is up to date, otherwise version compatibility may be an issue.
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -1,10 +1,10 @@
 # Pyserini: Detailed Installation Guide
 
-Pyserini is built on Python 3.10 (other versions might work, but YMMV).
+Pyserini is built on Python 3.10 (other versions might not work).
```
lintool, line 3: I think "might not work" is a bit too strong... which is why I wrote YMMV.

docs/installation.md
```diff
@@ -146,6 +146,7 @@ Then, in the `pyserini` clone, use `pip` to add an ["editable" installation](htt
 
 ```bash
 pip install -e .
+# If an error occurred during installation, try this: python -m pip install -e .
```
lintool, line 149: I think this is idiosyncratic to your setup... you're the first to have encountered the error you relayed to me over email...

docs/installation.md
```diff
@@ -1,10 +1,10 @@
 # Pyserini: Detailed Installation Guide
 
-Pyserini is built on Python 3.10 (other versions might work, but YMMV).
+Pyserini is built on Python 3.10 (other versions might not work).
 At a high level, we try to keep our [`requirements.txt`](../requirements.txt) up to date.
 Pyserini has a number of important dependencies:
 
-+ For sparse retrieval, Pyserini depends on [Anserini](http://anserini.io/), which is built on Lucene.
++ For sparse retrieval, Pyserini depends on [Anserini](http://anserini.io/), which is built on Lucene. Make sure that Anserini is up to date, otherwise version compatibility may be an issue.
```
lintool, line 7: if you pip install, you don't need this... this only applies if you want a dev installation - further down.

lintool:
hi @andrewxucs thanks for your suggestions, but I don't think this improves the documentation...

Issue #1999, opened 2024-09-30
LuceneSearcher.search `searchFields` JSimpleSearcher method typo


lgrz:
Method not found error occurs in `LuceneSearcher.search` when called with `fields` and `query_generator` arguments set. Refer to line 131 of [_searcher.py][bork]. It is trivial to fix.

For example:

```
from pyserini.search.lucene import LuceneSearcher
searcher = LuceneSearcher('indexes/foo')
query_generator = JDisjunctionMaxQueryGenerator(1.0)
hits = searcher.search("foo", k=10, query_generator=query_generator, fields={'title': 10.0})
...
```

[bork]: https://github.com/castorini/pyserini/blob/master/pyserini/search/lucene/_searcher.py#L131
[ref]: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java#L731

lintool:
Hi @lgrz thanks for reporting this bug. Would you mind adding a test case and sending a PR?

PR #2012, opened 2024-10-12
Fix for JSimpleSearcher method typo in LuceneSearcher

lgrz:
First pass on patch and tests to address #1999.


lintool (approved):
@lgrz Thanks for your contribution!

PR #2008, opened 2024-10-08
Refactoring to remove duplicates in pyserini.encode and pyserini.search.faiss

lintool:
Addresses #1728 which will in turn contributes to #2006

PR #2017, opened 2024-10-26
reproduced tasks for onboarding process

pxlin-09:
Hi all,
I am using a Macbook Pro with M2 chip and vscode

some issues I encountered:

1. not all unit tests passed after [development installation](https://github.com/castorini/pyserini/blob/master/docs/installation.md#development-installation), but I guess its fine since I saw a response saying that's ok in a ticket
2. Is IndexReader renamed to LuceneIndexReader?
3. AutoQueryEncoder is in pyserini.encode but the md files sometimes tells user to import from pyserini.search.faiss, I included the fix in this pr


lintool:
Hi @pxlin-09 thanks for fixing these issues, we had recent refactoring #2008 that changed a lot of things. Onboarding path must have gotten out of alignment.

lintool (approved):

PR #2029, opened 2024-11-24
Update experiments-msmarco-passage.md

Divyajyoti02:
Reproduced successfully on Macbook Air with M2 chip, 16 GB RAM, and 256 GB memory.

One tip while doing the development installation of pyserini. It seems that we need to install some packages to build everything. Therefore it would be useful to add the following installations to a setup.py file:

pip install numpy==1.26.4 blis thinc spacy

Additionally, nmslib is also necessary, but somehow the pip wheel breaks. I used conda for installation:

conda install -c conda-forge nmslib


lintool:
Please continue along the onboarding path until you reach the end of the exercises, and then we'll take a look all at once.

Divyajyoti02:
Done.

lintool (approved):

PR #2044, opened 2024-12-08
Refactor code to support Arctic models on MS MARCO v2.1

lintool:
General cleanup of both code and docs.


lintool:
+ Fixed broken tests
+ Updated md5 checksum of tarballs for downloading

ronakice (approved):
LGTM!

PR #2048, opened 2024-12-18
Update trec_eval.py for Python 3.12 regression

m-krastev:
Fixes Python 3.12 incompatibility as described in: https://github.com/python/cpython/issues/106614


lintool:
Hi @m-krastev thanks for your PR!

Since we're on Python 3.10, as clearly documented [here](https://github.com/castorini/pyserini/?tab=readme-ov-file#-installation), this isn't an issue (for now)....

PR #2053, opened 2024-12-23
Update _base.py

arthur-75:
Thank you for this great package! I encountered a small formatting issue when reading a JSON file after saving it using `index_reader.dump_documents_BM2`. I found a simple solution that worked well for me: adding `encoding="utf-8"` while saving the file.

Additionally, regarding the issue with **Arguana** that I raised recently, I discovered the root cause. It is related to the document with docid **'training-society-iasihbmubf-pro06a'**, which appears to be empty. To address such cases, I would like to suggest a workaround for handling empty documents until the Java code is fixed. This might help others avoid similar issues. Thank you for your consideration!

### DIFF
--- a/pyserini/index/lucene/_base.py
+++ b/pyserini/index/lucene/_base.py
@@ -369,7 +369,7 @@ class LuceneIndexReader:
         except Exception as e:
             # In case if doc is empty
             exception_type = type(e).__name__  # Get the exception's type name
-            if "JavaException" == exception_type:
+            if 'JavaException' == exception_type:
                 return {}
             else:
                 # Re-raise the exception if it doesn't match the desired name
--- a/pyserini/index/lucene/_base.py
+++ b/pyserini/index/lucene/_base.py
@@ -367,7 +367,7 @@ class LuceneIndexReader:
         try:
             doc_vector_map = self.object.getDocumentVector(self.reader, docid)
         except Exception as e:
-            #in case if doc is empty
+            # In case if doc is empty
             exception_type = type(e).__name__  # Get the exception's type name
             if "JavaException" == exception_type:
                 return {}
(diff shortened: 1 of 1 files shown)


pyserini/index/lucene/_base.py
```diff
@@ -364,7 +364,16 @@ def get_document_vector(self, docid: str) -> Optional[Dict[str, int]]:
         Optional[Dict[str, int]]
             A dictionary with analyzed terms as keys and their term frequencies as values.
         """
-        doc_vector_map = self.object.getDocumentVector(self.reader, docid)
+        try:
+            doc_vector_map = self.object.getDocumentVector(self.reader, docid)
+        except Exception as e:
+            #in case if doc is empty
```
lintool:
```suggestion
            # In case if doc is empty
```

pyserini/index/lucene/_base.py
```diff
@@ -569,7 +578,7 @@ def dump_documents_BM25(self, file_path, k1=0.9, b=0.4):
             BM25 b parameter.
         """
 
-        f = open(file_path, 'w')
+        f = open(file_path, 'w', encoding="utf-8")
```
lintool:
```suggestion
        f = open(file_path, 'w', encoding='utf-8')
```

pyserini/index/lucene/_base.py
```diff
@@ -364,7 +364,16 @@ def get_document_vector(self, docid: str) -> Optional[Dict[str, int]]:
         Optional[Dict[str, int]]
             A dictionary with analyzed terms as keys and their term frequencies as values.
         """
-        doc_vector_map = self.object.getDocumentVector(self.reader, docid)
+        try:
+            doc_vector_map = self.object.getDocumentVector(self.reader, docid)
+        except Exception as e:
+            # In case if doc is empty
+            exception_type = type(e).__name__  # Get the exception's type name
+            if "JavaException" == exception_type:
```
lintool:
```suggestion
            if 'JavaException' == exception_type:
```

lintool (approved):

PR #2099, opened 2025-04-20
Update conceptual-framework3.md

JJGreen0:
Student Linux Environment 

I figured out how to extend the CPU time limit. 

Had to make the directory: 

mkdir  encode 
cd encode 
mkdir nfcorpus.splade 

My token didn’t have the necessary permissions.  
Particularly: “Read access to contents of all public gated repos you can access” 
Would be nice to make this explicit  

I think the python code for encoding the corpus should have: 
 
doc_id = data["id"] 
text = data["contents"] 

Instead. 

I noticed that my encoding process kept getting killed at around 1300 lines processed. So I had to split the corpus into 4 1000 line chunks and encode them separately. Probably useful to know for other users of the student Linux environment. Not sure if there is a different workaround. 

All else worked as intended

### DIFF
--- a/docs/conceptual-framework3.md
+++ b/docs/conceptual-framework3.md
@@ -83,7 +83,8 @@ huggingface-cli login
 You’ll be prompted to enter your Hugging Face API token. You can generate a token from your Hugging Face account settings:
 1. Go to https://huggingface.co/settings/tokens.
 2. Click **"New token"** to generate a token.
-3. Copy the token and paste it into the terminal when prompted.
+3. For your token's permissions, give “Read access to contents of all public gated repos you can access”.
+4. Copy the token and paste it into the terminal when prompted.
 
 We are now all set to use SPLADE-v3 model!
 
--- a/docs/conceptual-framework3.md
+++ b/docs/conceptual-framework3.md
@@ -59,6 +59,8 @@ with open('collections/nfcorpus/pyserini-corpus/corpus.jsonl', 'w') as out:
             out.write(s + '\n')
 ```
 
+Note that we do not actually feed this munged file to the Lucene indexer in this guide, as we did with the previous ones. Thus, we don't really need the file that results from this munging. However, it's still good to have in case you want to swap back to BM25.
+
 We can then setup to use SPLADE-v3:
 First, we need to request access to SPLADE-v3 model on Hugging Face since it is gated:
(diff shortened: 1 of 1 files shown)


docs/conceptual-framework3.md
```diff
@@ -120,8 +120,8 @@ with open(corpus_file, "r") as infile, open(output_file, "w") as outfile:
         try:
             # Load the document
             data = json.loads(line)
-            doc_id = data["_id"]
-            text = data["title"] + " " + data["text"]  # Combine title and text
+            doc_id = data["id"]
```
lintool, line 123: Can you `git blame` and tag whoever worked on this before to reconcile/confirm?

lintool (commented):
Yes, this doc hasn't been vetted yet... feel free to tweak directly as necessary.

docs/conceptual-framework3.md
JJGreen0: @clides

lintool:
Looping in @lilyjge

docs/conceptual-framework3.md
clides: Oh yes, the code should be changed to id and contents because in the first step, the original corpus's columns are changed into just id and contents

```python
import json

with open('collections/nfcorpus/pyserini-corpus/corpus.jsonl', 'w') as out:
    with open('collections/nfcorpus/corpus.jsonl', 'r') as f:
        for line in f:
            l = json.loads(line)
            s = json.dumps({'id': l['_id'], 'contents': l['title'] + ' ' + l['text']})
            out.write(s + '\n')
 ```

Comment (lintool) [reply docs/conceptual-framework3.md:123]: @JJGreen0 can you take a look?

**Comment by lilyjge** [comment]
I can confirm @JJGreen0 's tweaks are necessary to repro successfully. However, in my opinion, a better way to tweak it might be to skip the first step of making a whole new corpus file, change the corpus_file in the encoding portion to the original corpus at "collections/nfcorpus/corpus.jsonl", and leave the body code as is with
```
doc_id = data["_id"]
text = data["title"] + " " + data["text"]
```

JJGreen0:
We could probably do as @lilyjge suggested and remove the corpus munging and revert the changes to the body code. For now, I added a note saying it isn't strictly necessary. And, a note about setting the permissions of the token.

PR #2107, opened 2025-05-05
Add script for tokenizing corpus and query files with Lucene tokenizer

lilyjge:
This script takes a corpus file and a query file in jsonl format and tokenizes their contents with Lucene's default analyzer, outputting results at specified file paths.


lintool (approved):
This is okay for now, but we'll need to refactor when generalizing to other corpora.

PR #2114, opened 2025-05-16
Onboarding - Add reproduction log

luisvenezian:
Setup

- Macbook Pro M3
- 16GB RAM
- macOS Sonoma
- Java 21.0.7

Other things in PR

I removed $ from Python calls to match file style.

### DIFF
--- a/docs/experiments-msmarco-passage.md
+++ b/docs/experiments-msmarco-passage.md
@@ -133,7 +133,7 @@ For example, setting `--threads 16 --batch-size 64` on a CPU with sufficient cor
 After the run finishes, we can evaluate the results using the official MS MARCO evaluation script, which has been incorporated into Pyserini:
 
 ```bash
-python -m pyserini.eval.msmarco_passage_eval \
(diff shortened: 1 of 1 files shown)


docs/experiments-msmarco-passage.md
```diff
@@ -133,7 +133,7 @@ For example, setting `--threads 16 --batch-size 64` on a CPU with sufficient cor
 After the run finishes, we can evaluate the results using the official MS MARCO evaluation script, which has been incorporated into Pyserini:
 
 ```bash
-$ python -m pyserini.eval.msmarco_passage_eval \
```
lintool, line 136: Please remove `$` because we don't want it picked up in copy/paste.

luisvenezian: Thanks @lintool , that makes sense! reverted in [b24f988](https://github.com/castorini/pyserini/pull/2114/commits/b24f9887987794b40e9ae60d0a1e6d7a56cbe086)

PR #2120, opened 2025-05-20
Sort as string if not all the same type

luisvenezian:
This PR aims to document what I did to address the situation described in [discussion #2119](https://github.com/castorini/pyserini/discussions/2119#discussion-8345321) where the code wasn't handling certain edge cases. This caused the process to break instead of continuing execution.

```python
def safe_sort(lst):
    try:
        return sorted(lst)
    except:
        return sorted(lst, key=str)

# tests
assert safe_sort([3, 2, 1]) == [1, 2, 3]
assert safe_sort(["A", "C", "B"]) == ["A", "B", "C"]
assert safe_sort([1, 2, "A", "B", 3, "C"]) == [1, 2, 3, "A", "B", "C"]
```

### DIFF
--- a/pyserini/query_iterator.py
+++ b/pyserini/query_iterator.py
@@ -41,7 +41,16 @@ class QueryIterator(ABC):
                         'msmarco-passage-test-subset'}
 
     def __init__(self, topics: dict, order: list = None):
-        self.order = order if order else sorted(topics.keys())
+        if order:
+            self.order = order
+        else:
+            try:
+                self.order = sorted(topics.keys())
+            except:
+                # if not all topic ids are integers neither are all string,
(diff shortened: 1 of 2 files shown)


lintool:
hi @luisvenezian can you please add a test case (e.g., from the discussion thread)?

luisvenezian:
```
$ python3 -m unittest test_topics_order.py

.Using pre-defined topic order for msmarco-passage-dev-subset
....
----------------------------------------------------------------------
Ran 5 tests in 0.257s

OK
```

lintool (approved):

lintool:
Thanks for the contribution!

PR #2134, opened 2025-05-30
Add to reproduction log and code changes for experiments-vector-store…

JJGreen0:
…s.md

I ran into issues with the collection already existing, and when I tried to reuse the collection I created before, the issue of duplicating my results became evident. Thus, ensuring to "start fresh" on a new collection seemed most intuitive to me.

I was lulled into a false sense of security by the first two tqdm bars, so I thought we should have a third.


lilyjge:
If it's just one run through then the collection existing wouldn't happen, but it doesn't hurt to check before hand. Always nice to have a tqdm bar.

lilyjge (approved):

Issue #2152, opened 2025-06-11
REST API feature parity with Anserini


lintool:
@sisixili says that REST API in Pyserini is still missing some of the shard stuff from Anserini. We should implement those.

PR #2158, opened 2025-06-20
sharded search for msmarco v2.1

sisixili:
Initial working sharded search, hardcoded for msmarco v2.1 arctic embed l. 

#2152


pyserini/server/search_controller.py
```diff
@@ -108,17 +115,45 @@ def search(
         results["candidates"] = candidates
 
         return results
-
-    def get_document(self, docid: str, index_name: str) -> Optional[Dict[str, Any]]:
+    
+    def sharded_search( # temporarily hardcode msmarco v2.1 arctic embed l shards
```
ronakice: modify to a todo

pyserini/server/search_controller.py
```diff
@@ -108,17 +115,45 @@ def search(
...
+        self,
+        query: str,
+        k: int,
+        ef_search: int,
+        encoder: str,
+    ) -> list[dict[str, float]]:   
+                
+        executor = ThreadPoolExecutor(max_workers=10)
```
ronakice: why is the max_wrokers here hard coded?

pyserini/server/search_controller.py
```diff
@@ -20,81 +20,88 @@
...
         
         if default_index in TF_INDEX_INFO.keys():
             self.add_index(
-                IndexConfig(
-                    name=default_index,
-                    type=IndexType.PREBUILT,
-                    path=default_index,
-                    description=TF_INDEX_INFO[default_index].get("description", ""),
-                )
+                IndexConfig(name=default_index)
```
ronakice, line 52: Why was this refactored

pyserini/server/search_controller.py
sisixili: It's hardcoded for the 10 msmarco v2.1 shards

pyserini/server/search_controller.py
sisixili: path isn't needed since prebuilt indexes go just by name, and I think it makes more sense to not save description since it's not used anywhere and is pretty easy to get

PR #2161, opened 2025-06-23
Add metadata loading from Anserini

lilyjge:
The metadata removed from Pyserini in #2159 has been added to Anserini as of castorini/anserini#2853 as part of .castorini/anserini#2852. This PR makes it so the additional metadata actually gets loaded into Pyserini from Anserini.


lilyjge:
Can close castorini/anserini#2852 once this is merged.

lintool (approved):

PR #2173, opened 2025-07-07
Add bindings for BRIGHT prebuilt indexes

lilyjge:


lintool (approved):

lintool:
hi @lilyjge do you mind if we revert this patch for now? Rationale is that I want to publish v1.2.0 after @sisixili 's patch #2174 - and this needs a new anserini artifact.

Issue #2176, opened 2025-07-07
Update REST API documentation


lintool:
https://github.com/castorini/pyserini/blob/master/docs/experiments-msmarco-v2.1.md#rest-api-and-webapp
https://github.com/castorini/pyserini/blob/master/docs/usage-rest.md

lilyjge:
#2191

lilyjge:
What about https://github.com/castorini/pyserini/blob/master/docs/experiments-msmarco-v2.1.md#rest-api-and-webapp? Should we just rip the section out?

lintool:
good catch @lilyjge - that entire page needs an update. I'll work on that.

PR #2178, opened 2025-07-08
update naming

sisixili:
(#2176)

### DIFF
--- a/docs/experiments-msmarco-v2.1.md
+++ b/docs/experiments-msmarco-v2.1.md
@@ -62,16 +62,16 @@ print(json.dumps(json.loads(doc.raw()), indent=2))
 
 ## REST API and Webapp
 
-Pyserini provides a REST API for programmatic access (in truth, it's just a wrapper around a Java application in Anserini):
+Pyserini provides a REST API for programmatic access:
 
 ```bash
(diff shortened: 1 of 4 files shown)


docs/experiments-msmarco-v2.1.md
```diff
@@ -62,16 +62,16 @@ print(json.dumps(json.loads(doc.raw()), indent=2))
 
 ## REST API and Webapp
 
-Pyserini provides a REST API for programmatic access (in truth, it's just a wrapper around a Java application in Anserini):
+Pyserini provides a REST API for programmatic access:
 
 ```bash
-python -m pyserini.server.AnseriniApplication --server.port=8082
+python -m pyserini.server.rest --port 8082
```
lintool, line 76: leave default port, but add documentation on how to specify alternative?

lintool (changes requested):
`/{index}/search/settings` is currently dense only settings right?

+ what about sparse settings?
+ should we add `hits` also?

docs/experiments-msmarco-v2.1.md
sisixili: wdym? Say how to specify port like in usage-rest docs?

lintool: yes

PR #2193, opened 2025-07-16
Update search_controller.py to support MS MARCO v2.1 segment field

rankun203:
The MS MARCO v2.1 segment corpus is using `segment` as the main text field.

I find the REST search API is trying to always return the main text on `doc` field, instead of returning whatever the index has on that record. Is the REST search API trying to serve other statically typed downstream tools, instead of being a generic search API?

### DIFF
--- a/pyserini/server/search_controller.py
+++ b/pyserini/server/search_controller.py
@@ -124,7 +124,7 @@ class SearchController:
                 doc = json.loads(hit.lucene_document.get('raw'))
             elif index_config.base_index:
                 doc = self.get_document(hit.docid, index_config.base_index)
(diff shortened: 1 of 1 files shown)


sisixili:
> This is a good update for the existing architecture. @sisixili do you remember why we made it return the main text instead of the the whole raw document? I recall talking about this and having a reason. Now I'm thinking we return main text with search but the whole raw document with `get_document`. wdyt?

Yeah that sounds good. I don't remember thinking about this...

sisixili (approved):

lilyjge (approved):

PR #2195, opened 2025-07-16
Add tests for REST API

lilyjge:


pyserini/server/rest/routes/indexes.py
```diff
@@ -80,7 +79,7 @@ async def update_index_settings(
     params: IndexSettingParams = Depends()
 ) -> dict[str, str]:
     try:
-        get_controller().update_settings(index, params.ef_search, params.encoder, params.query_generator)
+        get_controller().update_settings(index, params.efSearch, params.encoder, params.queryGenerator)
```
lintool: hrm... this is Python, so we should stick to `snake_case`?

tests/test_rest.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_get_index_settings(self):
+        response = client.get("v1/indexes/msmarco-v1-passage/settings")
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue("efSearch" in list(response.json().keys()))
+        self.assertTrue("encoder" in list(response.json().keys()))
+        self.assertTrue("queryGenerator" in list(response.json().keys()))
+
+    def test_post_index_settings(self):
+        data = {"efSearch": "100", "encoder": "BgeBaseEn15Encoder", "queryGenerator": "BagOfWordsGenerator"}
```
lintool, line 64: oh interesting... do we want JSON params to be `snake_case` or `smallCaps`... 🤔 

can you search on the interwebs and see if there's best practice?

lintool (approved):

Issue #2086, opened 2025-03-02
pyserini sparse vector encoding


clides:
right now pyserini.encode only encodes documents into dense vector representations, I will try to implement the script used in conceptual-framework3.md on the onboarding path into pyserini so there is an extra option for the command to encode documents into sparse vectors

clides:
I didn't get a chance to work on this yet, I'll reopen the issue once I finish other tasks

PR #2199, opened 2025-07-17
Add option for splade sparse corpus encoding with commands

clides:
Ref: https://github.com/castorini/pyserini/issues/2086

Example usage:
```bash
python -m pyserini.encode \
  input --corpus collections/nfcorpus/corpus.jsonl \
        --fields title text \
  output --embeddings encode/nfcorpus.splade-v3 \
  encoder --encoder naver/splade-v3 \
          --encoder-class splade \
          --fields title text \
          --max-length 512
```

### DIFF
--- a/docs/conceptual-framework3.md
+++ b/docs/conceptual-framework3.md
@@ -74,7 +74,7 @@ You’ll be prompted to enter your Hugging Face API token. You can generate a to
 
 We are now all set to use SPLADE-v3 model!
 
-Start by running the following pyserini command to encode the corpus into sparse vector representations.
+Start by running the following command to encode the corpus into sparse vector representations.
 
 ```bash
 python -m pyserini.encode \


lintool:
@lilyjge wdyt? IIRC you mentioned having some issues with cached queries on BRIGHT w/ SPLADE-v3?

lilyjge:
Update - The score difference between cached queries and Pyserini encoded queries is due to a difference in max length:
- Cached queries were encoded with max length of 512. Anserini's sparse ONNX encoding also uses max length of 512. 
- Pyserini's SpladeV3 encoder (this pr) uses max length of 256. 

512 seems to be the correct option, per https://huggingface.co/naver/splade-v3
Could you please fix @clides and we can merge this? It's needed for the BRIGHT 2CR #2234

lintool:
> Side not: Interestingly, max length of 256 does slightly better on average on BRIGHT...

@lilyjge probably noise...

@clides is there some lightweight way we can test this? it'd be ideal to have a test...

lilyjge:
yep, lgtm!

docs/conceptual-framework3.md
```diff
@@ -75,64 +74,26 @@ You’ll be prompted to enter your Hugging Face API token. You can generate a to
 
 We are now all set to use SPLADE-v3 model!
 
-Start by running the following script to encode the corpus into sparse vector representations.
-(we are using this custom script instead of pyserini.encode as pyserini.encode only encodes corpus into dense vectors)
+Start by running the following pyserini command to encode the corpus into sparse vector representations.
```
lintool:
```suggestion
Start by running the following command to encode the corpus into sparse vector representations.
```

Issue #2203, opened 2025-07-19
mcp test failure


lintool:
Hi @sisixili @lilyjge - I'm getting a test failure on mcp... can you repro?

```
$ pip list | grep mcp
mcp                       1.12.0

$ python -m unittest tests/test_mcp.py 
FEFEProcess terminated gracefully
F
======================================================================
ERROR: test_get_index_status_tool (tests.test_mcp.TestMCPyseriniServer)
----------------------------------------------------------------------
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 43, in tearDown
    self.server_process.stdin.close()
BrokenPipeError: [Errno 32] Broken pipe

======================================================================
ERROR: test_search_tool (tests.test_mcp.TestMCPyseriniServer)
----------------------------------------------------------------------
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 43, in tearDown
    self.server_process.stdin.close()
BrokenPipeError: [Errno 32] Broken pipe

======================================================================
FAIL: test_get_index_status_tool (tests.test_mcp.TestMCPyseriniServer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 86, in send_mcp_request
    self.server_process.stdin.write(init_json)
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 149, in test_get_index_status_tool
    response = self.send_mcp_request('tools/call', {
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 105, in send_mcp_request
    self.fail(f'Failed to initialize MCP connection: {e}')
AssertionError: Failed to initialize MCP connection: [Errno 32] Broken pipe

======================================================================
FAIL: test_search_tool (tests.test_mcp.TestMCPyseriniServer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/store/scratch/jimmylin/pyserini/tests/test_mcp.py", line 86, in send_mcp_request
[...]

PR #2204, opened 2025-07-19
Update dependencies list for MCP 

lilyjge:
Addresses #2203


lintool:
@sisixili can you verify that this is the correct fix? thanks!

Issue #2150, opened 2025-06-10
MCPyserini functionalities


lintool:
Opening this thread to brainstorm MCPyserini functionalities.

With MCP, we should be able to interact with an LLM to answer these questions, right @sisixili ?

+ What indexes do you have available?
+ Do you have index _x_ downloaded?
+ How big is this index?

Data for all of these should already be exposed via the status endpoint?

sisixili:
Yes, all of these functionalities are in status endpoint.

PR #2206, opened 2025-07-19
Add fusion and qrel capabilities for MCPyserini

lilyjge:
Fusion demonstration: 
image
Qrels demonstration: https://claude.ai/share/583b6469-3a4b-4bb3-9431-9b7952ec697d
Addresses https://github.com/castorini/pyserini/issues/2150#issuecomment-3092334699
A slightly inconvenient thing: qrels mark based on query id, but retrieval uses query text. So to retrieve and evaluate, you need to know both the query content and the query id.

### DIFF
--- a/pyserini/server/mcp/tools.py
+++ b/pyserini/server/mcp/tools.py
@@ -22,7 +22,7 @@ Register tools for the MCP server.
 from typing import Any
 
 from fastmcp import FastMCP
-from pyserini.server.search_controller import SearchController
+from pyserini.server.search_controller import SearchController, DenseSearchResult
 from pyserini.server.models import INDEX_TYPE
 
 def register_tools(mcp: FastMCP, controller: SearchController):
@@ -30,7 +30,7 @@ def register_tools(mcp: FastMCP, controller: SearchController):
 
     @mcp.tool(
         name='search',
-        description='Perform a BM25 search on a given index. Returns top‑k hits with docid, score, and snippet.',
+        description='Perform search on a given index. Returns top‑k hits with docid, score, and snippet.',
     )
     def search(
         query: str,
@@ -99,3 +99,25 @@ def register_tools(mcp: FastMCP, controller: SearchController):
             Dictionary with index information.
         """
         return controller.get_status(index_name)  
+    
+    @mcp.tool(
+        name="fuse_search_results",
+        description="Performs normalization fusion on search results to improve ranking."
+    )
+    def fuse_search_results(
+        hits1: list[DenseSearchResult], 
+        hits2: list[DenseSearchResult], 
+        k: int = 10
+    ) -> list[DenseSearchResult]:
+        """
+        Performs normalization fusion on search results to improve ranking.
+
+        Args:
+            hits1: First list of search results to merge with docid and score
+            hits2: Second list of search results to merge with docid and score
+            k: Number of results to return (default: 10)
+
+        Returns:
+            Dictionary with index information.
+        """
+        return controller.fuse(hits1, hits2)
--- a/pyserini/server/mcp/tools.py
+++ b/pyserini/server/mcp/tools.py
@@ -120,4 +120,21 @@ def register_tools(mcp: FastMCP, controller: SearchController):
         Returns:
             Dictionary with index information.
         """
-        return controller.fuse(hits1, hits2)
+        return controller.fuse(hits1, hits2, k)
+    
+    @mcp.tool(
+        name="get_qrels",
+        description="Returns relevant judgements for a given index and query."
+    )
+    def get_qrels(
+        index_name: str,
+        query_id: str
+    ) -> dict[str, str]:
+        """
+        Returns relevant judgements for a given index and query.
+
+        Args:
+            index_name: Name of the index to get relevant judgements for
+            query_id: Query ID to to get relevant judgements for
+        """
+        return controller.get_query_qrels(index_name, query_id)
--- a/pyserini/server/search_controller.py
+++ b/pyserini/server/search_controller.py
@@ -26,7 +26,8 @@ import json
 from typing import Any
 
 from pyserini.search.lucene import LuceneSearcher, LuceneHnswDenseSearcher, LuceneFlatDenseSearcher, LuceneImpactSearcher
-from pyserini.search.faiss import FaissSearcher
+from pyserini.search.faiss import FaissSearcher, DenseSearchResult
+from pyserini.search.hybrid import HybridSearcher
 from pyserini.encode import AutoQueryEncoder
 from pyserini.prebuilt_index_info import TF_INDEX_INFO, LUCENE_FLAT_INDEX_INFO, LUCENE_HNSW_INDEX_INFO, IMPACT_INDEX_INFO, FAISS_INDEX_INFO
 from pyserini.util import check_downloaded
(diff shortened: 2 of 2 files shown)


lintool:
> https://claude.ai/share/583b6469-3a4b-4bb3-9431-9b7952ec697d

hey @lilyjge is the computation for nDCG actually _right_? Can you cross-check against trec_eval?

Maybe we should call the endpoint to compute scores, rather than trust Claude to do it "in-model"?

lilyjge:
In this case it's right, but I agree that an endpoint would be safer. It's just that `trec_eval` is not a very 'modular' function, so using it to judge one query is going to be a little 'hacky' from what I can see. Possible options:
- writing run results to temp file, call the CLI trec_eval command with -q, and
  -  parse stdout for query id, or
  -  small refactor to pyserini's trec eval to get it to return parsed stdout results 
- alternatively, refactoring anserini trec_eval to provide an API for per query results
I think writing run results to temp file, calling the CLI trec_eval command with -q, and small refactor to pyserini's trec eval to get it to return parsed stdout results is probably the best option? wdyt?

lilyjge:
Demonstration of evaluation: https://claude.ai/share/5608ad11-bcb1-46de-9a17-a957d9661439
Supports ndcg, recall, map, recip_rank. Are there any I missed?
Evaluation is on a per-query basis currently, taking into consideration context length constraints. 

Also updates `get_document` to retrieve raw document, as discussed in [#2193. ](https://github.com/castorini/pyserini/pull/2193#issuecomment-3094631383)

tests/test_rest.py
```diff
@@ -37,9 +37,9 @@ def test_search(self):
     def test_get_doc(self):
         response = client.get("v1/indexes/msmarco-v1-passage/documents/7157707")
```
sisixili: nitpick: can import version # from rest server so this doesn't break on new versions

pyserini/server/search_controller.py
```diff
@@ -136,6 +139,38 @@ def search(
...
+        metric: str,
+        query_id,
+        hits: dict[str, float],
+        cutoff: int = 10
+    ) -> float:
+        hits = sorted(hits.items(), key=lambda item: item[1], reverse=True)
+        temp_file = f"{index_name}-{metric}-{cutoff}-{query_id}.txt"
+        with open(temp_file, "w") as f:
+            for rank, doc in enumerate(hits, start=1):
+                f.write(f"{query_id} Q0 {doc[0]} {rank} {doc[1]} mcp\n")
```
sisixili, line 162: This (and eval) can probably be streamed with StringIO, but considering this already works...

lilyjge: The problem is Pyserini eval outsources to Anserini which needs a file

docs/usage-mcp.md
```diff
@@ -146,10 +146,7 @@ Search for "what is a lobster roll" in the msmarco-v1-passage index, returning 5
 - `docid` (string, required): Document ID to retrieve
 - `index_name` (string, required): Name of the index containing the document
 
-**Returns:** Document object containing:
-- `docid`: Document identifier
-- `contents`: Full document text
-- `raw`: Raw document representation (if available)
+**Returns:** Document object containing the raw document representation.
```
ronakice, line 149: what about the fields?

pyserini/eval/trec_eval.py
```diff
@@ -52,90 +52,100 @@
...
-    os.remove(temp_file)
+    output = output.split("\n")
+    lines = {line.split("\t")[1]: line.split("\t")[2] for line in output}
+    if query_id:
+        return float(lines[query_id])
+    else:
+        return float(lines["all"])
+
+if __name__ == "__main__":
+    eval(sys.argv)
```
ronakice: end with empty line

pyserini/eval/trec_eval.py
```diff
@@ -52,90 +52,100 @@
...
-        with open(args[-1]) as f:
-            first_line = f.readline()
-        if 'Q0' not in first_line:
-            temp_file = tempfile.NamedTemporaryFile(delete=False).name
-            print('msmarco run detected. Converting to trec...')
-            run = pd.read_csv(args[-1], sep='\s+', header=None, names=['query_id', 'doc_id', 'rank'])
-            run['score'] = 1 / run['rank']
-            run.insert(1, 'Q0', 'Q0')
-            run['name'] = 'TEMPRUN'
+def eval(args, query_id=None) -> float:
```
ronakice: why just `eval`? Maybe add some clarity of what function it is actually by saying trec_eval or something

pyserini/server/mcp/tools.py
```diff
@@ -98,4 +98,68 @@ def get_index_status(index_name: str) -> dict[str, Any]:
...
-        return controller.get_status(index_name)  
\ No newline at end of file
+        return controller.get_status(index_name)  
+    
+    @mcp.tool(
+        name="fuse_search_results",
+        description="Performs normalization fusion on search results to improve ranking."
+    )
+    def fuse_search_results(
+        hits1: list[DenseSearchResult],
```
ronakice, line 108: Why `DenseSearchResult`

tests/test_rest.py
```diff
@@ -27,41 +27,41 @@ def test_docs_available(self):
...
         self.assertTrue("efSearch" in list(response.json().keys()))
         self.assertTrue("encoder" in list(response.json().keys()))
         self.assertTrue("queryGenerator" in list(response.json().keys()))
 
     def test_post_index_settings(self):
         data = {"efSearch": "100", "encoder": "BgeBaseEn15Encoder", "queryGenerator": "BagOfWordsGenerator"}
-        response = client.post("v1/indexes/msmarco-v1-passage/settings", json=data)
+        response = client.post(f"/{VERSION}/indexes/msmarco-v1-passage/settings", json=data)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json(), {"status": "success"})
```
ronakice: same

pyserini/server/search_controller.py
```diff
@@ -200,6 +231,10 @@ def get_status(self, index_name: str) -> dict[str, Any]:
         if status.get('size_bytes') is None:
             status['size_bytes'] = "Not available"
         return status
+    
+    def get_query_qrels(self, index_name: str, query_id: str) -> dict[str, str]:
```
ronakice, line 235: is this returning a mapping from document_id to relevance? why dict[str, str]

ronakice (changes requested):
ptal

docs/usage-mcp.md
lilyjge: right now it just returns whatever the raw document is, the fields of which varies

pyserini/server/mcp/tools.py
lilyjge: because that's what's Pyserini's fusion takes and return

pyserini/server/search_controller.py
lilyjge: yes, {doc_key: judgement}

pyserini/server/mcp/tools.py
ronakice: Strange, @lintool naming issue we should make an issue for and eventually return to. I will resolve for speed rn.

pyserini/server/search_controller.py
ronakice: judgment is a `str`?

lilyjge: I believe so? It gets parsed from the qrels file and it doesn't get casted explicitly anywhere

ronakice (approved):
LG

PR #2213, opened 2025-08-04
integrated uniir's encoding for pyserini

clides:
Integrated UniIR's encoding/embedding pipeline into pyserini

Command to run encoding with sample M-BEIR dataset:
```bash
python -m pyserini.encode \
  input --corpus collections/m-beir/WebQA2/mbeir_webqa_task2_cand_pool.jsonl \
        --fields img_path modality txt did \
        --docid-field did \
  output --embeddings encode/m-beir-webqa2-test.clipsf  \
  encoder --encoder clip_sf_large \
          --encoder-class uniir \
          --device cuda:0 \
          --multimodal \
          --fp16 \
          --fields img_path modality txt did
```

Further PRs (UniIR searching and documentation) will be sent after this one has been approved and merged.

### DIFF
--- a/pyserini/encode/_base.py
+++ b/pyserini/encode/_base.py
@@ -215,7 +215,7 @@ class JsonlRepresentationWriter(RepresentationWriter):
 
     def write(self, batch_info, fields=None):
         for i in range(len(batch_info['id'])):
-            contents = "\n".join([batch_info[key][i] if batch_info[key][i] else "" for key in fields])
+            contents = "\n".join([batch_info[key][i] if batch_info[key][i] is not None else "" for key in fields])
             vector = batch_info['vector'][i]
             vector = vector.tolist() if isinstance(vector, np.ndarray) else vector
             self.file.write(json.dumps({'id': batch_info['id'][i],


pyserini/encode/_base.py
```diff
@@ -126,10 +126,16 @@ def _parse_fields_from_info(self, info):
...
-        if all([field in info for field in self.fields]):
-            return [info[field].strip() for field in self.fields]
-
+        direct_fields = [
+            str(info[field]).strip() 
+            if info.get(field, None) is not None 
+            else None 
+            for field in self.fields
+        ]
+        if any(f is not None for f in direct_fields):
```
sahel-sh: all?

pyserini/encode/__main__.py
```diff
@@ -48,12 +48,16 @@ def init_encoder(encoder, encoder_class, device, pooling, l2_norm, prefix, multi
...
     if _encoder_class == 'sentence-transformers' or 'sentence-transformers' in encoder:
         kwargs.update(dict(pooling='mean', l2_norm=True))
-    if _encoder_class == 'contriever' or 'contriever' in encoder:
+    elif _encoder_class == 'contriever' or 'contriever' in encoder:
         kwargs.update(dict(pooling='mean', l2_norm=False))
-    if _encoder_class == 'auto':
+    elif _encoder_class == 'auto':
         kwargs.update(dict(pooling=pooling, l2_norm=l2_norm, prefix=prefix))
-    if _encoder_class == 'clip' or 'clip' in encoder:
+    elif _encoder_class == 'uniir':
```
sahel-sh: why not multimodal?

pyserini/encode/__main__.py
```diff
@@ -48,12 +48,16 @@ def init_encoder(encoder, encoder_class, device, pooling, l2_norm, prefix, multi
...
+        kwargs.update(dict(l2_norm=True))
+    elif _encoder_class == 'clip' or 'clip' in encoder:
         kwargs.update(dict(l2_norm=True, prefix=prefix, multimodal=multimodal))
+    else:
```
sahel-sh: shouldn't this be unreached?

pyserini/encode/_base.py
sahel-sh: This is not the same behavior as before.

pyserini/encode/_base.py
```diff
@@ -174,7 +180,7 @@ def _load(self, collection_path):
                     for i in range(len(fields_info)):
                         if 'path' in self.fields[i]:
                             _info = fields_info[i]
-                            if not _info.startswith(("http://", "https://")):
+                            if _info and not _info.startswith(("http://", "https://")):
```
sahel-sh: This shouldn't be none, it can be none after your change to all vs any

pyserini/encode/_base.py
```diff
@@ -207,7 +213,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     def write(self, batch_info, fields=None):
         for i in range(len(batch_info['id'])):
-            contents = "\n".join([batch_info[key][i] for key in fields])
+            contents = "\n".join([batch_info[key][i] if batch_info[key][i] is not None else "" for key in fields])
```
sahel-sh: ditto

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+            img = None
+        else:
+            img = Image.open(img_path).convert("RGB")
+            img = self.img_preprocess_fn(img)
+
+        did = entry.get("did", None)
+        did = hash_did(did)
+        cand_txt = entry.get("txt", "")
+        cand_txt = format_string(cand_txt)
+        cand_modality = entry.get("modality", None)
```
sahel-sh: shouldn't this always be available? why do you default to None?

It can also be inferred from whether or not imp_path and txt are present

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+from PIL import Image
+from torch.utils.data import DataLoader, Dataset
+
+from pyserini.uniir import (BLIPFeatureFusion, BLIPScoreFusion,
+                            CLIPFeatureFusion, CLIPScoreFusion,
+                            MBEIRCandidatePoolCollator, generate_embeds_and_ids_for_dataset_with_gather,
+                            format_string, hash_did)
+
+
+class CustomCorpusDataset(Dataset):
```
sahel-sh: Please decouple the dataset and the encoders so that other retrievers/encoders for M-BEIR can easily onboard without being tied to UnIR

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+
+class UniIRCorpusEncoder(UniIREncoder):
+    def __init__(self, model_name: str, device="cuda:0", l2_norm=False, **kwargs: Any):
+        super().__init__(model_name, device, l2_norm, **kwargs)
+
+    def encode(
+        self,
+        dids: List[int],
+        img_paths: Optional[List[str]] = None,
+        modalitys: Optional[List[str]] = None,
```
sahel-sh: Nit: modalities

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+        if not model_key:
+            raise ValueError(f"Unsupported model name for UniIR: {model_name}")
+
+        ModelClass, model_dir = MODEL_REGISTRY[model_key]
+        if "clip" in model_name:
+            model = ModelClass(model_name=clip_vision_model, device=device)
+        elif "blip" in model_name:
+            from pyserini.uniir import MED_CONFIG_PATH
+
+            config = SimpleNamespace()
```
sahel-sh: please use a config file similar to uniir, rather than hardcoding these all

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+
+            config = SimpleNamespace()  
+            config.tokenizer_max_length = 100
+            config.alpha = 0.4
+            config.embed_dim = 768
+            config.image_size = 224
+
+            model = ModelClass(
+                med_config=MED_CONFIG_PATH,
+                vit="large" if "large" in model_name else "base",
```
sahel-sh: ditto

pyserini/encode/_uniir.py
```diff
@@ -0,0 +1,197 @@
...
+
+        corpus_embeddings, _ = generate_embeds_and_ids_for_dataset_with_gather(  
+            self.model,  
+            dataloader,  
+            device=self.device,  
+            use_fp16=use_fp16,  
+        )  
+
+        if self.l2_norm:
+            corpus_embeddings = corpus_embeddings.astype('float32')
```
sahel-sh, line 134: ft16 is used everywhere else, why 32 here?

pyserini/encode/__main__.py
sahel-sh: The read me looks like --l2_norm is passed, but in line 56 it is hard coded to true?

pyserini/encode/_uniir.py
clides: Faiss's L2 Norm need it to be in float32, before adding this, I was getting errors

pyserini/encode/_base.py
clides: Yes, I added this because I was running into errors when encoding the global union dataset, some entries in the corpus file are missing keys like img_path, which caused the code to go to the other condition and errored because of the assert. I added this so that if at least 1 field is not None, then use the fields user provided. Let me know what you think is best to handle this.

pyserini/encode/__main__.py
clides: For the l2_norm, it is hardcoded (since UniIR's embeddings are L2 normed before indexing) and I saw the other model_classes and they also have hardcoded values so I added it here as well.

For multimodal, I could pass the key in but I don't think it will be used anywhere since the integration directly uses UniIR's function

pyserini/encode/__main__.py
clides: Looking at the encoder_class choices, I think if you choose an encoder-class like "dpr", "bpr", "tct_colbert", "ance", "openai-api", or "arctic", none of the if or elif conditions in init_encoder will be met, so the else block can be reached?

pyserini/encode/_base.py
clides: This is the same issue as the all vs any comment above, when I was encoding the global union corpus, some entries are missing img_path and this code errored out. Let me know what you think is best to handle this.

pyserini/encode/_uniir.py
clides: Fixed, it is present in every dataset so I think its cleaner to just retrieve it

pyserini/encode/_uniir.py
clides: I don't think this can be fixed because of the way it is set:

```python
            for field_name in args.encoder.fields:
                kwargs[f'{field_name}s'] = batch_info[field_name] 
```

pyserini/encode/_uniir.py
clides: Fixed, created new folder mbeir inside pyserini/encode and added mbeir dataset and the uniir submodule there

pyserini/encode/_uniir.py
clides: added config file for both clip and blip models

pyserini/encode/_base.py
sahel-sh: This will cause errors for the rest. Handle uniir missing fields properly when you are reading/loading the data

pyserini/encode/__main__.py
sahel-sh: If it is hard coded here based on the model class, why do you pass it? (implies that when I don't pass --l2_norm then it is false by default which clearly is not).
Maybe you only need to remove the --l2_norm and --multimodal args if they are not used?

pyserini/encode/__main__.py
sahel-sh: If you think this is a bug, fix it in a separate PR and @lilyjge, @lintool  or someone else more familiar with the code could review it. I'd like this PR to be an absolute no-op for anything else already integrated into Pyserini

pyserini/encode/_base.py
sahel-sh: handle it at the time of reading the data, for uniir specifically

pyserini/encode/_uniir.py
sahel-sh: I see...
l2_norm is hard coded to be always true, right?
I guess you can cast back to float16, but looking at the uniir embedder code, I don't see any l2_norm, why do you use it here?
https://github.com/TIGER-AI-Lab/UniIR/blob/main/src/common/mbeir_embedder.py

pyserini/encode/mbeir/mbeir_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+        self.img_preprocess_fn = img_preprocess_fn
+        self.kwargs = kwargs
+
+    def __len__(self):
+        return len(self.data)
+
+    def __getitem__(self, idx):
+        entry = self.data[idx]
+        img_path = entry.get("img_path", None)
+        if img_path is not None:
```
sahel-sh: isn't this enough to take care img_path None?
It is not clear to me why you need to make the changes to the fields and info handling logic

pyserini/encode/_uniir.py
clides: It is normalized right before indexing, but it saves the raw embeddings

This is from mbeir_retriever.py
```python
# Load the embeddings and IDs from the .npy files
        embedding_list = np.load(embed_data_path).astype("float32")
        hashed_id_list = np.load(embed_data_hashed_id_path)

        # Check unique ids
        assert len(hashed_id_list) == len(set(hashed_id_list)), "IDs should be unique"

        # Normalize the embeddings
        faiss.normalize_L2(embedding_list)
```

pyserini/encode/__main__.py
clides: Ok, I've reverted the conditions to the original, will file another PR later

pyserini/encode/_base.py
```diff
@@ -150,6 +153,11 @@ def _load(self, collection_path):
         else:
             for filename in os.listdir(collection_path):
                 filenames.append(os.path.join(collection_path, filename))
+
```
sahel-sh, line 156: NIT: Please add a comment about M-BEIR having missing fields

pyserini/encode/_base.py
```diff
@@ -207,7 +214,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     def write(self, batch_info, fields=None):
         for i in range(len(batch_info['id'])):
-            contents = "\n".join([batch_info[key][i] if batch_info[key][i] is not None else "" for key in fields])
+            contents = "\n".join([batch_info[key][i] for key in fields])
```
sahel-sh: key can exist but with value of None, why are removing the value check to replace None with ""? This will impact all the other integrations

pyserini/encode/mbeir/mbeir_dataset.py
sahel-sh: you can change this to `if img_path:`, so that "" is also treated as false.

pyserini/encode/_base.py
sahel-sh: New strict boolean takes care of this

pyserini/encode/_base.py
sahel-sh: this is fixed now with the strict boolean flag

pyserini/encode/_uniir.py
sahel-sh: use of config file fixed this

pyserini/encode/_uniir.py
sahel-sh: Got it, thanks for the pointer. 
The save it raw, but normalize it at inference time.
I agree that we can save it normalized but please make sure you cast it back to ft16, currently you will be saving it as ft32 after the faisss normalization

pyserini/encode/_base.py
clides: Fixed

pyserini/encode/_base.py
clides: Oops, rebase issue, fixed

sahel-sh:
Please remove `--multimodal` from the command in the PR description, since you are not passing this to the uniir encoder

pyserini/encode/mbeir/mbeir_dataset.py
clides: fixed

sahel-sh (approved):

PR #2218, opened 2025-08-12
Add fp16 option for faiss searching

clides:


pyserini/search/faiss/__main__.py
```diff
@@ -396,6 +398,7 @@ def init_query_encoder(
         default=1,
         help="maximum threads to use during search",
     )
+    parser.add_argument('--fp16', action='store_true', default=False)
```
sahel-sh: NIT:please add a help 
also a comment saying that this is used for UniIR models

clides: Done

sahel-sh (approved):
LGTM

Issue #2246, opened 2025-09-17
Uniir imports shouldn't be in pyserini/encode/__init__.py


lintool:
Since uniir is an optional dependency, imports should not be in `pyserini/encode/__init__.py`:

```
from ._uniir import UniIRCorpusEncoder, UniIRQueryEncoder
```
https://github.com/castorini/pyserini/blob/master/pyserini/encode/__init__.py#L37C1-L38C1

This causes test failures when I don't have dependencies installed.

Similarly, `pyserini/encode/_uniir.py` should be in `pyserini/encode/optional`?

PR #2250, opened 2025-09-17
Fix Uniir import and move uniir to optional

clides:
Fixing Uniir encoder's import errors when the uniir-for-pyserini package is not installed. Moving uniir to optional dependency and fixing readme accordingly.


pyserini/encode/__init__.py
```diff
@@ -34,7 +34,11 @@
 from ._tct_colbert import TctColBertDocumentEncoder, TctColBertQueryEncoder
 from ._tok_freq import TokFreqQueryEncoder
 from ._unicoil import UniCoilEncoder, UniCoilDocumentEncoder, UniCoilQueryEncoder
-from ._uniir import UniIRCorpusEncoder, UniIRQueryEncoder
+try:
```
lintool, line 37: Just wondering if this is best practice, or let the user _explicitly_ do the import?

@sahel-sh wdyt? you make the call please.

sahel-sh: This is what we use in rankllm too.
@clides the caller/user if this classes should check for none and print proper messages for installation of uniir. PTAL at pyserini import in rankllm as a reference

lintool: sgtm

sahel-sh: or maybe add them to the maps only conditionally:
if UniIRCorpusEncoder:
    document_encoder_class_map["uniir"] = UniIRCorpusEncoder

lintool: I'm okay either way.

clides: @sahel-sh I've added error messages for installation of uniir for both corpus and query encoders

lintool:
@clides in the future, please provide at least a short description to your PRs? In this case - a link to #2246 would be helpful.

sahel-sh (approved):
LGTM, thanks

PR #2256, opened 2025-09-20
Added tests for UniIR encoders

clides:
Added tests for UniIR corpus and query encoders to test if the UniIR encoders within pyserini can initialize successfully. The tests will run successfully if uniir-for-pyserini is installed and will error otherwise.


tests-optional/test_encoder_uniir.py
```diff
@@ -0,0 +1,23 @@
+import unittest
```
lintool, line 17: Please add the standard header etc.

clides: Added

lintool (approved):

PR #2263, opened 2025-09-24
Refactor tests into tests/core and tests/optional

lintool:
Previously, in separate directories... now, moved into same directory.

wdyt @sahel-sh ?


tests/core/test_search.py
```diff
@@ -54,7 +54,7 @@ def setUpClass(cls):
         # The current directory depends on if you're running inside an IDE or from command line.
         curdir = os.getcwd()
         if curdir.endswith('tests'):
-            corpus_path = '../tests/resources/sample_collection_json'
+            corpus_path = '../resources/sample_collection_json'
```
sahel-sh, line 57: why this change?

lintool:
@sahel-sh I refactored and fixed path issues - to make sure tests ran both in IDE and also command-line.

PR #2276, opened 2025-09-29
doc: correct how to install latest package

alanyee:


lintool:
Hi @alanyee thanks for your contribution, but explicitly specifying the version IMO is less bug prone.

PR #2278, opened 2025-10-01
Refactor integrations into integrations/core and integrations/optional

lintool:
Running the integrations right now. Not ready for merging yet.


lintool:
@sahel-sh @lilyjge ready for review!

integrations/optional/dense/test_tct_colbert-v2.py
```diff
@@ -20,7 +20,7 @@
 import os
 import unittest
 
-from integrations.utils import clean_files, run_command, parse_score
+from integrations.core.utils import clean_files, run_command, parse_score
```
sahel-sh: if utils in being used in both core and optional, it is better to bring it up under integrations

lintool: @sahel-sh do you mean move `utils` from `integration.core.utils` to just `integration.utils`?

sahel-sh (approved):
LGTM, thanks!

PR #2305, opened 2025-10-22
reproduction log for pyserini

RichHene:
I was able to complete the tutorial, but had some issues with JAVA_HOME. The issue still persisted, and wasn't able to see the results produced by the end of the tutorial, but it is well understood as the Anserini tutorial


lintool:
Why are your diffs so large? They should only include one additional line in each file.

RichHene:
> Why are your diffs so large? They should only include one additional line in each file.

It looks like my change unintentionally copied the entire markdown file instead of just adding the extra line. I’ll fix that so only the intended change remains.

I’m still getting used to GitHub

lintool:
> Hello, I’d prefer to leave the file as is for now to avoid causing any further issues. Please feel free to make any necessary adjustments on your end.

Sorry, it doesn't work like that... there's not "any adjustments" I can make "on my end". You'll have to figure out how to make a proper GitHub PR.

lintool:
Having seen no further progress, closing.

PR #2308, opened 2025-10-24
py-repro-changes

RichHene:
repro changes made


lintool:
Ref #2305 - please collapse into single PR and fix conflicts.

PR #2327, opened 2025-11-02
Update mbeir 2cr

clides:
Update mbeir 2cr to use ndcg@10, R@20, R@100 across all datasets. Also updated 2cr commands to use instructions as they give higher eval scores.

### DIFF
--- a/docs/2cr/m-beir.html
+++ b/docs/2cr/m-beir.html
@@ -1091,9 +1091,9 @@ python -m pyserini.eval.trec_eval \
 <td>0.7769</td>  
 <td>0.9792</td>  
 <td></td>  
-<td>0.1978</td>  
-<td>0.5618</td>  
-<td>0.7094</td>  
+<td>0.2724</td>  
+<td>0.7840</td>  
+<td>0.9792</td>  
 </tr>  
 <tr class="hide-table-padding">  
 <td></td>  


lintool:
hi @clides thanks for sending this. I'm in the middle of running a bunch of tests to validate the PyPI artifact. Let me wait until those finish before merging this?

pyserini/2cr/m_beir.py
```diff
@@ -65,7 +46,7 @@ def format_run_command(raw):
         .replace("--output-format", "\\\n  --output-format")
         .replace("--output ", "\\\n  --output ")
         .replace("--hits ", "\\\n  --hits ")
-        .replace("--remove-query", "\\\n  --remove-query")
```
sahel-sh, line 68: this remove query should stay, why delete it?

pyserini/2cr/m_beir.py
```diff
@@ -216,17 +189,17 @@ def generate_report(args):
                 
             for datasets in condition['datasets']:    
                 dataset = datasets['dataset']    
+                instruction_config = f"{dataset}_instr.yaml"
```
sahel-sh, line 190: IIRC some datasets are repeated in different tasks with different instructions, is dataset enough as a key? or is task number also included in dataset?

clides: Task number is already included in the dataset

pyserini/2cr/m_beir.py
clides: Fixed

lintool:
I'm specifically talking about the following in this PR:

```
--instruction-config collections/m-beir/query_instructions/${instruction_config}
```

How big are the instructions? Just include them in our codebase directly?

PR #2337, opened 2025-11-06
Refactor pulling in of prebuilt index symbols from Anserini

lintool:
In preparation for next Anserini release: refactoring how we pull in prebuilt index symbols from the Anserini fatjar.


lintool:
Hi @lilyjge I think this is ready for review - I unjanked code from pulling in prebuilt index symbols from Anserini.

lilyjge (approved):

PR #2347, opened 2025-11-20
BrowseComp-Plus Pre-Built BM25 and Qwen3-Embed-8B Indexes 

wu-ming233:
This PR connects the pre-built indexes for [BrowseComp-Plus](https://github.com/texttron/BrowseComp-Plus) to Pyserini, currently with BM25 and Qwen3-Embed-8B indexes, hosted on [hgf](https://huggingface.co/datasets/castorini/prebuilt-indexes-browsecomp-plus) (to be added to [our collection](https://huggingface.co/collections/castorini/prebuilt-indexes) later).

### DIFF
--- a/pyserini/encode/__main__.py
+++ b/pyserini/encode/__main__.py
@@ -24,7 +24,7 @@ from pyserini.encode import JsonlRepresentationWriter, JsonlCollectionIterator
 from pyserini.encode.optional import FaissRepresentationWriter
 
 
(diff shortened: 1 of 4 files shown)


wu-ming233:
I noticed that pyserini currently doesn't support eos pooling in faiss. Added it, along with being able to specify padding_side and fp16. These are used to inference with the qwen3-embedding-8b model

PR #2348, opened 2025-11-20
Increase min version for uniir-for-pyserini

sahel-sh:
The min version should include the fix landed in https://github.com/castorini/UniIR-for-Pyserini/pull/12


clides (approved):
LGTM

lintool (approved):

PR #2373, opened 2025-12-29
Add reproduction log for msmarco-passage BM25 baseline

claraashraf24:


lintool:
Please continue along the onboarding path until you reach the end of the exercises, and then we'll take a look all at once.

Also, update the PR with some information about your reproduction efforts.

PR #2378, opened 2025-12-31
Added support for MM datasets for MCP search tool

clides:
Added support for MM datasets for MCP search tool (currently just support UniIR encoders)

Sample testing message I've used:
```
Search for "Show three bottles of soft drink." with image path as "/store2/scratch/d6guo/encode/pyserini/collections/m-beir/mbeir_images/cirr_images/dev/dev-244-0-img0.jpg" and the instruction config as "/store2/scratch/d6guo/encode/pyserini/collections/m-beir/query_instructions/cirr_task7_instr.yaml" for the m-beir-cirr_task7.clip-sf-large index, returning 5 results. Use "clip_sf_large" as the encoder
```

### DIFF
--- a/docs/usage-mcp.md
+++ b/docs/usage-mcp.md
@@ -120,14 +120,17 @@ The Pyserini MCP server provides two main tools for information retrieval:
 **Description:** Perform a BM25 search on a given index and return top-k hits with document ID, score, and text snippet.
 
 **Parameters:**
-- `query` (string, required): Search query string
+- `query` (dict, required): Search query string
 - `index_name` (string, required): Name of the index to search
 - `k` (integer, optional): Number of results to return (default: 10)
+- `instruction_config` (string, optional): Local path to the instruction config file for UniIR or other MM encoders
+- `encoder` (string, optional): Name of the encoder model
 
 **Returns:** List of search results, each containing:
 - `docid`: Document identifier
 - `score`: BM25 relevance score
-- `contents`: Text snippet from the document
+- `query_text`: Text snippet from the document
+- `query_image`: Image snippet from the document (if applicable)
 - `index_name`: Name of the index searched
 
 **Example Usage in MCP Client:**
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -49,7 +49,7 @@ curl "http://localhost:8081/v1/indexes/msmarco-v1-passage/search?query=what%20is
     {
       "docid": "7157707",
       "score": 11.0082998275757,
-      "doc": "Cookbook: Lobster roll Media: Lobster roll A lobster-salad style roll from The Lobster Roll in Amagansett, New York on the Eastern End of Long Island A lobster roll is a fast-food sandwich native to New England made of lobster meat served on a grilled hot dog-style bun with the opening on the top rather than the side. The filling may also contain butter, lemon juice, salt and black pepper, with variants made in other parts of New England replacing the butter with mayonnaise. Others contain diced celery or scallion. Potato chips or french fries are the typical sides."
+      "document_text": "Cookbook: Lobster roll Media: Lobster roll A lobster-salad style roll from The Lobster Roll in Amagansett, New York on the Eastern End of Long Island A lobster roll is a fast-food sandwich native to New England made of lobster meat served on a grilled hot dog-style bun with the opening on the top rather than the side. The filling may also contain butter, lemon juice, salt and black pepper, with variants made in other parts of New England replacing the butter with mayonnaise. Others contain diced celery or scallion. Potato chips or french fries are the typical sides."
     }
   ]
 }


pyserini/server/mcp/tools.py
```diff
@@ -36,13 +36,15 @@ def register_tools(mcp: FastMCP, controller: SearchController):
         description='Perform search on a given index. Returns top‑k hits with docid, score, and snippet.',
     )
     def search(
-        query: str,
         index_name: str,
+        query_txt: str = None,
```
sahel-sh: don't add all these fields here flattened. Follow pyserini.
query can be either a str or dict, use the dict format with these as keys to handle multimodal queries. That would be generalizable to any other future query formats as well, adding video, audio, etc.

pyserini/server/mcp/tools.py
```diff
@@ -52,7 +54,44 @@ def search(
         Returns:
             List of search results with docid, score, and raw contents
         """
-        return controller.search(query, index_name, k, ef_search=ef_search, encoder=encoder, query_generator=query_generator)
+
+        if "m-beir" in index_name:
```
sahel-sh, line 59: if you add query modality to the dict above, this code won't be needed

pyserini/server/search_controller.py
```diff
@@ -113,30 +124,64 @@ def search(
...
                         encoder=encoder,
-                        query_generator=query_generator
+                        query_generator=query_generator,
+                        instruction_config=instruction_config,
                     )
                 )
-
+            
+                if index_config.name in FAISS_INDEX_INFO_M_BEIR.keys():
+                    query = {"qid": "1:1", **query} # dummy qid for UniIRQueryEncoder
```
sahel-sh: this way you are always overriding the qid passed to search, is that intentional?

pyserini/server/search_controller.py
```diff
@@ -113,30 +124,64 @@ def search(
...
+                    query["fp16"] = True
             hits = index_config.searcher.search(query, k)
 
-        results: dict[str, Any] = {'query': {'qid': qid, 'text': query}}
+        if isinstance(query, str) or query.get("query_img_path") is None: # text-only query
```
sahel-sh: how about this code:
if isinstance(query, str):
    results = {'query': {'qid': qid, 'text': query}}
else:
    assert isinstance(quer, dict)
    results = {'query': query} # preserves all query fields
   query_path = query.get("query_img_path", "")
   if query_path:
       results['query']['query_image']: get_image_byes(query_path) # add this helper function

pyserini/server/mcp/tools.py
clides: From the sync meeting: keep this code to handle query modality dynamically

pyserini/server/search_controller.py
clides: I was running into issues with qid before but now after testing i dont think its necessary. This is now handled in the search tool

pyserini/server/mcp/tools.py
clides: Fixed

pyserini/server/search_controller.py
clides: Fixed

pyserini/server/mcp/tools.py
```diff
@@ -33,26 +33,69 @@ def register_tools(mcp: FastMCP, controller: SearchController):
...
+            if query.get('query_txt') and query.get('query_img_path'):
+                query['query_modality'] = "image,text"
+            elif query.get('query_img_path'):
+                query['query_modality'] = "image"
+            else:
+                query['query_modality'] = "text"
+        else:
+            if not query.get('query_txt'):
+                raise ValueError("Missing query text for single modality dataset! Please provide a query text for this index!")
+            query = query.get('query_txt')
```
sahel-sh: nit: after raising the value error it is safe to use query['query_txt']

pyserini/server/mcp/tools.py
```diff
@@ -33,26 +33,69 @@ def register_tools(mcp: FastMCP, controller: SearchController):
...
+            query: Search query dictionary with optional text and image paths
             index_name: Name of index to search (default: use default index)
             k: Number of results to return (default: 10)
         Returns:
             List of search results with docid, score, and raw contents
         """
-        return controller.search(query, index_name, k, ef_search=ef_search, encoder=encoder, query_generator=query_generator)
+
+        if "m-beir" in index_name:
+            query['qid'] = "1:1" # dummy qid for m-beir format
```
sahel-sh: should you check not to overwrite qid if it already exists in the query dict?

clides: in the current implementation we dont need to check since the query from mcp client does not contain qid, if we want to add additional functionality so that the user have the option to also provide the qid in the chat message we can check it.
do you want me to add this functionality? @sahel-sh

sahel-sh (approved):
LGTM with nits, thanks

pyserini/server/mcp/tools.py
clides: Fixed

pyserini/server/mcp/tools.py
```diff
@@ -33,26 +33,71 @@ def register_tools(mcp: FastMCP, controller: SearchController):
...
-        query: str,
+        query: Dict[str, Any],
         index_name: str,
+        intruction_config: str = None,
         k: int = 10,
         ef_search: int = 100,
         encoder: str = None,
         query_generator: str = None
-    ) -> dict[str, Any]:
+    ):
```
lilyjge, line 48: do we know the return type? I think it'd be nice to provide some info about what the method returns either here or in the tool description (so the llm can also see it)

pyserini/server/mcp/tools.py
```diff
@@ -33,26 +33,71 @@ def register_tools(mcp: FastMCP, controller: SearchController):
...
     @mcp.tool(
         name='search',
-        description='Perform search on a given index. Returns top‑k hits with docid, score, and snippet.',
+        description='''Perform search on a given index. Returns top‑k hits with docid, score, and snippet.
+        The "query" argument is a dictionary with format {"query_txt": "...", "query_img_path": "..."}.
+        '''
     )
     def search(
-        query: str,
+        query: Dict[str, Any],
```
lilyjge, line 41: not super necessary for this PR but once you're done with all the MCP updates you'll probably want to update the doc as well https://github.com/castorini/pyserini/blob/master/docs/usage-mcp.md?plain=1#L122

pyserini/server/search_controller.py
```diff
@@ -113,30 +139,50 @@ def search(
...
-                    'score': float(hit.score),
-                    'doc': raw,
-                }
-            )
-        results['candidates'] = candidates
+            encoded_img = doc.get('encoded_img')
+            cand_data = {
+                'docid': hit.docid,
+                'score': float(hit.score),
+                'document_text': raw,
```
lilyjge, line 176: make sure the key change (doc -> document_text) doesn't break anything? this https://github.com/castorini/pyserini/blob/master/docs/usage-rest.md?plain=1#L52 at least is going to need to be updated, but similar to the other doc stuff doesn't necessarily need to be this pr

pyserini/server/mcp/tools.py
clides: no, if i add return type will get some pydantic error, ill look more into it in the future and add if i can find return type

pyserini/server/search_controller.py
clides: Fixed, thanks for letting me know

clides: There are some incompatibilities with the current REST model. I will create an issue and fix it later

Issue #2384, opened 2026-01-06
Incompatibility between new mcp server's search_controller and current REST apis


clides:
Incompatibility in parameters and return types

From PR: https://github.com/castorini/pyserini/pull/2378

PR #2385, opened 2026-01-10
Fixed REST API compatibility issues + refractored mcp

clides:
Issue Ref: https://github.com/castorini/pyserini/issues/2384

- Fixed REST API compatibility issues; the search endpoint now also works with multimodal datasets; modified its documentation respectively
- Refactored MCP and search_controller so it is more compatible with REST APIs
- Fixed unittests

### DIFF
--- a/pyserini/server/mcp/tools.py
+++ b/pyserini/server/mcp/tools.py
@@ -34,7 +34,7 @@ def register_tools(mcp: FastMCP, controller: SearchController):
     @mcp.tool(
         name='search',
         description='''Perform search on a given index. Returns top‑k hits with docid, score, and snippet.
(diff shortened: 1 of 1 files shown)


pyserini/server/search_controller.py
```diff
@@ -128,6 +120,29 @@ def search(
         instruction_config: str | None = None,
     ) -> Dict[str, Any]:
         """Perform search on specified index."""
+
+        # Handle string query input from REST API for simple text queries
+        if isinstance(query, str):
+            query = {"query_txt": query}
+
+        if "m-beir" in index_name:
+            if not query.get('qid'):
```
lilyjge, line 129: maybe add this 'qid' key to the mcp tool description if you ever expect the user to provider a qid?

clides: Added!

PR #2388, opened 2026-01-12
Fix HF truncation warning

### LINKED ISSUE #2360: Fix warning with mrtydi
Warning is probably much more pervasive, but came across it running the following:

```
python -m pyserini.2cr.mrtydi --condition mdpr-tied-pft-msmarco --language ru --display-commands --directory runs/
```

The warning:

```
Asking to truncate to max_length but no maximum length is provided and the model has no predefined maximum length. Default to no truncation.
```

zdann15:
Fixes #2360 by removing truncation argument that didn't do anything and was producing a warning. Set default max_length to 512 tokens.

### DIFF
--- a/pyserini/encode/_auto.py
+++ b/pyserini/encode/_auto.py
@@ -92,10 +92,11 @@ class AutoQueryEncoder(QueryEncoder):
             self.pooling = pooling
             self.l2_norm = l2_norm
             self.prefix = prefix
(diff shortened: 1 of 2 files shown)


MXueguang:
I think hf have many version change since original implementation, as long as the regression test passed I think it should be good.

lintool (approved):

PR #2391, opened 2026-01-16
Added automatic downloading for M-BEIR query images and default instruction files

clides:
- Added logic to automatically download M-BEIR query images and default instruction files if not provided
- Updated M-BEIR 2cr respectively

### DIFF
--- a/pyserini/encode/optional/_uniir.py
+++ b/pyserini/encode/optional/_uniir.py
@@ -108,8 +108,7 @@ class UniIRQueryEncoder:
         fp16 = kwargs.get("fp16", False)
 
         if self.instruction_config is None:
-            self.instruction_config = kwargs.get("instr_file", None)
-            self._get_instruction_config()
+            self.instruction_config = self._get_instruction_config(kwargs.get("instr_file", None))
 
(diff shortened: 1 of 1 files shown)


pyserini/encode/query.py
sahel-sh, line 1: if this is a bug, fix it in a separate pr, it will be buried here.

clides: Ok, I've restored it to original

pyserini/query_iterator.py
```diff
@@ -198,6 +199,7 @@ def get_query(self, id_):
         topic = self.topics[id_]
 
         query_data = {
+            'topics_path': topic['topics_path'],
```
sahel-sh: why are you including topics_path in query_data?
You only need the topics_path at init time, no? in the from_topics method.

sahel-sh (approved):
LGTM, thanks

pyserini/query_iterator.py
sahel-sh: had a slack sync, topics_path was needed for the encoder to find the right config file from its dict. Now the config file is being passed and the dict map is kept here.

PR #2392, opened 2026-01-17
Add reproduction log

izzat5233:
## Setup
- OS: macOS 26.2
- Chip: Apple M3 Pro

## Results
All numbers match the markdown results exactly.

## Issue Encountered
Running `pyserini.eval.trec_eval` breaks due to missing `__init__.py` files in `pyserini/resources/` and `pyserini/resources/jars/`. This is because these files are called in this line: `importlib.resources.files("pyserini.resources.jars")` so they need to be modules not folders. I added them in the PR to fix this.

I also noticed a typo in one of the commands and fixed it.


lintool:
hi @izzat5233 can you provide the simplest example the triggers an error requiring `__init__.py`?

lintool:
Ah. It says in our README: https://github.com/castorini/pyserini/?tab=readme-ov-file#-installation

> Pyserini is built on Python 3.11...

Seems like you're on Python 3.12, which is why none of our other users have encountered this issue before.

PR #2394, opened 2026-01-19
Add unittests for query iterators

clides:
Added unittest for MBEIR, Default, Kilt, Multimodal query iterators

### DIFF
--- a/tests/core/test_query_iterator.py
+++ b/tests/core/test_query_iterator.py
@@ -96,7 +96,7 @@ class TestQueryIterators(unittest.TestCase):
             self.assertEqual(result['query_modality'], 'text')
 
             with self.assertRaises(FileNotFoundError):
-                iterator.get_query("invalid_img")
+                iterator.get_query("invalid_image")
 
             result = iterator.get_query("empty_text_string")
             self.assertEqual(result['query_txt'], '')


tests/core/test_query_iterator.py
```diff
@@ -0,0 +1,116 @@
...
+                "query_img_path": "",
+                "query_modality": "text",
+                "instr_file": "sample_instructions.txt"
+            },
+            "missing_field": {
+                "qid": "missing_field",
+                "query_txt": "Query with missing image path field",
+                "query_modality": "text",
+                "instr_file": "sample_instructions.txt"
+            }
```
sahel-sh, line 62: please add image only as well,
also one with invalid image path

tests/core/test_query_iterator.py
```diff
@@ -0,0 +1,116 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            iterator = MBEIRQueryIterator(test_topics, topic_dir=temp_dir)
+            
+            result = iterator.get_query("text_only")
+            self.assertIsNone(result['query_img_path'])
+            self.assertEqual(result['query_modality'], 'text')
+            
+            result = iterator.get_query("null_string")
+            expected_path1 = os.path.join(temp_dir, "null")
+            self.assertEqual(result['query_img_path'], expected_path1)
```
sahel-sh: hm, I thought you replace this with none, how come it is still tempdir/null?

tests/core/test_query_iterator.py
```diff
@@ -0,0 +1,116 @@
...
+            result = iterator.get_query("empty_string")
+            self.assertIsNone(result['query_img_path'])
+            self.assertEqual(result['query_modality'], 'text')
+            
+            result = iterator.get_query("valid_image")
+            expected_path2 = os.path.join(temp_dir, "images/test.jpg")
+            self.assertEqual(result['query_img_path'], expected_path2)
+            self.assertEqual(result['query_modality'], 'image,text')
+
+            result = iterator.get_query("missing_field")
```
sahel-sh, line 94: some times the txt key can be missing, could please add a test for this case as well?

tests/core/test_query_iterator.py
clides: Added image only
Currently the MBEIRQueryIterator does not check for invalid image path, do you want me to add a test for it in this PR and add the check in another PR? Also currently the MBEIRQueryIterator does not check if the image path is empty string (""), I could add that in the other PR as well.

tests/core/test_query_iterator.py
clides: Oops, fixed

tests/core/test_query_iterator.py
clides: Added

PR #2396, opened 2026-01-21
Integrating MM-Embed

clides:
Integrated MM-Embed to Pyserini (both document and query encoders)

TODO:
- [ ] Currently the instructions handling code is copied directly from uniir-for-pyserini, maybe move them to a utils file in the future so both UniIR models and MM-Embed can use it
- [ ] Figure out the compatibility issues for the dependencies (the working dependencies with MM-Embed are listed under the optional dependency section I've added)
- [ ] If approved by Jimmy or Sahel, add 2CR pages and prebuilt indexes

For recall@5 scores, pyserini integration for CIRR dataset have 0.5211 vs 0.53 for the score recorded in the paper

### DIFF
--- a/pyserini/resources/index-metadata/faiss-flat.m-beir.20250813.a9ec58.README.md
+++ b/pyserini/resources/index-metadata/faiss-flat.m-beir.20250813.a9ec58.README.md
@@ -46,3 +46,24 @@ python -m pyserini.index.faiss \
     --metric inner
 ```
 
+MM-Embed model:
+```bash
+python -m pyserini.encode \
+  input --corpus M-BEIR/cand_pool/local/mbeir_{dataset}_cand_pool.jsonl \
+        --fields img_path modality txt did \
+        --docid-field did \
+  output --embeddings encode/m-beir-{dataset}.mm-embed \
+  encoder --encoder nvidia/MM-Embed \
+          --encoder-class mm-embed \
+          --device cuda:0 \
+          --multimodal \
+          --fields img_path modality txt did \
+          --batch-size 16 --max-length 4096
+
+python -m pyserini.index.faiss \
+    --input encode/m-beir-{dataset}.mm-embed \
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -64,3 +64,12 @@ optional = [
     "pybind11>=2.11.0",
     "uniir-for-pyserini>=0.1.2"
 ]
+mm-embed = [
```
sahel-sh, line 67: only freeze what is absolutely needed and remove those that are part of the core dependencies already:
like scipy and numpy, check for other duplicates.

pyserini/encode/__main__.py
```diff
@@ -58,6 +58,8 @@ def init_encoder(encoder, encoder_class, device, pooling, l2_norm, prefix, multi
         kwargs.update(dict(l2_norm=True))
         if encoder_class is None: # check if the uniir-for-pyserini package is installed
             raise ValueError("UniIR's corpus encoder class is not available (as the uniir-for-pyserini package is not installed or CLIP is not installed). Please run 'pip install pyserini[optional]' to install the uniir-for-pyserini package and run 'pip install git+https://github.com/openai/CLIP.git' to install CLIP.")
+    if _encoder_class == 'mm-embed':
```
sahel-sh, line 63: hm, it is strange that this one doesn't do normalization, I didn't see anything related to pooling either, do you know what's the situation with pooling?

pyproject.toml
clides: I've added those since when i downgraded torch and transformers before, those versions also need to be changed otherwise theres pip errors

pyserini/encode/__main__.py
clides: IIRC ive checked before and the model does internal normalization, when i checked the embedding weights, it adds up to 1, not sure about pooling

PR #2399, opened 2026-01-26
Add 2CR page for DSE

zdann15:
Adds 2cr page for DSE integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ Additional reproduction guides below provide detailed step-by-step instructions.
 
 ## 📜️ Release History
 
++ v1.5.0 (w/ Anserini v1.5.0): January 14, 2026 [[Release Notes](docs/release-notes/release-notes-v1.5.0.md)]
 + v1.4.0 (w/ Anserini v1.4.0): December 16, 2025 [[Release Notes](docs/release-notes/release-notes-v1.4.0.md)]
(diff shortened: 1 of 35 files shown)


docs/2cr/dse.html
```diff
@@ -0,0 +1,422 @@
...
+<div style="padding-top: 20px"/>
+<h4>Wiki-SS-NQ Results</h4>
+
+<div class="table-responsive">
+  <table class="table">
+    <thead>
+      <tr>
+        <th class="headertop"></th>
+        <th class="headertop"></th>
+        <th class="headertop" colspan="4"><b>Phi3-vision</b></th>
```
sahel-sh: -Phi3-vision reads like the out of the box model, but this is the SDE model finetuned from Phi-3. Please rename it to SDE-Phi-3... based on the model that you have used from here: https://huggingface.co/Tevatron/models
-Also add [1] for the reference, look into other 2crs as an example

docs/2cr/dse.html
```diff
@@ -0,0 +1,422 @@
...
+      </tr>
+      <tr>
+        <th class="headerbottom" scope="col"></th>
+        <th class="headerbottom" scope="col"></th>
+        <th class="headerbottom" scope="col">Top-1</th>
```
sahel-sh: I know this is what the paper uses, but it is not clear what it means.
change them to Accuracy@k

PR #2401, opened 2026-01-28
Qwen3 encoder integration for bright

Raghav0005:


pyserini/encode/__main__.py
```diff
@@ -148,7 +148,9 @@ def parse_args(parser, commands):
             }
             # Prepare input_kwargs for the encoder
             if not args.encoder.multimodal:
-                kwargs['texts'] = batch_info['text'] # pyserini text encoders takes 'texts' as default input    
+                # Use the first encoder field as the primary text source
+                default_field = args.encoder.fields[0] if args.encoder.fields else 'text'
```
sahel-sh: This is a dangerous change since it will impact all the other encoders. Don't do this. Look into how bge model is handling this. https://castorini.github.io/pyserini/2cr/bright.html
you should not need this change. @lilyjge who has worked with bright topics and corpus before

pyserini/encode/_diver.py
```diff
@@ -0,0 +1,89 @@
...
+            self.l2_norm = l2_norm
+            self.tokenizer = AutoTokenizer.from_pretrained(tokenizer_name or encoder_dir, padding_side='left',
+                                                           trust_remote_code=True, clean_up_tokenization_spaces=True)
+            self.model = AutoModel.from_pretrained(encoder_dir, trust_remote_code=True)
+            self.model.to(self.device).eval()
+            self.has_model = True
+        if (not self.has_model) and (not self.has_encoded_query):
+            raise Exception('Neither query encoder model nor encoded queries provided. Please provide at least one.')
+
+    def encode(self, query: str, max_length=8192, **kwargs):
```
sahel-sh: @Raghav0005 did you set the max_length in your runs?

Pyserini by default uses 256:
https://github.com/castorini/pyserini/blob/master/pyserini/encode/__main__.py#L115
which is definitely too low for bright corpus.

This is while the run.py in the diver code is using 16k
https://github.com/AQ-MedAI/Diver/blob/main/Retriever/retrievers.py#L287


1- Add a print inside encode functions to see what is the max_length that you are getting both for query and for corpus. It would be either 256 or 8k, none of which match what diver is using.
2- Pass 16 in your runs

pyserini/encode/_diver.py
```diff
@@ -0,0 +1,89 @@
...
+def last_token_pool(last_hidden_states: Tensor, attention_mask: Tensor) -> Tensor:
+    left_padding = (attention_mask[:, -1].sum() == attention_mask.shape[0])
+    if left_padding:
+        return last_hidden_states[:, -1]
+    else:
+        sequence_lengths = attention_mask.sum(dim=1) - 1
+        batch_size = last_hidden_states.shape[0]
+        return last_hidden_states[torch.arange(batch_size, device=last_hidden_states.device), sequence_lengths]
+
+class DiverDocumentEncoder(DocumentEncoder):
```
sahel-sh: Diver is just reusing the Qwen3 inference logic. Rename these classes to Qwen3Document and Query encoders

pyserini/encode/__main__.py
```diff
@@ -58,8 +58,8 @@ def init_encoder(encoder, encoder_class, device, pooling, l2_norm, prefix, multi
         kwargs.update(dict(l2_norm=True))
         if encoder_class is None: # check if the uniir-for-pyserini package is installed
             raise ValueError("UniIR's corpus encoder class is not available (as the uniir-for-pyserini package is not installed or CLIP is not installed). Please run 'pip install pyserini[optional]' to install the uniir-for-pyserini package and run 'pip install git+https://github.com/openai/CLIP.git' to install CLIP.")
-    if _encoder_class == 'dse' or 'dse' in encoder.lower():
-        kwargs.update(dict(l2_norm=True, multimodal=multimodal, pooling=pooling))
+    if _encoder_class == 'diver':
```
sahel-sh: qwen3 class is more accurate

pyserini/encode/_diver.py
Raghav0005: Yeah I think it was defaulting to 256, I'll use 16384 now.

sahel-sh:
closing in favor of PR #2408

PR #2408, opened 2026-01-31
qwen3 embed integration + diver retriever for bright

sahel-sh:
This pr:
- adds the qwen3 encoder
- adds and experiments*.md file for bright + diver retriever

Tested:
- pony + diver gets ndcg@10 higher than the original paper [improvement is due to some minor bug in the original code]

### DIFF
--- a/docs/experiments-qwen3-bright.md
+++ b/docs/experiments-qwen3-bright.md
@@ -1,5 +1,8 @@
 # Pyserini: Reproducing Bright leaderboard with Diver Model - Finetuned from Qwen3
+This guide contains steps to reproduce retrieval results for BRIGHT using the Diver Model introduced in the following paper:
 
+Meixiu Long and Duolin Sun and Dan Yang and Junjie Wang and Yue Shen and Jian Wang and Peng Wei and Jinjie Gu and Jiahai Wang.
+[DIVER: A Multi-Stage Approach for Reasoning-intensive Information Retrieval](https://arxiv.org/abs/2508.07995) _arXiv:2508.07995_
 
 ## Data Preparation
 You need to obtain the relevant corpra from [Bright Corpus](https://huggingface.co/datasets/castorini/collections-bright/tree/main/corpus).


docs/experiments-qwen3-bright.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Evaluation
+`nDCG@10` is the metric used in the Bright leaderboard and is what we use here for easior comparison.
+```
+python -m pyserini.eval.trec_eval -c -m ndcg_cut.10 -m recall.100 \
+  bright-${dataset_name} \
+  runs/run.bright-${dataset_name}.${model_name}.txt
+```
+
```
lilyjge: do we have the results? we should probably add them here?

sahel-sh: not yet, just for pony

lilyjge: maybe we could add that with comparison to their official results, and iterate as we get more splits? I don't feel too strongly though

docs/experiments-qwen3-bright.md
```diff
@@ -0,0 +1,76 @@
...
+```
+gunzip pony.jsonl.gz
+```
+
+## Encoding and indexing the Corpus
+For all the following commands, ensure to update the directory paths accordingly.
+Encode document screenshots into dense vectors:
+
+```bash
+model="AQ-MedAI/Diver-Retriever-4B"
```
lilyjge, line 24: let's cite the paper for the model?

docs/experiments-qwen3-bright.md
sahel-sh: added the result

docs/experiments-qwen3-bright.md
sahel-sh: done

lilyjge (approved):

Raghav0005:
> @Raghav0005 PTAL

LGTM!

PR #2423, opened 2026-02-27
Fix test cases for anserini v1.6.0

lintool:
This was _not_ aided by AI at all. As expected, Codex failed to propagate changes from Anserini.

### DIFF
--- a/tests/optional/test_nfcorpus_faiss.py
+++ b/tests/optional/test_nfcorpus_faiss.py
@@ -70,11 +70,9 @@ class TestNFCorpus(unittest.TestCase):
                   --output {run_file} \
                   --batch 32 --threads 4 \
                   --hits 10'
-        #print(cmd)
 
         os.system(cmd)
-        results = subprocess.check_output(
-            f'python -m pyserini.eval.trec_eval -c -m ndcg_cut.10 {self.qrels} {run_file}', shell=True)
+        results = subprocess.check_output(f'python -m pyserini.eval.trec_eval -c -m ndcg_cut.10 {self.qrels} {run_file}', shell=True)
         results = results.decode('utf-8').split('\n')
         ndcg_line = results[-2]
         ndcg_score = float(ndcg_line.split('\t')[-1])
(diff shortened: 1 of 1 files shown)


pyserini/prebuilt_index_info.py
```diff
@@ -1996,193 +2014,226 @@ def import_from_lucene(enum):
...
+    "beir-v1.0.0-scifact.bge-base-en-v1.5.hnsw": import_from_hnsw_lucene(JPrebuiltHnswIndex.get('beir-v1.0.0-scifact.bge-base-en-v1.5.hnsw')),
 }
 
 LUCENE_HNSW_INDEX_INFO = {**LUCENE_HNSW_INDEX_INFO_MSMARCO,
                           **LUCENE_HNSW_INDEX_INFO_BEIR}
 
 
+JPrebuiltFlatIndex = autoclass('io.anserini.index.prebuilt.PrebuiltFlatIndex')
+
+def import_from_flat_lucene(index_metadata):
```
lilyjge, line 2176: does this need to be a different method from import_from_hnsw_lucene?

tests/optional/test_nfcorpus_faiss.py
```diff
@@ -61,13 +61,16 @@ def setUpClass(cls):
...
+
         cmd = f'python -m pyserini.search.faiss \
                   --encoder-class contriever --encoder facebook/contriever-msmarco \
-                  --index {self.dense_index_dir}/faiss.nfcorpus.contriever-msmarco \
+                  --index {self.dense_index_dir}/{subdir} \
                   --topics {self.queries} \
                   --output {run_file} \
                   --batch 32 --threads 4 \
                   --hits 10'
+        #print(cmd)
```
lilyjge: should this be kept?

pyserini/prebuilt_index_info.py
lintool: I'm mapping each Java class to a Python method for future proofing - in case they diverge in the future.

tests/optional/test_nfcorpus_faiss.py
lintool: yupp, good catch.

pyserini/prebuilt_index_info.py
lilyjge: ah, makes sense!

PR #2446, opened 2026-03-22
Add extracting queries and documents from TREC run

lilyjge:
This has been on the backburner for a while now since the initial implementation of `GenerateRerankerRequests` in Anserini, but with its recent refactor into `ExtractQueriesAndDocumentsFromTrecRun` this is as good a time as any to finally add bring it to Pyserini. Functionally it pretty much does the same things. Cursor wrote most of it but I looked through and sanity checked, especially the tests. 
This will make it more convenient for Pyserini runs -> reranker requests for RankLLM.

### DIFF
--- a/pyserini/eval/extract_queries_and_documents_from_trec_run.py
+++ b/pyserini/eval/extract_queries_and_documents_from_trec_run.py
@@ -1,0 +1,240 @@
+#
+# Pyserini: Reproducible IR research with sparse and dense representations
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pyserini/eval/extract_queries_and_documents_from_trec_run.py
```diff
@@ -0,0 +1,240 @@
...
+            return topics[k].get(field)
+    return None
+
+
+def _maybe_parse_json_document(raw: str) -> Any:
+    """If raw looks like JSON, parse it; otherwise return the string (matches Anserini behavior)."""
+    if not raw:
+        return raw
+    i = 0
+    while i < len(raw) and raw[i].isspace():
```
sahel-sh, line 81: why not?
try:
   return json.loads(raw.strip())
except ....:
   return raw

sahel-sh (commented):
LGTM % Nit,

Should we add a run time flag, like --generate-reranking-queries or something similar to the highlevel retrieval command? Or do you want to keep this as a explicit/separate post processing step?

lintool:
This is defunct... I don't think we have a need for this anymore?

PR #2460, opened 2026-04-08
Fix trec eval python312

namatvir:
Issue:
When running pyserini.eval.trec_eval on Python 3.12, a StopIteration error occurs. This is due to the way importlib.resources.files().joinpath('') behaves in Python 3.12, which seems to differ from previous versions or expects a different usage pattern.

Changed:
jar_directory = str(importlib.resources.files("pyserini.resources.jars").joinpath('')) 
to:
jar_directory = str(importlib.resources.files("pyserini") / "resources" / "jars")
which is compatible with Python 3.12's updated behaviour.


lintool:
@namatvir why are there diffs on the onboarding docs and conflicts?

can you add test cases?

PR #2464, opened 2026-04-14
Fix transformers 5.x compatibility for ANCE, UniCoil, and CLIP encoders

namatvir:
After upgrading to transformers 5.X, 3 issues were found.

**Issue 1. "all_tied_weights_keys" attribute was missing in ANCE and UniCoil**
- Transformers5.X introduces new logic which expects models to define all_tied_weights_keys
- Custom encoder classes did not implement this which led to attribute error

**Issue 2. Incorrect/missing weight loading in ANCE and UniCoil**
- "from_pretrained()" no longer correctly loads all weights for custom model components.
- In ANCE: embeddingHead and norm were not loaded and left at default initialization
- in UniCoil: tok_proj weights were not loaded correctly

**Issue 3. Changed return type in CLIP**
- In transformers5.X, get_text_features() doesn't return a tensor but instead a BaseModelOutputWithPooling object
- Existing code assumed a tensor and called .detach() directly which caused an AttributeError

**Implemented fix 1:**
Added all_tied_weights_keys property for ANCE and UniCoil encoders

**Implemented fix 2:**
Manually loaded weights and bias. 
(embeddingHead.weight/.bias for ANCE and coil_encoder.tok_proj.weight/.bias for UniCoil)

**Implemented fix 3:**
Updated code to extract correct tensor with new return type
embeddings = text_features.pooler_output.detach().cpu().numpy()

_verified with: python -m pytest tests/core/ -v --ignore=tests/core/test_rest.py --ignore=tests/core/test_mcp.py 2>&1 | tail -30_

### DIFF
--- a/pyserini/encode/_ance.py
+++ b/pyserini/encode/_ance.py
@@ -18,15 +18,19 @@ from typing import Optional, List
 
 import torch
 from transformers import PreTrainedModel, RobertaConfig, RobertaModel, RobertaTokenizer, requires_backends
+from transformers.utils import cached_file
 
 from pyserini.encode import DocumentEncoder, QueryEncoder
 
 
 class AnceEncoder(PreTrainedModel):
     config_class = RobertaConfig
-    base_model_prefix = 'ance_encoder'
+    base_model_prefix = 'roberta'
     load_tf_weights = None
-
+    @property
+    def all_tied_weights_keys(self):
+        return {}
+    
     def __init__(self, config: RobertaConfig):
         requires_backends(self, 'torch')
         super().__init__(config)
@@ -74,6 +78,17 @@ class AnceDocumentEncoder(DocumentEncoder):
     def __init__(self, model_name, tokenizer_name=None, device='cuda:0'):
         self.device = device
         self.model = AnceEncoder.from_pretrained(model_name)
+        # Manually load embeddingHead and norm weights for transformers 5.x compatibility
+        weights_path = cached_file(model_name, 'pytorch_model.bin')
+        state_dict = torch.load(weights_path, map_location='cpu', weights_only=True)
+        self.model.embeddingHead.load_state_dict({
+            'weight': state_dict['embeddingHead.weight'],
+            'bias': state_dict['embeddingHead.bias']
+            })
+        self.model.norm.load_state_dict({
+            'weight': state_dict['norm.weight'],
+            'bias': state_dict['norm.bias']
+            })
         self.model.to(self.device)
         self.tokenizer = RobertaTokenizer.from_pretrained(tokenizer_name or model_name,
                                                           clean_up_tokenization_spaces=True)
@@ -100,6 +115,18 @@ class AnceQueryEncoder(QueryEncoder):
         if encoder_dir:
             self.device = device
             self.model = AnceEncoder.from_pretrained(encoder_dir)
+            self.model = AnceEncoder.from_pretrained(encoder_dir)
+# Manually load embeddingHead and norm weights for transformers 5.x compatibility
+            weights_path = cached_file(encoder_dir, 'pytorch_model.bin')
+            state_dict = torch.load(weights_path, map_location='cpu', weights_only=True)
+            self.model.embeddingHead.load_state_dict({
+                'weight': state_dict['embeddingHead.weight'],
+                'bias': state_dict['embeddingHead.bias']
+            })
+            self.model.norm.load_state_dict({
+                'weight': state_dict['norm.weight'],
+                'bias': state_dict['norm.bias']
+            })
             self.model.to(self.device)
             self.tokenizer = RobertaTokenizer.from_pretrained(tokenizer_name or encoder_dir,
                                                               clean_up_tokenization_spaces=True)
--- a/pyserini/encode/_ance.py
+++ b/pyserini/encode/_ance.py
@@ -25,6 +25,22 @@ from packaging.version import Version
 from transformers import __version__ as transformers_version
 
 
+def _load_ance_model(model_name_or_path: str, device: str):
+    model = AnceEncoder.from_pretrained(model_name_or_path)
+    if Version(transformers_version) >= Version("5.0.0"):
+        load_head_weights(model, model_name_or_path, {
+            'embeddingHead': {
+                'weight': 'embeddingHead.weight',
+                'bias': 'embeddingHead.bias'
+            },
+            'norm': {
+                'weight': 'norm.weight',
+                'bias': 'norm.bias'
+            }
+        })
+    model.to(device)
+    return model
+
 class AnceEncoder(PreTrainedModel):
     config_class = RobertaConfig
     base_model_prefix = 'ance_encoder'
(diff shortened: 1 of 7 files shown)


pyserini/encode/_ance.py
```diff
@@ -18,15 +18,19 @@
...
 from transformers import PreTrainedModel, RobertaConfig, RobertaModel, RobertaTokenizer, requires_backends
+from transformers.utils import cached_file
 
 from pyserini.encode import DocumentEncoder, QueryEncoder
 
 
 class AnceEncoder(PreTrainedModel):
     config_class = RobertaConfig
-    base_model_prefix = 'ance_encoder'
+    base_model_prefix = 'roberta'
```
lintool, line 28: Why are we changing this?

pyserini/encode/_ance.py
```diff
@@ -74,6 +78,17 @@ class AnceDocumentEncoder(DocumentEncoder):
     def __init__(self, model_name, tokenizer_name=None, device='cuda:0'):
         self.device = device
         self.model = AnceEncoder.from_pretrained(model_name)
+        # Manually load embeddingHead and norm weights for transformers 5.x compatibility
+        weights_path = cached_file(model_name, 'pytorch_model.bin')
+        state_dict = torch.load(weights_path, map_location='cpu', weights_only=True)
+        self.model.embeddingHead.load_state_dict({
+            'weight': state_dict['embeddingHead.weight'],
+            'bias': state_dict['embeddingHead.bias']
+            })
```
lintool, line 87: we should add a linter but this code formatting isn't correct PEP 8?

pyserini/encode/_ance.py
```diff
@@ -100,6 +115,18 @@ def __init__(self, encoder_dir: str = None, tokenizer_name: str = None,
         if encoder_dir:
             self.device = device
             self.model = AnceEncoder.from_pretrained(encoder_dir)
+            self.model = AnceEncoder.from_pretrained(encoder_dir)
+# Manually load embeddingHead and norm weights for transformers 5.x compatibility
```
lintool, line 119: Indent?

pyserini/encode/_unicoil.py
```diff
@@ -29,6 +30,9 @@ class UniCoilEncoder(PreTrainedModel):
     config_class = BertConfig
     base_model_prefix = 'coil_encoder'
     load_tf_weights = None
+    @property
```
lintool, line 33: probably want empty line before?

pyserini/encode/_ance.py
lintool: and elsewhere below?

lintool (changes requested):
Can you add test cases?

namatvir:
Added test cases for ANCE and UniCoil models, will add a Clip test case next PR. 
Fixed previous formatting errors.

lintool:
@sahel-sh @ronakice PTAL

pyserini/encode/_unicoil.py
```diff
@@ -78,6 +83,12 @@ class UniCoilDocumentEncoder(DocumentEncoder):
     def __init__(self, model_name, tokenizer_name=None, device='cuda:0'):
         self.device = device
         self.model = UniCoilEncoder.from_pretrained(model_name)
+        weights_path = cached_file(model_name, 'pytorch_model.bin')
```
sahel-sh, line 86: repeated code, create a function please, also guard the logic with the transformers version so that it is a noop on earlier version or early return if version is <5

pyserini/encode/_clip.py
```diff
@@ -109,7 +109,9 @@ def encode(self, texts, max_length=77, **kwargs):
         with torch.no_grad():
             text_features = self.model.get_text_features(**inputs)
 
-        embeddings = text_features.detach().cpu().numpy()
+        # In transformers 5.x, get_text_features() returns BaseModelOutputWithPooling
+        # instead of a tensor directly; use pooler_output to get the text embeddings
+        embeddings = text_features.pooler_output.detach().cpu().numpy()
```
sahel-sh, line 114: use an if else to ensure pyserini still works with older transformer versions

tests/core/test_encoder_model_ance.py
```diff
@@ -51,8 +51,15 @@ def test_ance_encoder(self):
             encoded_vector = np.array(encoder.encode(topics[t]['title']))
 
             l1 = np.sum(np.abs(cached_vector - encoded_vector))
+            # print(f'l1 values is {l1}')
```
sahel-sh, line 54: remove?

tests/core/test_encoder_model_ance.py
```diff
@@ -51,8 +51,15 @@ def test_ance_encoder(self):
...
             self.assertTrue(l1 < 0.0005)
 
+    def test_ance_weights_loaded(self):
+        encoder = AnceQueryEncoder('castorini/ance-msmarco-passage')
+        norm_weights = encoder.model.norm.weight.detach().cpu().numpy()
+        self.assertFalse(np.allclose(norm_weights, np.ones_like(norm_weights)),
```
sahel-sh, line 60: check bias too?

tests/core/test_encoder_model_unicoil.py
```diff
@@ -89,6 +90,12 @@ def test_onnx_encode_unicoil(self):
         os.remove(tarball_name)
         shutil.rmtree(index_dir)
 
+    def test_unicoil_weights_loaded(self):
+        encoder = UniCoilDocumentEncoder('castorini/unicoil-msmarco-passage', device='cpu')
+        tok_proj_weights = encoder.model.tok_proj.weight.detach().cpu().numpy()
```
sahel-sh, line 95: ditto

namatvir:
Added CLIP test case and fixed previous problems

sahel-sh (approved):

pyserini/encode/_ance.py
```diff
@@ -69,11 +76,28 @@ def forward(self, input_ids: torch.Tensor, attention_mask: Optional[torch.Tensor
         pooled_output = self.norm(self.embeddingHead(pooled_output))
         return pooled_output
 
+def load_head_weights(model, model_name, weight_map):
```
sahel-sh: move this to a util under encode so that it is not repeated across files

pyserini/encode/_ance.py
```diff
@@ -69,11 +76,28 @@ def forward(self, input_ids: torch.Tensor, attention_mask: Optional[torch.Tensor
...
+    state_dict = torch.load(weights_path, map_location='cpu', weights_only=True)
+    for module_name, keys in weight_map.items():
+        module = getattr(model, module_name)
+        module.load_state_dict({k: state_dict[v] for k, v in keys.items()})
 
 class AnceDocumentEncoder(DocumentEncoder):
     def __init__(self, model_name, tokenizer_name=None, device='cuda:0'):
         self.device = device
         self.model = AnceEncoder.from_pretrained(model_name)
+        if Version(transformers_version) >= Version("5.0.0"):
```
sahel-sh: here and line 127 except for the model_name vs encoder_dir are identical. create a helper for this

pyserini/encode/_unicoil.py
```diff
@@ -73,11 +80,24 @@ def forward(
...
+    for module_name, keys in weight_map.items():
+        module = getattr(model, module_name)
+        module.load_state_dict({k: state_dict[v] for k, v in keys.items()})
 
 class UniCoilDocumentEncoder(DocumentEncoder):
     def __init__(self, model_name, tokenizer_name=None, device='cuda:0'):
         self.device = device
         self.model = UniCoilEncoder.from_pretrained(model_name)
+        if Version(transformers_version) >= Version("5.0.0"):
+            load_head_weights(self.model, model_name, {
```
sahel-sh: ditto

pyserini/encode/_ance.py
```diff
@@ -99,8 +120,7 @@ def __init__(self, encoder_dir: str = None, tokenizer_name: str = None,
         super().__init__(encoded_query_dir)
         if encoder_dir:
             self.device = device
-            self.model = AnceEncoder.from_pretrained(encoder_dir)
-            self.model.to(self.device)
+            self.model = _load_ance_model(encoder_dir, device)
```
lintool: Can you explain the rationale for a class calling a function outside the class? Doesn't this violate the point of OO abstractions?

namatvir: I moved the helper function (load ance/unicoil model) into the encoder classes as class methods at the end of each encoder class. I can move them higher at the beginning of each class or after __init__ if that is more readable.

lintool (approved):

Issue #2467, opened 2026-04-16
Align Pyserini REST API with Anserini REST API


lintool:
We already have REST APIs in both Pyserini and Anserini.

The one in Anserini is more refined, complete with openAPI spec: https://github.com/castorini/anserini/blob/master/src/main/resources/rest/openapi.yaml

We should bring the Pyserini REST API up to speed, and ensure that changes are propagated over the MCP.

PR #2471, opened 2026-04-18
Align REST API with Anserini 

lilyjge:
Addresses #2467

For the sake of speed as to not block anything, this new REST API aligns strictly with the OpenAPI specification from Anserini. This means some features of the existing REST API are lost, eg listing indexes. We also use a new backend separate from the search controller used by the existing REST API and the MCP server. I think this is fine, we can consolidate the two, straighten out and extend schemas ie add back functionalities/align MCP later based on the foundation of Anserini's specification.

Also, Anserini's REST server uses its [CLI utils](https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/cli/CliUtils.java) and I don't think Pyserini has an equivalent, so the necessary logic is added directly for the REST API in the form of `pyserini/server/rest/document_format.py`. If Pyserini gets a CLI later this code could probably be moved there instead, or to a more general place for utilities if needed.


lintool:
@codex review

lintool:
First testing end to end.

```bash
% cat .rest-config
indexes:
  fineweb-edu-100b: /Users/jimmylin/indexes/lucene-inverted.fineweb-edu-100b-karpathy.20260416
```

Swagger UI works fine.

Screenshot 2026-04-18 at 8 30 17 AM

Something like this works fine also:

```bash
% curl -X 'GET' \
  'http://localhost:8081/v1/fineweb-edu-100b/search?query=Albert%20Einstein&hits=5' \
  -H 'accept: application/json' | jq
```

docs/usage-rest.md
```diff
@@ -1,87 +1,140 @@
-# Pyserini: REST API Server with FastAPI
+# Pyserini: REST API server (FastAPI)
```
lintool, line 1: Is this automatically written by an LLM? If so, that's fine... we can just ask the LLM to sync/update between openapi spec and this.

If you actually wrote this by hand... maybe in the future ask the LLM to update?

pyserini/server/rest/app.py
```diff
@@ -1,58 +1,151 @@
...
+
+
+def _load_openapi_text() -> str:
+    return resources.files('pyserini.server.rest').joinpath('openapi.yaml').read_text(encoding='utf-8')
+
+
+def create_app(index_config_path: str | None = None) -> FastAPI:
+    effective_config = index_config_path
+    if effective_config is None:
+        effective_config = os.environ.get('PYSERINI_INDEX_CONFIG')
```
lintool: should we be reading from an env variable?

pyserini/server/rest/app.py
```diff
@@ -1,58 +1,151 @@
...
+    parser.add_argument('--host', type=str, default='0.0.0.0', help='Bind address (default: 0.0.0.0)')
+    parser.add_argument('--port', type=int, default=8081, help='Port (default: 8081)')
     parser.add_argument(
-        '--port',
-        type=int,
-        default=8081,
-        help='Port to run the server on (default: 8081)',
+        '--index-config',
+        type=str,
+        default=None,
```
lintool: I think it's fine to have this on a single line, even if it runs long...

pyserini/server/rest/backend.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Any
+from urllib.parse import unquote
+
+from pyserini.search.lucene import LuceneSearcher
+from pyserini.server.rest.document_format import format_lucene_document
+from pyserini.server.rest.index_config import load_index_aliases
+from pyserini.util import download_prebuilt_index
+
+
+class LuceneRestBackend:
```
lintool: rename to `LuceneSearcherRestBackend` just to be explicit?

pyserini/server/rest/backend.py
```diff
@@ -0,0 +1,118 @@
...
+            path = self.resolve_index_dir(index_token)
+            if path is None:
+                return None
+            try:
+                self._searchers[index_token] = LuceneSearcher(path)
+            except Exception:
+                return None
+        return self._searchers[index_token]
+
+    def search(self, index_token: str, query: str, hits: int, parse: bool) -> dict[str, Any] | None:
```
lintool, line 76: why not raise an exception instead of returning `None`?

pyserini/server/rest/document_format.py
```diff
@@ -0,0 +1,71 @@
...
+
+import json
+from typing import Any
+
+from pyserini.index.lucene import Document
+
+_SKIP_FIELDS = frozenset({'id', '_id', 'docid'})
+
+
+def _convert_json_value(value: Any) -> Any:
```
lintool, line 33: Does it make sense to have an object, something like `LuceneDocumentFormatter` instead of a bunch of functions?

pyserini/server/rest/document_format.py
```diff
@@ -0,0 +1,71 @@
+#
```
lintool, line 1: Can we write some test cases to test these functions in isolation? should be pretty easy to test.

pyserini/server/rest/index_config.py
```diff
@@ -0,0 +1,61 @@
+#
```
lintool, line 1: Don't feel too strongly, but instead of a file with a single function, maybe fold this into `backend.py`?

lintool:
@ronakice @sahel-sh can we get your 👀 on this also, maybe the next iteration once @lilyjge addresses my feedback?

docs/usage-rest.md
lilyjge: yep, I had the LLM write this after I was happy with the code!

pyserini/server/rest/app.py
lilyjge: I'm agnostic, my agent decided to include it. should I take it out?

lintool: yea, I think so... let's not add things unless absolutely necessary to reduce slop.

docs/usage-rest.md
lintool: sg

pyserini/server/rest/backend.py
lilyjge: the `None` still gets caught and turned into a `400`: https://github.com/castorini/pyserini/pull/2471/changes#diff-83184a168a99d9b8930e063f25d5552a7f406d0cc6dc27a5a4344d95aebc69f6R81

I think it's a bit cleaner for now than having custom exception types and handlers

lintool: ok, don't feel too strongly.

pyserini/server/rest/document_format.py
lilyjge: not sure it would add much value?

lintool: don't feel strongly.

pyserini/server/rest/index_config.py
lilyjge: I'll keep it separate for now, because the backend might be refactored for MCP but the index config is just for REST

lintool: Actually @lilyjge we probably want the index config to be part of MCP also... but we can deal with this later in a separate PR.

ronakice:
Mostly looks good to me.

A couple of broader questions for follow-up, not necessarily blockers for this PR:

Concurrency / non-blocking behavior: we should be aiming for multiple agent requests to proceed independently, rather than having later search calls wait on earlier ones. In Pyserini this PR already moves the request handler off the event loop via asyncio.to_thread, but I’m not sure what the effective backend concurrency story is. Has this been tested under concurrent load? Codex says it looks like there may still be serialization around the cached searcher, so if true concurrency is a goal, this may need corresponding thought on the Anserini side as well. Maybe @lintool knows better?

Scores in the response: what is the intended value of returning raw scores with many significant figures? I can see the case for human debugging / inspection, but for agent-facing usage this feels like a lot of tokens for limited value. It might be worth thinking about whether scores should be optional, rounded, or omitted by default in some future iteration, although I realize they are part of the current schema/API parity story.

I’m not sure yet how this was surfaced or should be surfaced in the paper so probably best to wait for Sahel for that.

lintool:
@lilyjge @ronakice re: score

https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/rerank/lib/ScoreTiesAdjusterReranker.java

```
      //   1 docA 23.439316
      //   2 docS 22.087432
      //   3 docT 22.087432
      //   4 docZ 21.602508
```

Anserini generates to 6 digits (needed due to tie-breaking, etc. - see above). Make it consistent here?

lintool (approved):
+1 from me.

Maybe change the number of digits in score thing?

PR #2474, opened 2026-04-19
Refactor MCP and REST APIs to use same backend

lilyjge:
Consolidates the REST backend added in #2471 with the MCP backend

REST API behaviour does not change. MCP tool param names change slightly. 
This looks really big, but it's mostly just moving stuff around.


lintool:
@lilyjge I took a pass... lgtm, but I might be missing some context so I'd like @sahel-sh and @clides to take a look also...

clides:
Yep lgtm

docs/usage-mcp.md
```diff
@@ -105,59 +114,60 @@ Modify your Claude Desktop configuration file `claude_desktop_config.json` with
...
-**Tool Name:** `search`
+Runs retrieval against a Pyserini index (BM25 for standard sparse indexes; dense / impact / FAISS as appropriate for the index type). Results are returned as a **rich list** (human-readable lines plus optional **`Image`** parts for multimodal indexes), not as a single JSON table.
 
-**Description:** Perform a BM25 search on a given index and return top-k hits with document ID, score, and text snippet.
+| Parameter | Notes |
+|-----------|--------|
+| `query` | String, or for multimodal / m-BEIR-style indexes a dict such as `{"query_txt": "...", "query_img_path": "..."}` (local path or image URL where supported). |
+| `index` | Prebuilt index name or alias (default: `msmarco-v2.1-doc-segmented`). Use `list_indexes` to discover names by type. |
+| `hits` | Top‑k (default: 10). |
+| `parse` | Kept for parity with the REST API; MCP output is still rendered as rich content. |
```
sahel-sh: using the same backend doesn't mean the same list of parameters. if this is a noop for mcp with no plans to add support, we should drop this

docs/usage-mcp.md
```diff
@@ -105,59 +114,60 @@ Modify your Claude Desktop configuration file `claude_desktop_config.json` with
...
+| `ef_search` | HNSW / dense Lucene HNSW indexes (default: 100). |
+| `encoder` | Encoder id when the index requires one (dense / FAISS / etc.). |
```
sahel-sh: some of the params have lost their type and whether or not they are optional, it is no longer needed? It might still be helpful to keep them?

docs/usage-mcp.md
```diff
@@ -105,59 +114,60 @@ Modify your Claude Desktop configuration file `claude_desktop_config.json` with
...
-**Returns:** List of search results, each containing:
-- `docid`: Document identifier
-- `score`: BM25 relevance score
-- `query_text`: Text snippet from the document
-- `query_image`: Image snippet from the document (if applicable)
-- `index_name`: Name of the index searched
+### `get_document`
 
-**Example Usage in MCP Client:**
+Fetch one stored document by id. Parameters: `docid`, `index`, optional `parse`. Returns rich content (text and optional **`Image`** when the document includes an image field).
```
sahel-sh: keep the format consistent, |parameters | note| for all tools

sahel-sh: here and for all the other tools

sahel-sh: better use | Name | Required | Default | Description | from the rest md

docs/usage-rest.md
```diff
@@ -68,16 +70,19 @@ If the index cannot be opened, the API responds with **400** and a message such
...
 **Example**
 
 ```bash
-curl -X GET "http://localhost:8081/v1/msmarco-v1-passage/search?query=what%20is%20a%20lobster%20roll&hits=1"
+curl "http://localhost:8081/v1/msmarco-v1-passage/search?query=what%20is%20a%20lobster%20roll&hits=1"
 ```
 
 **Example response (shape)**
 
+Scores are **rounded to six decimal places** in JSON.
```
sahel-sh: it says six decimal, I see 4 in line 96

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+        enc_changed = encoder is not None and _norm_opt_str(encoder) != _norm_opt_str(config.encoder)
+
+        if idx == 'lucene_flat' or idx == 'impact':
+            return enc_changed
+        if idx == 'lucene_hnsw':
+            ef_changed = ef_search is not None and ef_search != config.ef_search
+            return enc_changed or ef_changed
+        if idx == 'faiss':
+            if config.name in FAISS_INDEX_INFO_M_BEIR:
+                instr_changed = instruction_config is not None and _norm_opt_str(instruction_config) != _norm_opt_str(
```
sahel-sh: what if it not none before now it is none? why condition on instrcution_config is not none?

sahel-sh: the same logic is repeated throughout

tests/core/test_mcp.py
```diff
@@ -93,5 +96,50 @@ def test_list_indexes_tool(self):
...
+        ]
+        result = self._run_async(self._call_tool('fuse_search_results', {
+            'hits1': hits1,
+            'hits2': hits2,
+            'k': 2,
+        }))
+        self.assertFalse(result.is_error, msg=getattr(result, 'content', result))
+        self.assertIsNotNone(result.content)
+
+    def test_get_qrels_tool(self):
```
sahel-sh, line 175: are there any invalid cases for this one that should be tested?

tests/core/test_mcp.py
```diff
@@ -93,5 +96,50 @@ def test_list_indexes_tool(self):
...
+            DenseSearchResult('doc_b', 0.5),
+        ]
+        hits2 = [
+            DenseSearchResult('doc_a', 0.4),
+            DenseSearchResult('doc_c', 0.8),
+        ]
+        result = self._run_async(self._call_tool('fuse_search_results', {
+            'hits1': hits1,
+            'hits2': hits2,
+            'k': 2,
```
sahel-sh: if the main api is switching from k to hits, we should be consistent here as well

tests/core/test_mcp.py
```diff
@@ -93,5 +96,50 @@ def test_list_indexes_tool(self):
...
+            mcp = _make_mcp_server()
+            async with run_server_async(mcp, transport='streamable-http') as url:
+                async with Client(StreamableHttpTransport(url)) as client:
+                    await client.call_tool('list_indexes', {'index_type': 'not_a_valid_index_type'})
+
+        with self.assertRaises(ToolError) as ctx:
+            self._run_async(call_invalid())
+        self.assertIn('Index type must be one of', str(ctx.exception))
+
+    def test_get_document_tool(self):
```
sahel-sh, line 128: maybe another test for document not found?

pyserini/server/config.py
```diff
@@ -0,0 +1,66 @@
...
+from pyserini.search.lucene import (
+    LuceneFlatDenseSearcher,
+    LuceneHnswDenseSearcher,
+    LuceneImpactSearcher,
+    LuceneSearcher,
+)
+
+
+@dataclass
+class IndexConfig:
```
sahel-sh, line 36: would it make sense to define a custom __eq__ here of out some of the field out with compare=false and use that to implement _params_require_searcher_rebuild?

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+                    extension = '.png'
+                else:
+                    raise BadSearchRequestError(f'URL does not point to a valid image format: {content_type}')
+                save_path = os.path.join(save_dir, f'{abs(hash(url))}{extension}')
+                with open(save_path, 'wb') as file:
+                    file.write(response.content)
+                query['query_img_path'] = save_path
+
+            if instruction_config is None or not instruction_config.strip():
+                name_to_instr_file = {
```
sahel-sh: this should not be inlined

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+from pyserini.server.config import INDEX_TYPE, SHARDS, IndexConfig
+from pyserini.server.index_config import load_index_aliases
+from pyserini.server.document_format import format_lucene_document
+from pyserini.util import check_downloaded, download_prebuilt_index, download_url, get_cache_home
+
+# Cap for m-beir query images fetched from user-supplied URLs (DoS mitigation: bounded RAM and disk).
+_MAX_M_BEIR_QUERY_IMAGE_BYTES = 50 * 1024 * 1024
+
+
+class BackendError(Exception):
```
sahel-sh: errors can go into different file

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+            config = self.indexes.get(index_name)
+            if config and config.searcher is not None:
+                need_rebuild = config.index_type != 'tf' and self._params_require_searcher_rebuild(
+                    config, ef_search=ef_search, encoder=encoder, instruction_config=instruction_config
+                )
+                if need_rebuild:
+                    try:
+                        config.searcher.close()
+                    except Exception:
+                        pass
```
sahel-sh: maybe log something here with info or warning, this  just silently ignores the exception

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+                    return config
+
+            config = config or IndexConfig(
+                name=index_name,
+                ef_search=ef_search,
+                encoder=encoder,
+                instruction_config=instruction_config,
+            )
+            searcher: LuceneSearcher | LuceneFlatDenseSearcher | LuceneHnswDenseSearcher | LuceneImpactSearcher | FaissSearcher | None = None
+            if config.name in TF_INDEX_INFO:
```
sahel-sh: this code defines an implicit mapping from groups of config.name to hard-coded index_types that should be in sync with INDEX_TYPE.
Please define a lookup function next to INDEX_TYPE's definition, and just call it here in lines 174, 178, etc

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+        doc_config = self._ensure_index(doc_search_index, allow_local=True)
+        raw = doc_config.searcher
+        if isinstance(raw, LuceneSearcher):
+            return raw
+        raise IndexNotAvailableError(
+            f'Document retrieval requires a Lucene sparse index; got {type(raw).__name__!r} '
+            f'(doc index key {doc_search_index!r}).'
+        )
+
+    def _bulk_format_documents(
```
sahel-sh: nit: this is not just format,  fetch and format is a more accurate name

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+    def _resolve_query_generator(self, query_generator_str: str | None, searcher: LuceneSearcher):
+        if not query_generator_str or not query_generator_str.strip():
+            return None
+        name = query_generator_str.strip().lower()
+        if name in ('bagofwords', 'bag_of_words'):
+            return JBagOfWordsQueryGenerator()
+        if name in ('disjunctionmax', 'dismax'):
+            return JDisjunctionMaxQueryGenerator(0.0)
+        if name in ('querysidebm25', 'bm25qs'):
+            return JQuerySideBm25QueryGenerator(0.9, 0.4, searcher.index_reader.reader)
```
sahel-sh, line 278: For future discussion/separate PR:
if we want to do Climbmix with bm25 retrieval only, probably we should enable tuning of the params rather than hardcoding them here.

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+        if name == 'covid19':
+            return JCovid19QueryGenerator()
+        return None
+
+    def _prepare_query(self, query: str | dict[str, Any], index_name: str, instruction_config: str | None) -> tuple[str | dict[str, Any], str | None]:
+        if isinstance(query, str):
+            query = {'query_txt': query}
+
+        if 'm-beir' in index_name:
+            if not query.get('qid'):
```
sahel-sh, line 306: nit: maybe adde a helper function for _prepare_mbeir_query?

pyserini/server/backend.py
```diff
@@ -0,0 +1,483 @@
...
+            doc = docs_by_id[docid]
+            image_path = None
+            encoded_img = None
+            if isinstance(doc, dict):
+                image_path = doc.get('img_path')
+                encoded_img = doc.get('encoded_img')
+            candidates.append(
+                {
+                    'docid': docid,
+                    'score': round(score, 6),
```
sahel-sh: nit: define a constant var for 6

docs/usage-mcp.md
lilyjge: good catch! this is not a noop, doc is wrong

docs/usage-rest.md
lilyjge: that specific example just happens to be zeros after, along with a lot of scores actually. I'll modify the note

pyserini/server/backend.py
lilyjge: for datasets that need it, I think it doesn't make sense for encoder etc. to be turned into none if there's an existing value? in the future, encoder shouldn't even be part of config, we should have some internal mapping since every dense index needs a specific corresponding encoder anyways...

pyserini/server/config.py
lilyjge: I'm going to leave it as if statements for now, most fields shouldn't be updated anyways eg one index should only ever use one encoder
the only field that maybe makes sense to be able to update its value is ef_search, and that only applies Lucene HNSW indexes, so I think it's fine to leave it as is

pyserini/server/backend.py
lilyjge: agreed, let's leave that for later, it'd be a bit annoying since you'd have to reinitialize the searcher: https://github.com/castorini/anserini/blob/master/src/main/java/io/anserini/search/SimpleSearcher.java#L410

pyserini/server/backend.py
lilyjge: I think it's okay after taking out the instruction config resolution

lintool:
@codex review please

lintool (approved):

PR #2480, opened 2026-04-24
Add authentication and logging for REST API 

lilyjge:
Config now allows specifying API keys in the same config YAML as custom indexes
Adds flag for disabling prebuilt indexes, only allow the ones specified in config 
Adds custom logging, looks like this:
`2026-04-23 22:27:42 INFO pyserini.server.rest.auth auth_request client=127.0.0.1 method=GET path=/v1/msmarco/search query=query=lobster&hits=10&parse=true key_id=ed1e1dcf9719 status=200`
Bundles `openapi.yaml` so the Swagger UI is up to date
Includes tests + doc

After this, I think it's ready for public deployment


lintool:
@codex review

pyserini/server/mcp/mcpyserini.py
```diff
@@ -44,17 +44,17 @@ def main():
         help="Port number for HTTP transport (default: 8000)"
     )
     parser.add_argument(
-        "--index-config",
+        "--config",
         type=str,
         default=None,
-        help="YAML file mapping index aliases to paths (Anserini --index-config)",
+        help="YAML server config: indexes: alias -> path (same idea as Anserini --index-config)",
```
lintool:
```suggestion
        help="YAML server config with API tokens and index mappings",
```

lintool: no longer lines up with Anserini, so no need to state

docs/usage-rest.md
```diff
@@ -121,16 +166,21 @@ The `doc` field may be `null`, a string, or a JSON value depending on the index
 curl "http://localhost:8081/v1/msmarco-v1-passage/doc/7157707"
 ```
 
+With API key auth enabled (`api_keys` in `--config`), for example:
+
+```bash
+curl -H "X-API-Key: replace-with-long-random-secret-1" \
```
lintool: let be consistent and use

```
curl -H "Authorization: Bearer replace-with-long-random-secret-1"  ...
```

pyserini/server/rest/openapi.yaml
```diff
@@ -126,6 +143,19 @@ paths:
...
+      type: apiKey
+      in: header
+      name: X-API-Key
+      description: Shared secret issued for this deployment (same value as Bearer token).
+    BearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: API token
+      description: >-
+        Same secret as X-API-Key; send header Authorization with scheme bearer
```
lintool: okay to have this description on a single line

pyserini/server/backend.py
```diff
@@ -220,12 +239,20 @@ def _ensure_index(
             return config
 
     def get_indexes(self, index_type: str) -> list[str]:
+        if self._no_prebuilt_indexes:
+            if index_type != 'tf':
```
lintool: Why is this the case? in theory we could be serving Faiss indexes or impact indexes explicitly?
In fact, for RAG, we want to serve both?

docs/usage-rest.md
lintool: actually, just use api_keys to be consistent?

pyserini/server/rest/app.py
```diff
@@ -65,10 +71,127 @@ def _load_openapi_text() -> str:
...
+    if token is None:
+        return 'missing'
+    t = str(token).strip()
+    if not t:
+        return 'missing'
+    return hashlib.sha256(t.encode('utf-8')).hexdigest()[:12]
+
+
+def _request_query(request: Request) -> str:
+    raw = request.url.query
```
lintool, line 108: seems to be misnamed? this function generates like a shorter version right? rename?

pyserini/server/rest/app.py
```diff
@@ -65,10 +71,127 @@ def _load_openapi_text() -> str:
...
+    raw = request.headers.get('x-api-key') or request.headers.get('X-API-Key')
+    if raw is not None and str(raw).strip():
+        return str(raw).strip()
+    auth = request.headers.get('authorization') or request.headers.get('Authorization')
+    if auth and str(auth).lower().startswith('bearer '):
+        return str(auth[7:]).strip()
+    return None
+
+
+def _token_fingerprint(token: str | None) -> str:
```
lintool: go practice to begin names with verbs?

pyserini/server/rest/app.py
```diff
@@ -130,13 +290,56 @@ def main():
...
+        action='store_true',
+        help='Disable uvicorn default request access logs.',
+    )
     args = parser.parse_args()
 
     if args.port <= 0 or args.port > 65535:
         raise SystemExit('Error: --port must be in [1, 65535]')
 
-    uvicorn.run(create_app(args.index_config), host=args.host, port=args.port)
+    if args.no_prebuilt_indexes:
```
lintool, line 339: seems like you have error trapping _both_ in here and `create_app`?
only need it once? here seems better?

lintool (changes requested):
I'll do multiple rounds of CRs. This is initial skim.

hrm, I'm just thinking outloud - would it make sense to put `--no-prebuilt-indexes` in the config file itself? otherwise we'll have an endless proliferation of command-line options?

pyserini/server/rest/app.py
```diff
@@ -65,10 +71,127 @@ def _load_openapi_text() -> str:
     return resources.files('pyserini.server.rest').joinpath('openapi.yaml').read_text(encoding='utf-8')
 
 
-def create_app(index_config_path: str | None = None) -> FastAPI:
+@lru_cache(maxsize=1)
```
lingwei-gu, line 74: do we need the lru_cache here?

lingwei-gu:
are we using the same auth tokens for mcp as well?

docs/usage-rest.md
lilyjge: do you mean instead of `X-API-Key` and `Authorization: Bearer` the header for auth should just be `api_keys`?

pyserini/server/backend.py
lilyjge: okay, I'm going to add to the index config an option to specify index type then. currently there's no way to tell which type/searcher to use so this is the 'fallback'

pyserini/server/rest/app.py
lilyjge: I'll tell my agent to do better!

lilyjge:
@lingwei-gu I think it's fine to use the same tokens for MCP unless we want permission scopes, in which case we can iterate. however, the code introduced in this PR will not provide MCP with auth since it uses a different system and we need to more careful with it since MCP may be more susceptible to prompt injection attacks

lilyjge:
@lintool I think `--no-prebuilt-indexes` makes sense in the CLI instead of the config file for now, config file can be more 'deployment specific' where this flag always does the same thing. we can change that if CLI options keeps growing a lot

pyserini/server/rest/app.py
lilyjge: allows us to not have to reload the yaml every time

docs/usage-rest.md
```diff
@@ -46,6 +94,9 @@ python -m pyserini.server.rest --index-config /path/to/indexes.yaml
 
 All search and document routes use the **`GET`** method only. Errors return JSON `{"error": "<message>"}` with a 4xx/5xx status where applicable.
 
+When `api_keys` is configured, every `/v1/*` route requires authentication; you can use either
+`Authorization: Bearer <token>` or `X-API-Key: <token>` on any endpoint.
```
lintool: here you say `<token>` below you say `{api-key}`, make consistent?

docs/usage-rest.md
```diff
@@ -140,6 +201,7 @@ curl "http://localhost:8081/v1/msmarco-v1-passage/doc/7157707"
 |------|---------|
 | 200 | Success |
 | 400 | Invalid parameters (e.g. missing `query`, invalid `hits` or `parse`), or cannot open index |
+| 401 | Missing or invalid API credentials (when `api_keys` is configured) |
```
lintool: plural?

pyserini/server/rest/app.py
```diff
@@ -130,13 +294,53 @@ def main():
     parser = argparse.ArgumentParser(description='Run the Pyserini REST API server (Anserini-compatible).')
     parser.add_argument('--host', type=str, default='0.0.0.0', help='Bind address (default: 0.0.0.0)')
     parser.add_argument('--port', type=int, default=8081, help='Port (default: 8081)')
-    parser.add_argument('--index-config', type=str, default=None, help='YAML file mapping index aliases to paths (Anserini --index-config)')
+    parser.add_argument(
+        '--config',
+        type=str,
+        default=None,
+        help='YAML server config: indexes: alias -> path or object {path,index_type,...}; optional non-empty api_keys: enables /v1/ Bearer or X-API-Key auth',
```
lintool: Simplify?

> YAML server config with API tokens and index mapping

Use the same a mcpyserini? propagate to elsewhere where you provide doc for `--config`

pyserini/server/rest/openapi.yaml
```diff
@@ -52,6 +57,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
+        "401":
+          description: Missing or invalid API credentials (only when the server is configured with api_keys).
```
lintool: `api_keys` looks weird... "API keys"?

similar below.

lintool (approved):
nits... fix and then merge.

lintool:
@lilyjge one more thing to add - in an auth error, make sure to include the email to get the auth token? cc @lingwei-gu @nourj98 - to prompt the agent to tell the user to request an API key.

PR #2485, opened 2026-04-27
Add authentication and logging for MCP Server

lilyjge:
#2480 but for MCP 

Pulled out some common code and also added the ability to start both servers at once with a shared backend.


lintool:
@codex review

lintool:
shoot, sorry - must have missed this... I really don't know much about MCP, can we get @sahel-sh or @nli33 to look at this?

PR #2486, opened 2026-04-28
Fixed null condition for bottom_fb_docs and modified test

k464wang:
- Modified bottom hits to take empty list if there are 0 bottom_fb_docs
- Removed unused imports from test

### DIFF
--- a/pyserini/search/lucene/_searcher.py
+++ b/pyserini/search/lucene/_searcher.py
@@ -176,7 +176,7 @@ class LuceneSearcher:
             top_docids = [hit.docid for hit in top_hits]
 
             # Bottom-N docs (optional)
-            bottom_hits = hits[-self.rocchio.bottom_fb_docs:]
+            bottom_hits = [] if self.rocchio.bottom_fb_docs == 0 else hits[-self.rocchio.bottom_fb_docs:]
             bottom_docids = [hit.docid for hit in bottom_hits]
 
             # Extract feedback vectors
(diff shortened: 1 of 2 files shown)


tests/base/test_python_rerankers.py
```diff
@@ -6,14 +6,11 @@
 from urllib.request import urlretrieve
 
 from pyserini.search.lucene import LuceneSearcher
-from pyserini.search.lucene.rerank.rm3_reranker import RM3Reranker
-from pyserini.search.lucene.rerank.rocchio_reranker import RocchioReranker
```
nourj98, line 10: Any reason we are remove this from the tests?

pyserini/search/lucene/_searcher.py
```diff
@@ -350,7 +350,7 @@ def second_pass_search(tid: str, expanded_query):
 
                 # Positive (top) and negative (bottom) feedback docs
                 top_hits = hits[:self.rocchio.top_fb_docs]
-                bottom_hits = hits[-self.rocchio.bottom_fb_docs:]
+                bottom_hits = [] if self.rocchio.bottom_fb_docs == 0 else hits[-self.rocchio.bottom_fb_docs:]
```
nourj98, line 353: Were you able to verify that this gets the same results as the java version when self.rocchio.bottom_fb_docs == 0?

Could be good to perform some tests to check that the results of Rocchio w/ negatives are the same as java...

nourj98 (changes requested):

nourj98 (approved):

PR #2493, opened 2026-05-04
[codex] Add Pyserini uv install skill

lintool:
## Summary

- Add a local Codex skill for installing Pyserini with `uv`.
- Add a counterpart local Codex skill for cloning Pyserini and installing it in editable development mode with `pip install -e .`.
- Include OpenAI agents UI metadata for both skills.

## Validation

- `python /Users/jimmylin/.codex/skills/.system/skill-creator/scripts/quick_validate.py .agents/skills/install-pyserini-uv`
- `python /Users/jimmylin/.codex/skills/.system/skill-creator/scripts/quick_validate.py .agents/skills/install-pyserini-dev`

Validation note: these commands were attempted locally but failed because the current Python environment is missing `PyYAML` (`ModuleNotFoundError: No module named 'yaml'`). Both skills were manually checked against the validator's structural rules: required `SKILL.md`, valid frontmatter, valid hyphen-case names, and matching `agents/openai.yaml`.


lintool:
Following anserini, starting to agentify pyserini - 1st pass, no doubt will need to iterate.

nourj98 (approved):
Excited to try this out!

PR #2495, opened 2026-05-04
Fixed ClipImage and ClipText encoders for Transformers 5.x

namatvir:
Issue: 
Pyserini assumed get_image_features() and get_text_features() always returned tensors and called .detach() directly. Newer transformers versions can return structured outputs instead which causes failures in CLIP multimodal search path. 

Fix:
Added shared output handling logic in BaseClipEncoder so both image and text encoders safely unwrap CLIP feature outputs before converting to NumPy.

Tested with:
_python -m pytest tests/optional/test_multimodal_search.py -k image2image_search -v
python -m pytest tests/core/test_encoder_model_clip.py -v_


namatvir:
@sahel-sh Could you please take a look at this PR for a CR?

namatvir:
Issue resolved in #2497

PR #2497, opened 2026-05-05
[codex] Support transformers 5

lintool:
## Summary

- Require `transformers>=5,<6` in `pyproject.toml`.
- Fix CLIP image encoding and Aggretriever loading compatibility with Transformers 5.x.
- Lazy-load optional UniIR backend support so core MCP tools do not fail when `uniir_for_pyserini` is incompatible or absent.
- Split UniIR-backed MCP coverage into `tests/optional/`.

## Validation

- `mamba run -n pyserini-dev3 python -m unittest tests.optional.test_multimodal_search.TestMultimodalSearch`
- `mamba run -n pyserini-dev3 python -m unittest tests.core.test_mcp`
- `mamba run -n pyserini-dev3 python -m unittest tests.optional.test_mcp_uniir`
- `mamba run -n pyserini-dev3 python -m unittest tests.core.test_mcp tests.optional.test_mcp_uniir`
- `mamba run -n pyserini-dev3 python -m unittest tests.optional.test_encode.TestEncode.test_clip_encoder_cmd_image`

Note: the optional UniIR MCP test is discovered but skipped in `pyserini-dev3` when the installed `uniir_for_pyserini` is incompatible with `transformers 5.7.0`; it passes when the UniIR import succeeds because heavy index/model loading is mocked.

### DIFF
--- a/pyserini/encode/_aggretriever.py
+++ b/pyserini/encode/_aggretriever.py
@@ -26,12 +26,19 @@ if torch.cuda.is_available():
 from transformers import DistilBertConfig, BertConfig
 from transformers import AutoModelForMaskedLM, AutoTokenizer, PreTrainedModel
 from pyserini.encode import DocumentEncoder, QueryEncoder
+from pyserini.encode._base import load_head_weights
+from packaging.version import Version
+from transformers import __version__ as transformers_version
 
 
 class BertAggretrieverEncoder(PreTrainedModel):
     config_class = BertConfig
     base_model_prefix = 'encoder'
     load_tf_weights = None
+
+    @property
+    def all_tied_weights_keys(self):
+        return {}
 
     def __init__(self, config: BertConfig):
         super().__init__(config)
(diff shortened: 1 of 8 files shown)


lintool:
This worked on both my MacBook Air and `linux.cs`:

```
python -m unittest discover -s tests/core
```

`optional` tests fail because:

> `uniir_for_pyserini` is incompatible with transformers 5.7.0.

That's addressed by:

https://github.com/castorini/UniIR-for-Pyserini/pull/17

lintool:
@sahel-sh let's unbork the "core" tests first and then come back to fix "optional". thanks!

pyserini/encode/_aggretriever.py
```diff
@@ -26,13 +26,20 @@
...
+from packaging.version import Version
+from transformers import __version__ as transformers_version
 
 
 class BertAggretrieverEncoder(PreTrainedModel):
     config_class = BertConfig
     base_model_prefix = 'encoder'
     load_tf_weights = None
 
+    @property
```
sahel-sh, line 39: why is this needed? and why empty?

lintool: 🤷‍♂️ I'll get my agent on it.

pyserini/encode/_clip.py
```diff
@@ -80,7 +80,13 @@ def encode(self, paths, **kwargs):
         with torch.no_grad():
             image_features = self.model.get_image_features(**inputs)
 
-        embeddings = image_features.detach().cpu().numpy()
+        # In transformers 5.x, get_image_features() returns BaseModelOutputWithPooling
```
sahel-sh: Nit: code dup between text and image encoders, probably can move to base

sahel-sh (approved):
LGTM % Nit

pyserini/encode/_aggretriever.py
lintool: This is for Transformers 5.x compatibility. The new `from_pretrained()` loading path expects custom `PreTrainedModel` subclasses to expose `all_tied_weights_keys` and calls `.keys()` on it while finalizing missing/meta-loaded weights. Aggretriever does not define any additional tied-weight groups beyond the wrapped HF encoder, so the correct mapping for this wrapper is empty. This matches the compatibility shim already used by Pyserini’s other custom encoders, e.g., ANCE and UniCOIL.

PR #2500, opened 2026-05-06
Document results from @annagul on MRR@10 performance

annagulannamyradova-lang:
Added results reproduced by @annagul with specific metrics and environment details.


lintool:

PR #2501, opened 2026-05-06
Document results from @annagul for MRR and MAP metrics

annagulannamyradova-lang:
Added results reproduced by @annagul with metrics for MRR@10, MAP, and Recall@1000.


lintool:

PR #2502, opened 2026-05-06
Add reproduction results from @annagul

annagulannamyradova-lang:
Added reproduction results from @annagul on 2026-05-06.


lintool:

PR #2503, opened 2026-05-07
Add backpressure and caching for REST API

lilyjge:
## Summary

This PR adds lightweight overload control and caching to the REST API while keeping behavior aligned with current `/v1` semantics.

- Added simple **backpressure** for authenticated `/v1/*` requests:
  - rolling 60s latency window + per-key request counts
  - p99 latency trigger
  - shed busiest key(s) with HTTP `429`
- Added `--rest-p99-ms` CLI flag (default `750`) to control shedding threshold.
- Added lightweight **LRU caching** in `SharedSearchBackend` for REST-style repeated requests:
  - cached string-query `search()`
  - cached `get_document()`
- Updated REST docs/spec:
  - `429` documented in `openapi.yaml`
  - concise backpressure usage in `docs/usage-rest.md`
- Added REST tests for:
  - backpressure `429` behavior
  - OpenAPI `429` responses
  - repeated request cache hits for search/doc

## Performance snapshot

Using the included benchmark script (`scripts/rest_concurrency_benchmark.py`) with auth enabled:

- Original backpressure implementation showed a major throughput bottleneck.
- Lightweight backpressure (current implementation) recovers throughput close to master auth-only baseline:
  - at concurrency `8`: ~`906.81 req/s` (lightweight) vs ~`941.80 req/s` (master auth-only), ~`-3.7%`

### DIFF
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -64,10 +64,19 @@ python -m pyserini.server.rest --config /path/to/server.yaml --no-prebuilt-index
 
 When `--no-prebuilt-indexes` is set, the server only accepts index names declared under `indexes:` in `--config`.
 
-With `api_keys` in `--config`, **`--rest-p99-ms`** sets the latency threshold (milliseconds, default **750**) for simple load shedding: if rolling p99 over the last minute is above it, the busiest API key(s) may get **429**. Omitting `api_keys` disables this (and auth) on `/v1/*`.
+With `api_keys` in `--config`, **`--load-shedding-threshold`** sets the latency threshold (milliseconds, default **750**) for simple load shedding: if rolling p99 over the last minute is above it, the busiest API key(s) may get **429**. Omitting `api_keys` disables this (and auth) on `/v1/*`.
 
 ```bash
-python -m pyserini.server.rest --config /path/to/server.yaml --rest-p99-ms 500
+python -m pyserini.server.rest --config /path/to/server.yaml --load-shedding-threshold 500
+```
+
+The backend uses LRU caching for repeated requests:
+
+- **`--search-cache-size`** (default: **2048**): cache size for search results with string queries
+- **`--document-cache-size`** (default: **4096**): cache size for document fetches by docid
+
+```bash
+python -m pyserini.server.rest --search-cache-size 4096 --document-cache-size 8192
 ```
 
 ### Logging
@@ -208,7 +217,7 @@ curl -H "X-API-Key: {api-key}" \
 | 200 | Success |
 | 400 | Invalid parameters (e.g. missing `query`, invalid `hits` or `parse`), or cannot open index |
 | 401 | Missing or invalid API credential (when `api_keys` is configured) |
-| 429 | Load shedding (with `api_keys`): p99 over `--rest-p99-ms` |
+| 429 | Load shedding (with `api_keys`): p99 over `--load-shedding-threshold` |
 | 404 | Unknown route, or document not found for `GET .../doc/{docid}` |
 | 405 | Method not allowed (only **GET** is supported on these routes) |
 | 500 | Unhandled server error |
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -66,7 +66,7 @@ When `--no-prebuilt-indexes` is set, the server only accepts index names declare
 
 To limit deployment load, enable adaptive shedding when starting the server:
 
-- `--adaptive-shedding` turns on overload detection (host load and rolling request latency) and, when hot, temporarily rejects the busiest API key in a recent window with `429`.
+- `--adaptive-shedding` turns on overload detection from rolling **request latency** (p99 in a recent window) and, when high, temporarily rejects the busiest API key in that window with `429`.
 - `--adaptive-retry-after-seconds <int>` (optional) sets the `Retry-After` header on those responses (default: 1).
 
 ### Logging
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -64,6 +64,12 @@ python -m pyserini.server.rest --config /path/to/server.yaml --no-prebuilt-index
 
 When `--no-prebuilt-indexes` is set, the server only accepts index names declared under `indexes:` in `--config`.
 
+To limit deployment load, use the following options when starting the server to apply backpressure:
+
+- `--max-concurrent-per-key <int>` enables keyed backpressure (with `--max-queue-per-key`) by capping in-flight `/v1/*` requests per key.
+- `--max-queue-per-key <int>` caps queued + in-flight `/v1/*` requests per key; requests above this bound are rejected with `429`.
+- `--backpressure-retry-after-seconds <int>` sets the `Retry-After` response header for those `429` rejections.
+
 ### Logging
 
 REST server logging options:
@@ -78,6 +84,8 @@ Example:
 python -m pyserini.server.rest \
   --config /path/to/server.yaml \
   --no-prebuilt-indexes \
+  --max-concurrent-per-key 8 \
+  --max-queue-per-key 64 \
   --server-log-file logs/rest.server.log \
   --auth-log-file logs/rest.auth.log
 ```
@@ -204,6 +212,7 @@ curl -H "X-API-Key: {api-key}" \
 | 401 | Missing or invalid API credential (when `api_keys` is configured) |
 | 404 | Unknown route, or document not found for `GET .../doc/{docid}` |
 | 405 | Method not allowed (only **GET** is supported on these routes) |
+| 429 | Too many queued requests for a key (when keyed backpressure is enabled) |
 | 500 | Unhandled server error |
 
 The full list of operations, parameters, and response schemas is in **`/openapi.yaml`**.
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -64,7 +64,7 @@ python -m pyserini.server.rest --config /path/to/server.yaml --no-prebuilt-index
 
 When `--no-prebuilt-indexes` is set, the server only accepts index names declared under `indexes:` in `--config`.
 
-With `api_keys` in `--config`, **`--load-shedding-threshold`** sets the latency threshold (milliseconds, default **2000**) for simple load shedding: if rolling p99 over the last minute is above it, the busiest API key(s) may get **429**. Omitting `api_keys` disables this (and auth) on `/v1/*`.
+With `api_keys` in `--config`, **`--load-shedding-threshold`** sets the latency threshold (milliseconds, default **3000**) for simple load shedding: if rolling p99 over the last minute is above it, the busiest API key(s) may get **429**. Omitting `api_keys` disables this (and auth) on `/v1/*`.
 
 ```bash
 python -m pyserini.server.rest --config /path/to/server.yaml --load-shedding-threshold 500
@@ -217,7 +217,7 @@ curl -H "X-API-Key: {api-key}" \
 | 200 | Success |
 | 400 | Invalid parameters (e.g. missing `query`, invalid `hits` or `parse`), or cannot open index |
 | 401 | Missing or invalid API credential (when `api_keys` is configured) |
-| 429 | Load shedding (with `api_keys`): p99 over `--load-shedding-threshold` |
+| 429 | Load shedding (with `api_keys`): p99 over `--load-shedding-threshold`; body suggests retry timing; `Retry-After` may be set |
 | 404 | Unknown route, or document not found for `GET .../doc/{docid}` |
 | 405 | Method not allowed (only **GET** is supported on these routes) |
 | 500 | Unhandled server error |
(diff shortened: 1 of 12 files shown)


ronakice:
Can we bypass our own keys (we don't need to hold ourselves to the same standards as others ;))

lintool:
> Can we bypass our own keys (we don't need to hold ourselves to the same standards as others ;))

Actually, I disagree... ;)

I want safeguards to make sure we don't knock over the server ourselves either?

lintool:
@lingwei-gu @sahel-sh can you take a look also?

docs/usage-rest.md
```diff
@@ -64,6 +64,12 @@ python -m pyserini.server.rest --config /path/to/server.yaml --no-prebuilt-index
 
 When `--no-prebuilt-indexes` is set, the server only accepts index names declared under `indexes:` in `--config`.
 
+With `api_keys` in `--config`, **`--rest-p99-ms`** sets the latency threshold (milliseconds, default **750**) for simple load shedding: if rolling p99 over the last minute is above it, the busiest API key(s) may get **429**. Omitting `api_keys` disables this (and auth) on `/v1/*`.
```
lintool: How about something like `--load-shedding-threshold`? `ms` can be a metavar, so doesn't need to be in arg name?

docs/usage-rest.md
```diff
@@ -202,6 +208,7 @@ curl -H "X-API-Key: {api-key}" \
 | 200 | Success |
 | 400 | Invalid parameters (e.g. missing `query`, invalid `hits` or `parse`), or cannot open index |
 | 401 | Missing or invalid API credential (when `api_keys` is configured) |
+| 429 | Load shedding (with `api_keys`): p99 over `--rest-p99-ms` |
```
lintool: same, make sure to propagate changes...

pyserini/server/rest/app.py
```diff
@@ -56,6 +59,108 @@
 ROUTE_ERROR = 'Expected route /v1/{index}/search or /v1/{index}/doc/{docid}'
 AUTH_TOKEN_REQUEST_EMAIL = 'get-pyserini@googlegroups.com'
 
+
+class RestBackpressure:
+    """Rolling p99 latency vs. threshold; shed callers tied for max /v1/ hits in the last minute."""
+
+    __slots__ = (
```
lintool, line 87: can you document what each of these means?

pyserini/server/rest/app.py
```diff
@@ -56,6 +59,108 @@
...
+        '_max_count',
+        '_cached_p99_ms',
+        '_cached_p99_at',
+    )
+    _window_sec = 60.0
+    _min_latency_samples = 20
+    _p99_refresh_sec = 0.25
+
+    def __init__(self, p99_threshold_ms: float) -> None:
+        self._p99_threshold_ms = float(p99_threshold_ms)
```
lintool: the threshold can just be a private var, no? does it need to be in slots?

pyserini/server/rest/app.py
```diff
@@ -56,6 +59,108 @@
...
+        self._lock = threading.Lock()
+        self._latencies: deque[tuple[float, float]] = deque()
+        self._key_hits: deque[tuple[float, str]] = deque()
+        self._key_counts: dict[str, int] = {}
+        self._max_count = 0
+        self._cached_p99_ms: float | None = None
+        self._cached_p99_at = 0.0
+
+    def _prune_latencies(self, now: float) -> None:
```
lintool, line 111: maybe in the top level docs, can you explain what the algo is supposed to do?

pyserini/server/backend.py
```diff
@@ -100,6 +101,8 @@ def __init__(self, config_path: str | None = None, *, no_prebuilt_indexes: bool
         # Serialize get-or-create per logical index name so different indexes can open in parallel.
         self._index_lock_registry = threading.Lock()
         self._index_locks: dict[str, threading.Lock] = {}
+        self._search_cached = lru_cache(maxsize=2048)(self._search_impl)
```
lintool: should we make these cache sizes configurable?

scripts/rest_concurrency_benchmark.py
```diff
@@ -0,0 +1,152 @@
...
+from pathlib import Path
+
+import requests
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description='Benchmark concurrent REST search throughput.')
+    parser.add_argument('--url', default='http://127.0.0.1:8081', help='REST server base URL.')
+    parser.add_argument('--index', default='msmarco-v1-passage', help='Index name in /v1/{index}/search.')
+    parser.add_argument(
```
lintool: let's have everything be on one line, or every arg be multi-line... but not a mix of both?

looks odd.

lintool (changes requested):
Initial round of comments...

pyserini/server/rest/app.py
lilyjge: but `__slots__` restricts instances to only the defined variables

lintool: ok

lintool:
@sahel-sh @lingwei-gu can you take a look also?

pyserini/server/rest/app.py
```diff
@@ -172,6 +289,9 @@ def create_app(
     config_path: str | None = None,
     *,
     no_prebuilt_indexes: bool = False,
+    load_shedding_threshold_ms: float = 750.0,
```
lingwei-gu: We can prob increase the threshold a bit. Like 2s?

pyserini/server/rest/app.py
```diff
@@ -172,6 +289,9 @@ def create_app(
...
+    search_cache_size: int = 2048,
```
lingwei-gu, line 302: Is this for the lru_cache size? Could be larger. Lru_cache wouldn't take much memory unless we return the raw document

pyserini/server/backend.py
```diff
@@ -513,6 +534,23 @@ def search(
...
+        qid: str = '',
+        parse: bool = True,
+        allow_local_index: bool = True,
+        ef_search: int | None = None,
+        encoder: str | None = None,
+        query_generator: str | None = None,
+    ) -> dict[str, Any]:
+        # Cache only REST-style string queries; multimodal dict payloads stay uncached.
+        if isinstance(query, str):
+            return self._search_cached(query, index_name, hits, qid, parse, allow_local_index, ef_search, encoder, query_generator)
```
lingwei-gu, line 551: If we just use it as an annotation it should be enough. Like @lru_cache at the top of the function

pyserini/server/backend.py
```diff
@@ -122,6 +132,8 @@ def close_all(self) -> None:
             except Exception:
                 logger.warning('Failed to close searcher for index %s during backend shutdown.', config.name, exc_info=True)
         self.indexes.clear()
+        self._search_cached.cache_clear()
```
lingwei-gu, line 135: In this case if two indexes are queried alternatively at the same time, then the cache would be invalidated? We could use index name as part of the cache key

pyserini/server/backend.py
lilyjge: the actual search implementation is shared between REST API and MCP, and it accepts a dictionary for the sake of MCP multimodal search, which is unhashable for the cache annotation directly, which is why we need to split it into the two cases here and only apply cache for string queries

pyserini/server/backend.py
```diff
@@ -20,6 +20,7 @@
 import os
 import traceback
 import threading
```
lingwei-gu, line 22: Just curious. Would we support batch endpoints? Saw that Pyserini supports batch_search. This could further lower QPS down

pyserini/server/rest/app.py
lilyjge: yep, wdyt would be a reasonable number? but also the search tool does return raw documents

pyserini/server/backend.py
lingwei-gu: Ah makes sense. Thanks for clarification!

pyserini/server/backend.py
lilyjge: yeah that would be easy to implement, is this something we want @lintool ?

pyserini/server/rest/app.py
lingwei-gu: I see thought only document search returns raw document. If search also returns then let's keep it as is then. Usually I use 5-10% of memory as cache but also depending on use cases

pyserini/server/backend.py
lingwei-gu: Just realized we clear the whole index... in that case it's fine to clear the cache...

lingwei-gu (approved):

pyserini/server/backend.py
lintool: Let's circle back on that? I don't want to over-engineer... let's keep the API simple for now?

pyserini/server/rest/app.py
```diff
@@ -56,6 +60,119 @@
...
+            self._key_counts[key_id] = self._key_counts.get(key_id, 0) + 1
+            self._max_count = max(self._max_count, self._key_counts[key_id])
+            self._prune_key_hits(now)
+            self._prune_latencies(now)
+            p99 = self._p99_ms(now)
+            if p99 is None:
+                return False
+            if p99 <= self._load_shedding_threshold_ms:
+                return False
+            return self._key_counts.get(key_id, 0) == self._max_count and self._max_count >= 2
```
sahel-sh, line 164: hm, this wouldn't catch cases where too many api keys are each sending a single request.  Do we have any cap on the max qps that we can handle? regardless which key(s) they are coming from.

pyserini/server/rest/app.py
```diff
@@ -172,6 +289,9 @@ def create_app(
     config_path: str | None = None,
     *,
     no_prebuilt_indexes: bool = False,
+    load_shedding_threshold_ms: float = 2000.0,
```
sahel-sh: hm, is this tested, is there any latency profiling? I remember requests that took more than 2 seconds when projecting on climbmix.
I think this should be corpus dependent, like fineweb vs climbmix should have different tresholds?

pyserini/server/rest/app.py
```diff
@@ -239,18 +362,27 @@ async def rest_api_key_and_access_log(request: Request, call_next):
...
-            response.status_code,
-        )
-        return response
+        bp: RestBackpressure | None = getattr(request.app.state, 'rest_backpressure', None)
+        t0 = time.perf_counter()
+        if bp is not None and bp.should_shed(key_id, t0):
+            _log_auth_attribution('auth_request', client, request, query, key_id, 429)
+            return JSONResponse(
+                status_code=429,
+                content={'error': 'Service temporarily overloaded; retry later.'},
```
sahel-sh: nit: tell how much later.

scripts/rest_concurrency_benchmark.py
```diff
@@ -0,0 +1,144 @@
...
+        response = get_session().get(
+            search_url,
+            params={'query': query, 'hits': args.hits, 'parse': args.parse},
+            timeout=args.timeout_sec,
+        )
+        return response.status_code
+
+    def run_once(workers: int) -> tuple[float, float, int, dict[str, int]]:
+        t0 = time.perf_counter()
+        with ThreadPoolExecutor(max_workers=workers) as pool:
```
sahel-sh, line 68: With Python's GIL, this is not a true concurrency. needs multiprocessing?

pyserini/server/rest/app.py
```diff
@@ -239,18 +362,27 @@ async def rest_api_key_and_access_log(request: Request, call_next):
...
+            )
+
+        status = 500
+        try:
+            response = await call_next(request)
+            status = response.status_code
+            return response
+        except Exception:
+            raise
```
sahel-sh, line 400: Nit: should this be logged in before raise?

pyserini/server/rest/app.py
```diff
@@ -239,18 +362,27 @@ async def rest_api_key_and_access_log(request: Request, call_next):
...
+        finally:
+            if bp is not None:
+                now = time.perf_counter()
+                bp.record_latency((now - t0) * 1000.0, now)
```
sahel-sh: hm, I think this should only measure the successful cases, raised exception can artificially keep this low. What possible cases are captured in this generic raise?

lintool (approved):

pyserini/server/rest/app.py
lilyjge: no, @lingwei-gu suggested the number... I'll put in issue for iteration

lilyjge: changed to 3 seconds for now

pyserini/server/rest/app.py
lilyjge: I will leave this for now and address in issue, possibly related to broader qps thresholds

pyserini/server/rest/app.py
lilyjge: there really shouldn't be much that gets here, just a precaution I think

scripts/rest_concurrency_benchmark.py
lilyjge: this is close enough to produce roughly concurrent results for now, I can try benchmarking with multiprocessing in next iteration

PR #2517, opened 2026-05-09
[codex] Fix Transformers 5 encoder loading

lintool:
## Summary

- Fix ANCE loading with Transformers 5 by forcing PyTorch checkpoint loading and supporting prefixed ANCE head keys.
- Fix SPLADE loading for PyTorch-only checkpoints by disabling noisy safetensors conversion probing during model load.
- Add SPLADE encoder coverage and ANCE prefixed-head loading coverage.
- Disable a brittle ECIR 2023 optional paper test whose scores no longer match exactly after the Transformers 5 upgrade.

## Validation

- `python -m unittest tests.core.test_encode_base`
- `python -m unittest tests.base.test_encoder_model_ance`
- `python -m unittest tests.base.test_encoder_model_splade`
- `python -m py_compile pyserini/encode/_ance.py pyserini/encode/_splade.py tests/core/test_encode_base.py tests/base/test_encoder_model_splade.py`

### DIFF
--- a/integrations/optional/papers/test_ecir2023.py
+++ b/integrations/optional/papers/test_ecir2023.py
@@ -82,15 +82,16 @@ class TestECIR2023(unittest.TestCase):
     def test_section5_sub2_second(self):
         """Sample code of the second command in Section 5.2."""
 
-        cmd_nq = 'python scripts/repro_matrix/run_all_odqa.py --topics nq'
-        cmd_tqa = 'python scripts/repro_matrix/run_all_odqa.py --topics nq'
+        # 2026/05/09 - Removing this test and leaving a tombstone in place.
+        # The ECIR 2023 PyGaggle paper references these two commands in Section 5.2:
+        #
+        # python scripts/repro_matrix/run_all_odqa.py --topics nq
+        # python scripts/repro_matrix/run_all_odqa.py --topics nq
+        #
+        # Given the upgrade to Transformers 5, the scores no long match exactly (minor differences).
+        # Thus, it makes no sense to continue running this test.
 
-        # run both commands, check if all tests passed (i.e., returned OK)
-        stdout_nq, stderr_nq = run_command(cmd_nq)
-        self.assertEqual(stdout_nq.count('[OK]'), 21)
-
-        stdout_tqa, stderr_tqa = run_command(cmd_tqa)
-        self.assertEqual(stdout_tqa.count('[OK]'), 21)
+        pass
 
     def tearDown(self):
         clean_files(self.temp_files)
--- a/pyserini/encode/_ance.py
+++ b/pyserini/encode/_ance.py
@@ -14,15 +14,22 @@
 # limitations under the License.
 #
 
-from typing import Optional, List
+import os
+from typing import List, Optional
 
 import torch
-from transformers import PreTrainedModel, RobertaConfig, RobertaModel, RobertaTokenizer, requires_backends
+from packaging.version import Version
+from transformers import (
+    PreTrainedModel,
+    RobertaConfig,
+    RobertaModel,
+    RobertaTokenizer,
+    requires_backends
+)
+from transformers import __version__ as transformers_version
 
 from pyserini.encode import DocumentEncoder, QueryEncoder
 from pyserini.encode._base import load_head_weights
-from packaging.version import Version
-from transformers import __version__ as transformers_version
 
 
 class AnceEncoder(PreTrainedModel):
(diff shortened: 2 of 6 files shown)


pyserini/encode/_ance.py
```diff
@@ -78,11 +85,19 @@ def forward(self, input_ids: torch.Tensor, attention_mask: Optional[torch.Tensor
     
     @classmethod
     def load_pretrained_encoder(cls, model_name_or_path: str, device: str):
-        model = cls.from_pretrained(model_name_or_path)
+        previous_disable_conversion = os.environ.get('DISABLE_SAFETENSORS_CONVERSION')
+        os.environ['DISABLE_SAFETENSORS_CONVERSION'] = '1'
+        try:
+            model = cls.from_pretrained(model_name_or_path, use_safetensors=False)
```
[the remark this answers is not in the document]
lintool: These are old models we have no control over, so this is fine.

pyserini/encode/_splade.py
```diff
@@ -13,13 +13,32 @@
...
 
-from pyserini.encode import QueryEncoder, DocumentEncoder
+from pyserini.encode import DocumentEncoder, QueryEncoder
+
+
+def _from_pretrained_without_safetensors_conversion(model_name_or_path, **kwargs):
+    # Some SPLADE checkpoints, such as naver/splade-v3, only publish PyTorch
+    # weights. Newer Transformers versions still load them successfully, but
+    # then spawn a background safetensors conversion thread that can fail noisily.
+    kwargs.setdefault('use_safetensors', False)
```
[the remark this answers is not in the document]
lintool: These are old models we have no control over, so this is fine.

integrations/optional/papers/test_ecir2023.py
```diff
@@ -82,15 +82,16 @@ def test_section5_sub2_first(self):
     def test_section5_sub2_second(self):
         """Sample code of the second command in Section 5.2."""
 
-        cmd_nq = 'python scripts/repro_matrix/run_all_odqa.py --topics nq'
```
sahel-sh, line 85: seems like the correct fix is to update the script with new scores instead of removing it.

pyserini/encode/_ance.py
```diff
@@ -78,11 +85,19 @@ def forward(self, input_ids: torch.Tensor, attention_mask: Optional[torch.Tensor
...
+        finally:
+            if previous_disable_conversion is None:
+                del os.environ['DISABLE_SAFETENSORS_CONVERSION']
+            else:
+                os.environ['DISABLE_SAFETENSORS_CONVERSION'] = previous_disable_conversion
         if Version(transformers_version) >= Version("5.0.0"):
             load_head_weights(model, model_name_or_path, {
-                'embeddingHead': 'embeddingHead',
-                'norm': 'norm'
+                'embeddingHead': ['ance_encoder.embeddingHead', 'embeddingHead'],
```
sahel-sh, line 99: This was just changed in the previous pr, without an actual regression test with 2cr, it is really hard to verify the correctness.

pyserini/encode/_splade.py
```diff
@@ -89,30 +107,26 @@ def encode(self, texts, titles=None, max_length=512, add_sep=False, **kwargs) ->
...
-                                               * input_attention.unsqueeze(-1), dim=1)
+        batch_aggregated_logits, _ = torch.max(torch.log(1 + torch.relu(batch_logits)) * input_attention.unsqueeze(-1), dim=1)
         batch_aggregated_logits = batch_aggregated_logits.cpu().detach().numpy()
         raw_weights = self._output_to_weight_dicts(batch_aggregated_logits)
         return self._get_encoded_query_token_wight_dicts(raw_weights)
 
 
 class SpladeQueryEncoder(QueryEncoder, SpladeEncoder):
-    def __init__(self, model_name_or_path, tokenizer_name=None, device='cpu'):
+    def __init__(self, model_name_or_path, tokenizer_name=None, device='cpu', **kwargs):
```
sahel-sh: why is **kwargs  passed in init?
_from_pretrained_without_safetensors_conversion can handle the 'use_safetensors' similar to line 91 in _ance

integrations/optional/papers/test_ecir2023.py
lintool: I went back and forth on this...

This test case is supposed to capture the output of commands in an ECIR 2023 paper... yes, we can certainly update with new scores, but then it would no longer reproduce the expected results reported in the paper... which would defeat the purpose of having this test case to begin with...

thoughts?

pyserini/encode/_ance.py
lintool: yes, I've run the various tests and this was the error that was surfaced

lintool:
hi @sahel-sh I have better aligned `_splade.py` and `_ance.py` - they're the same underlying issues. wdyt?

integrations/optional/papers/test_ecir2023.py
sahel-sh: if the results are not reproducible anyway, I am fine with removing it

sahel-sh (approved):
LGTM, thanks

PR #2522, opened 2026-05-14
[codex] Refactor prebuilt index size metadata (for Anserini v2.1.0)

lintool:
## Summary

- Rename prebuilt index metadata size field from `"size compressed (bytes)"` to `"size"`.
- Update download and verification utilities to read the new size metadata key.
- Update tests and BEIR metadata generation to match the refactored metadata shape.
- Use the resolved Lucene flat searcher index directory in the dense search test.

## Validation

- `mamba run -n pyserini-dev3 python -m unittest tests.base.test_prebuilt_index.TestPrebuiltIndexes`

Validation passed: `Ran 25 tests in 113.449s`, `OK`.

### DIFF
--- a/pyserini/prebuilt_index_info.py
+++ b/pyserini/prebuilt_index_info.py
@@ -18,18 +18,17 @@ from pyserini.pyclass import autoclass
 
 JPrebuiltInvertedIndex = autoclass('io.anserini.index.prebuilt.PrebuiltInvertedIndex')
 
+
 def import_from_inverted_lucene(index_metadata):
     info = {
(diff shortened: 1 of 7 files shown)


lintool:
@lilyjge @sahel-sh This is refactoring for Anserini v2.1.0 to bring up to date with new metadata field.

"size compressed (bytes)" has be renamed to "size" because new indexes are just now tarred, not gzipped; discovered that compression was very slow and not worthwhile.

lilyjge (approved):

PR #2531, opened 2026-05-18
Add BM25 k1 and b configuration for REST API and MCP Server

lilyjge:
Inspired by https://arxiv.org/abs/2605.10848

## Summary

Adds optional BM25 tuning (`k1`, `b`) to the Pyserini REST and MCP search APIs for sparse (TF) indexes. Parameters must be sent together (matching Pyserini Lucene CLI behaviour); omit both to use Anserini defaults (0.9 / 0.4). Custom settings use a thread-safe pool of immutable `LuceneSearcher` instances so the default searcher is never mutated under concurrency.

### API
- **REST:** `GET /v1/{index}/search?k1=...&b=...` — query parsing, validation, and 400s for invalid or partial params; OpenAPI updated.
- **MCP:** `search` tool accepts optional `k1` and `b`, forwarded through `SharedSearchBackend.search`.
- **Docs:** `usage-rest.md` and `usage-mcp.md` document sparse-only behavior and “omit both” semantics.

### Backend (`SharedSearchBackend`)
- Central validation: paired params, finite numbers, `k1 ≥ 0`, `b ∈ [0, 1]`.
- Canonical rounding (6 decimals) for cache keys and variant pool keys.
- Default path: omit both (or explicit 0.9 / 0.4) → shared index searcher.
- Custom path: per-`(k1, b)` searcher with an active-use count, LRU eviction (default pool size 4), and safe close on eviction/shutdown.
- LRU search cache includes `bm25_k1` / `bm25_b` so different settings do not share cached results.
- Rejects `k1` / `b` on non-`tf` index types with a clear error.

### Tests
- REST: acceptance, validation (partial params, bounds, NaN, non-numeric), score differences, cache isolation, non-sparse rejection, variant pool reuse/eviction.
- MCP: happy path, `k1`-only / `b`-only errors, custom params change ranking line.

### DIFF
--- a/docs/usage-mcp.md
+++ b/docs/usage-mcp.md
@@ -137,8 +137,8 @@ Runs retrieval against a Pyserini index (BM25 for standard sparse indexes; dense
 | `ef_search` | `int` | no | `100` | HNSW / dense Lucene HNSW search parameter. |
 | `encoder` | `str` | no | `""` | Encoder id when the index requires one (dense / FAISS / etc.). |
 | `query_generator` | `str` | no | `""` | Sparse (TF) indexes only: `BagOfWords`, `DisjunctionMax` / `dismax`, `QuerySideBm25` / `bm25qs`, `Covid19`. Omit for downstream default (`BagOfWords`). |
-| `k1` | `float \| null` | no | `0.9` | BM25 k1 for sparse (TF) indexes. Must be finite, non-negative, and set together with `b`. |
-| `b` | `float \| null` | no | `0.4` | BM25 b for sparse (TF) indexes. Must be finite, in `[0, 1]`, and set together with `k1`. |
+| `k1` | `float \| null` | no | — (omit both) | BM25 k1 for sparse (TF) indexes. Must be finite, non-negative, and set together with `b`. Omit both `k1` and `b` for Anserini defaults (0.9 / 0.4). |
+| `b` | `float \| null` | no | — (omit both) | BM25 b for sparse (TF) indexes. Must be finite, in `[0, 1]`, and set together with `k1`. Omit both for Anserini defaults (0.9 / 0.4). |
 
 **Returns:** `list[Any]` rich content parts, intentionally returned as a list so MCP clients can render mixed text + images correctly.
 Typical layout is:
--- a/docs/usage-mcp.md
+++ b/docs/usage-mcp.md
@@ -137,8 +137,8 @@ Runs retrieval against a Pyserini index (BM25 for standard sparse indexes; dense
 | `ef_search` | `int` | no | `100` | HNSW / dense Lucene HNSW search parameter. |
 | `encoder` | `str` | no | `""` | Encoder id when the index requires one (dense / FAISS / etc.). |
 | `query_generator` | `str` | no | `""` | Sparse (TF) indexes only: `BagOfWords`, `DisjunctionMax` / `dismax`, `QuerySideBm25` / `bm25qs`, `Covid19`. Omit for downstream default (`BagOfWords`). |
-| `k1` | `float \| null` | no | `0.9` | BM25 k1 for sparse (TF) indexes. Must be set together with `b`. |
-| `b` | `float \| null` | no | `0.4` | BM25 b for sparse (TF) indexes. Must be set together with `k1`. |
+| `k1` | `float \| null` | no | `0.9` | BM25 k1 for sparse (TF) indexes. Must be finite, non-negative, and set together with `b`. |
+| `b` | `float \| null` | no | `0.4` | BM25 b for sparse (TF) indexes. Must be finite, in `[0, 1]`, and set together with `k1`. |
 
 **Returns:** `list[Any]` rich content parts, intentionally returned as a list so MCP clients can render mixed text + images correctly.
 Typical layout is:
--- a/docs/usage-mcp.md
+++ b/docs/usage-mcp.md
@@ -137,8 +137,8 @@ Runs retrieval against a Pyserini index (BM25 for standard sparse indexes; dense
 | `ef_search` | `int` | no | `100` | HNSW / dense Lucene HNSW search parameter. |
 | `encoder` | `str` | no | `""` | Encoder id when the index requires one (dense / FAISS / etc.). |
 | `query_generator` | `str` | no | `""` | Sparse (TF) indexes only: `BagOfWords`, `DisjunctionMax` / `dismax`, `QuerySideBm25` / `bm25qs`, `Covid19`. Omit for downstream default (`BagOfWords`). |
-| `k1` | `float \| null` | no | `0.9` | BM25 k1 for sparse (TF) indexes (Anserini default). Omit to keep the default. |
-| `b` | `float \| null` | no | `0.4` | BM25 b for sparse (TF) indexes (Anserini default). Omit to keep the default. |
+| `k1` | `float \| null` | no | `0.9` | BM25 k1 for sparse (TF) indexes. Must be set together with `b`. |
+| `b` | `float \| null` | no | `0.4` | BM25 b for sparse (TF) indexes. Must be set together with `k1`. |
 
 **Returns:** `list[Any]` rich content parts, intentionally returned as a list so MCP clients can render mixed text + images correctly.
 Typical layout is:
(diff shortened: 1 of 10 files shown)


pyserini/server/backend.py
```diff
@@ -119,22 +155,71 @@ def _lock_for_index_name(self, index_name: str) -> threading.Lock:
...
+            self._close_index_config(config)
         self.indexes.clear()
         self._search_cached.cache_clear()
         self._document_cached.cache_clear()
 
+    def _acquire_bm25_searcher(
+        self,
+        index_name: str,
+        config: IndexConfig,
+        bm25: tuple[float, float],
```
lintool: I don't think this is a good name for the params?

lintool: do you think it'd be better to take something more descriptive like config, as a JSON, something like:

`{ "k1": 0.9, "b": 0.4 }`

so it'd be more future-proof for other BM25 variants if we decide to have them?

pyserini/server/backend.py
```diff
@@ -119,22 +155,71 @@ def _lock_for_index_name(self, index_name: str) -> threading.Lock:
...
+    ) -> tuple[LuceneSearcher, tuple[float, float]]:
+        to_close: list[LuceneSearcher] = []
+        try:
+            with self._lock_for_index_name(index_name):
+                slot = config.bm25_searchers.get(bm25)
+                if slot is None:
+                    while len(config.bm25_searchers) >= self._bm25_variant_cache_size:
```
lintool: not sure if `_variant` is the best name?

pyserini/server/backend.py
```diff
@@ -479,6 +566,8 @@ def _search_impl(
         ef_search: int | None = None,
         encoder: str | None = None,
         query_generator: str | None = None,
+        bm25_k1: float | None = None,
```
lintool: why inconsistent naming? `k1` vs `bm25_k1`?

pyserini/server/backend.py
```diff
@@ -545,11 +660,39 @@ def search(
...
+        k1: float | None = None,
+        b: float | None = None,
     ) -> dict[str, Any]:
+        _validate_bm25_params(k1, b)
+        bm25_k1, bm25_b = _canonical_bm25_params(k1, b)
         # Cache only REST-style string queries; multimodal dict payloads stay uncached.
         if isinstance(query, str):
-            return self._search_cached(query, index_name, hits, qid, parse, allow_local_index, ef_search, encoder, query_generator)
-        return self._search_impl(query, index_name, hits, qid, parse, allow_local_index, ef_search, encoder, query_generator)
+            return self._search_cached(
```
lintool: hrm... whenever an arg list gets so long, I kinda think... is there a better way?

pyserini/server/utils.py
```diff
@@ -35,6 +36,14 @@
 )
 
 
+@dataclass
+class Bm25SearcherSlot:
+    """Cached immutable LuceneSearcher for a specific BM25 setting."""
```
lintool: maybe we can standardize nomenclature? how about make everything consistent "config"?

lintool (changes requested):
comments...

lintool:
@ronakice @sahel-sh please take a look also?

pyserini/server/backend.py
```diff
@@ -89,6 +97,48 @@ def _norm_opt_str(value: str | None) -> str:
...
 
+def _validate_bm25_params(k1: float | None, b: float | None) -> None:
+    if (k1 is None) != (b is None):
+        raise BadSearchRequestError('BM25 parameters k1 and b must be set together')
+    try:
+        k1_value = float(k1) if k1 is not None else None
+        b_value = float(b) if b is not None else None
+    except (TypeError, ValueError) as e:
+        raise BadSearchRequestError('BM25 parameters k1 and b must be numbers') from e
+    if k1_value is not None and not math.isfinite(k1_value):
```
sahel-sh: The theoretical cap for k1 is infinity, why do yo check is finite?
 

nit: you can check some of this conditions together and merge the error messages: one error for k1 should be a non-negative finite number, the second one for b between [0 and 1], you don't need to check b being finite

pyserini/server/backend.py
```diff
@@ -472,34 +575,51 @@ def _search_impl(
...
-        else:
-            if query_generator and index_config.index_type == 'tf' and isinstance(index_config.searcher, LuceneSearcher):
-                jquery_gen = self._resolve_query_generator(query_generator, index_config.searcher)
-                results = index_config.searcher.search(query, hits, query_generator=jquery_gen)
+            results = self.sharded_search(query, options.hits, options.ef_search or 100, options.encoder or 'ArcticEmbedL')
+        elif index_config.index_type == 'tf' and isinstance(index_config.searcher, LuceneSearcher):
+            if bm25_config == Bm25Config(k1=BM25_DEFAULT_K1, b=BM25_DEFAULT_B):
+                bm25_config = None
+
+            qg_k1 = bm25_config.k1 if bm25_config is not None else BM25_DEFAULT_K1
```
sahel-sh, line 596: if you are handling b and k1 separately, why are you asking for the requests to specify both or none?

pyserini/server/backend.py
```diff
@@ -119,22 +171,71 @@ def _lock_for_index_name(self, index_name: str) -> threading.Lock:
...
+        index_name: str,
+        config: IndexConfig,
+        bm25_config: Bm25Config,
+    ) -> tuple[LuceneSearcher, Bm25Config]:
+        to_close: list[LuceneSearcher] = []
+        try:
+            with self._lock_for_index_name(index_name):
+                slot = config.bm25_searchers.get(bm25_config)
+                if slot is None:
+                    while len(config.bm25_searchers) >= self._bm25_searcher_cache_size:
```
sahel-sh, line 206: regardless of which config was used last, this always evicts the config with smallest k1,b tuple even when it is the most recently used one. Please check this from chatgpt:

https://chatgpt.com/share/6a0c6b6e-b524-83ea-b551-32c5fa3dacce

pyserini/server/backend.py
lilyjge: for consistency with current Pyserini Lucene search  https://github.com/castorini/pyserini/blob/master/pyserini/search/lucene/__main__.py#L37

pyserini/server/backend.py
lilyjge: the impl is already using `OrderedDict` for that: cache hits call `move_to_end(bm25_config)`, and eviction scans `config.bm25_searchers.items()` from the front, so the first idle entry found is the least recently used idle config. since `Bm25Config` is only used as the key, its tuple-like ordering does not affect eviction here.

added a test to make this clear!

sahel-sh: ah, had missed it, thanks for clarifying and for adding the test

lintool:
Actually, just occurred to me: why do we need a separate instance for each config setting? Why not just change the parameters each time before we call?

Sorry, did I miss something?

lilyjge:
that was my initial implementation, however the problem is that k1 and b modify the underlying searcher state, which creates a race condition. example from codex:
```
1. request A sets BM25 to `(k1=0.8, b=0.3)`
2. request B sets the same searcher to `(k1=1.2, b=0.75)`
3. request A’s search may now run with B’s parameters
````
also tried putting a lock around that, but that would serialize the requests and mess up concurrency. 

so that's why I went with the separate instance approach!

File: docs/usage-mcp.md [hunk 9]
```diff
@@ -137,6 +137,8 @@ Runs retrieval against a Pyserini index (BM25 for standard sparse indexes; dense
 | `ef_search` | `int` | no | `100` | HNSW / dense Lucene HNSW search parameter. |
 | `encoder` | `str` | no | `""` | Encoder id when the index requires one (dense / FAISS / etc.). |
 | `query_generator` | `str` | no | `""` | Sparse (TF) indexes only: `BagOfWords`, `DisjunctionMax` / `dismax`, `QuerySideBm25` / `bm25qs`, `Covid19`. Omit for downstream default (`BagOfWords`). |
+| `k1` | `float \| null` | no | — (omit both) | BM25 k1 for sparse (TF) indexes. Must be finite, non-negative, and set together with `b`. Omit both `k1` and `b` for Anserini defaults (0.9 / 0.4). |
+| `b` | `float \| null` | no | — (omit both) | BM25 b for sparse (TF) indexes. Must be in `[0, 1]`, and set together with `k1`. Omit both for Anserini defaults (0.9 / 0.4). |
```
ronakice, line 141: why are defaults grouped

docs/usage-rest.md
```diff
@@ -137,14 +137,21 @@ If the index cannot be opened, the API responds with **400** and a message such
...
 **Example**
 
 ```bash
 curl "http://localhost:8081/v1/msmarco-v1-passage/search?query=what%20is%20a%20lobster%20roll&hits=1"
 ```
 
+Custom BM25 parameters (sparse indexes only; both required together):
+
+```bash
+curl "http://localhost:8081/v1/cacm/search?query=information%20retrieval&hits=5&k1=0.8&b=0.3"
```
ronakice, line 152: why cacm here, does this work?

docs/usage-mcp.md
lilyjge: because k1 and b must be set together or neither set, for consistency with current Pyserini Lucene search https://github.com/castorini/pyserini/blob/master/pyserini/search/lucene/__main__.py#L37

docs/usage-rest.md
lilyjge: because cacm is small and used for running the REST API tests, so it does work!

lintool (approved):

PR #2545, opened 2026-05-26
Add document truncation option for REST API search tool 

lilyjge:
## Summary

Currently, the search tool returns the raw document contents of the top k retrieved results. This could blow-up context with really long documents and renders the get document tool effectively useless, because the search tool already returns full document contents. This PR adds an optional fix for this.

Adds a `max_doc_chars` query parameter to the REST search API so callers can request truncated document payloads for each search result.

When `max_doc_chars` is omitted, search responses preserve the existing behavior and return the full document payload. When provided, each candidate document payload is truncated to the first N characters.

## Details

- Added `max_doc_chars` parsing and validation to `GET /v1/{index}/search`
- Plumbed `max_doc_chars` through `SharedSearchBackend.search`
- Added document payload truncation helper
- Truncates string payloads directly
- For parsed dict payloads, truncates common text fields only: `body`, `content`, `contents`, and `text`
- Updated REST docs and OpenAPI schema
- Added tests for truncation behavior and invalid parameter handling

### DIFF
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -139,7 +139,7 @@ If the index cannot be opened, the API responds with **400** and a message such 
 | `parse` | no | `true` | If `true`, parse the stored `raw` field when it is JSON (see `format_lucene_document` / Anserini-style formatting); if `false`, return the raw stored string. |
 | `k1` | no | `0.9` | BM25 k1 for sparse (TF) indexes. Must be finite, non-negative, and sent together with `b`. |
 | `b` | no | `0.4` | BM25 b for sparse (TF) indexes. Must be in `[0, 1]`, and sent together with `k1`. |
-| `max_doc_chars` | no | — | Return only the first N characters of each candidate document payload. If omitted, the full document payload is returned. |
+| `max_doc_length` | no | — | Maximum characters to return for each parsed candidate document. If omitted, return the full document. Requires `parse=true`. |
 
 **Example**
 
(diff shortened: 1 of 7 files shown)


lintool (commented):
Clarification: are we truncating _only_ the doc content?

"payload" makes it confusing, because if we're not parsing, the payload is JSON, so if we truncate json, we might end up with something malformed?

We should probably document somewhere? What happens in the `parse={true,false}` code paths? With examples?

lintool:
Also, wouldn't something like `max_doc_length` be more canonical, and write in documentation that length is measured in chars?

lintool:
Yea, this is confusing:

> Truncates string payloads directly
> For parsed dict payloads, truncates common text fields only: body, content, contents, and text

Can we explain in docs somewhere?

lilyjge:
Yep, I will clarify in docs that truncation only applies if parse=true

> Also, wouldn't something like max_doc_length be more canonical, and write in documentation that length is measured in chars?

yeah makes sense, will change. current name is probably artifact from discussion I had with codex on whether to do token-based truncation vs char based since it's inconsistent across other existing codebases (ended up going with char based to not introduce additional tokenizer latency)

docs/usage-rest.md
```diff
@@ -152,6 +153,16 @@ Custom BM25 parameters (sparse indexes only; both required together):
...
 ```
 
+Limit document text returned for each hit:
+
+```bash
+curl "http://localhost:8081/v1/msmarco-v1-passage/search?query=what%20is%20a%20lobster%20roll&hits=5&max_doc_length=500"
+```
+
+`max_doc_length` is measured in characters. For parsed object documents, only `body`, `content`,
+`contents`, and `text` are limited. Other fields are unchanged. Combining `parse=false` with
```
lintool:
```suggestion
`contents`, and `text` are truncated. Other fields are unchanged. Combining `parse=false` with
```

pyserini/server/document_format.py
```diff
@@ -69,3 +70,32 @@ def format_lucene_document(document: Document | None, parse: bool) -> Any:
...
+        return payload
+    if payload is None:
+        return None
+    if isinstance(payload, str):
+        return payload[:max_doc_length]
+    if isinstance(payload, dict):
+        keys_to_truncate = {k for k in payload if str(k).lower() in _TRUNCATABLE_FIELDS}
+        if keys_to_truncate:
+            return {
+                k: truncate_document_payload(v, max_doc_length=max_doc_length)
```
lintool: why is this function written recursively? like, why do we need recursion to call `payload[:max_doc_length]`?

lintool (changes requested):

pyserini/server/document_format.py
lilyjge: codex over generalized, fixed

lintool:
haha, much simpler once you got rid of needless recursion...

@codex wtf were you thinking

lintool (approved):

PR #2550, opened 2026-06-01
Add unified JSONL REST request logging

lilyjge:
basically, before we had two log file options, one for default uvicorn logs and one for the auth/access logs we're interested in
now it's just one log file with the interesting auth/access logs and uvicorn logs are discarded by default with a flag to opt in and send them to the same log file.
the auth/access logs are now in jsonl as well

codex:
## Summary
- replace separate REST auth/access logging with one JSONL request logger
- log auth failures, load shedding, anonymous requests, and successful requests in one stream
- simplify REST logging CLI to --log-file and opt-in --keep-uvicorn-logs
- update REST usage docs and request logging tests

## Tests
- python -m py_compile pyserini/server/rest/app.py tests/core/test_rest.py
- python -m unittest tests.core.test_rest

### DIFF
--- a/docs/usage-rest.md
+++ b/docs/usage-rest.md
@@ -88,9 +88,11 @@ REST server logging options:
   uvicorn access logs are disabled because the JSONL request log is the canonical access log.
   If combined with `--log-file`, uvicorn text access lines are appended to the same file.
 
-Each JSONL request record includes the timestamp, client, method, path, full query string,
-status, latency, auth outcome, and a non-reversible API-key fingerprint when credentials are present.
-Auth failures and load-shedding responses are written to the same log as successful requests.
+Each JSONL request record includes the timestamp, request ID, client, method, path, query string
+(capped at 1000 characters), status, latency, auth outcome, and a non-reversible API-key fingerprint
+when credentials are present. Auth failures and load-shedding responses are written to the same log as
+successful requests. The server generates a request ID for each request, logs it, and returns it as
+`X-Request-ID`.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


lintool:
hey @lilyjge - I'm not familiar with uvicorn logs... what typically goes there?

lilyjge:
I'm not super familiar either but it's basically the 'default' server logs you'd see with every request, which I'd noticed in the initial implementation to be a subset of the information contained in our custom logging. here's a more detailed explanation from codex:

Typically uvicorn has two kinds of logs:

`uvicorn.access`: one text line per HTTP response, roughly:
```text
127.0.0.1:54321 - "GET /v1/cacm/search?query=x HTTP/1.1" 200
```
This is the classic web-server access log: client, method, path/query, protocol, status. In our case it is now mostly redundant with the JSONL request log, which has the same core facts plus latency, auth outcome, key fingerprint, load-shed status, etc.

`uvicorn.error`: server/runtime logs, for example startup/shutdown messages, bind failures, lifespan errors, tracebacks, protocol warnings, and other uvicorn internals. Despite the name, it also carries normal server lifecycle messages.

We?re only disabling uvicorn **access logs** by default, not suppressing all uvicorn/server errors. The reason is that our JSONL middleware is now the canonical request log. Keeping uvicorn access logs on by default would duplicate every request in a different text format, and if directed into `--log-file`, it would make the file no longer valid pure JSONL.

So the default is:

```bash
python -m pyserini.server.rest --log-file logs/rest.requests.jsonl
```

This gives one structured request record per request, including auth failures.

If someone still wants uvicorn?s traditional access lines for local debugging or compatibility with existing log scraping, they can opt in:

```bash
python -m pyserini.server.rest \
  --log-file logs/rest.requests.jsonl \
  --keep-uvicorn-logs
```

Then they knowingly get mixed JSONL + uvicorn text access lines in the same file.

lingwei-gu:
overall looks good and safe for a logging behaviour PR. just curious do we have requestIds for incoming requests, and do we want to log these as well? @lilyjge

nourj98 (approved):
OK, I think this looks good, overall. 

The only minor question is if maybe it's still worth keeping the queries truncated to a max number of chars (like 1k) so our log files don't get to big, but I think this is a minor issue

nourj98:
Code review fixes look all good! I think this is ready to get merged in

PR #2582, opened 2026-06-24
[codex] Refactor Lucene prebuilt inverted index metadata loading

lintool:
## Summary

- Add lazy loaders for Lucene prebuilt inverted index metadata from remote and package-local JSON catalogs.
- Move Mr.TyDi, MIRACL, and CIRAL TF index metadata into local JSON resources.
- Preserve lazy aliases and existing registry counts for affected collections.

## Validation

- `mamba run -n pyserini-dev3 python -m py_compile pyserini/prebuilt_index_info.py`
- `mamba run -n pyserini-dev3 python -c "import json; json.load(open('pyserini/resources/prebuilt-indexes/lucene/mrtydi-inverted.json')); json.load(open('pyserini/resources/prebuilt-indexes/lucene/miracl-inverted.json')); json.load(open('pyserini/resources/prebuilt-indexes/lucene/ciral-inverted.json'))"`
- Focused no-network lookup checks for Mr.TyDi, MIRACL, and CIRAL
- Aggregate count checks for Mr.TyDi, MIRACL, and CIRAL


lintool:
hi @lilyjge @sahel-sh - this is an initial impl of https://github.com/castorini/anserini/discussions/3212

I want to do this in stages, not in all one huge PR - and test each step of the way.

I'm using `resources/prebuilt-indexes/lucene/` as a holding area for now because I don't want to thrash on external repo. Once everything lands I'll bulk move over to prebuilt-indexes repo.

I will test everything before merging, but you can proceed parallel w/ reviewing.

pyserini/prebuilt_index_info.py
```diff
@@ -186,633 +290,45 @@ def import_from_inverted_lucene(index_metadata):
...
-        "size": 94610259,
-        "total_terms": 13693080,
-        "documents": 82095,
-        "unique_terms": 236638,
-        "downloaded": False
-    },
+TF_INDEX_INFO_BEIR = _LazyLucenePrebuiltInvertedIndexJson(
+    'https://raw.githubusercontent.com/castorini/prebuilt-indexes/main/lucene/beir-inverted-flat.json',
+    'https://raw.githubusercontent.com/castorini/prebuilt-indexes/main/lucene/beir-inverted-multifield.json',
+    key_prefixes=('beir-v1.0.0-',)
```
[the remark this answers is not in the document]
lintool: External dependence on GitHub is fine.

lilyjge (commented):
I can +1 if we want to iterate and circle back, but I don't quite like the design of having the  `_MergedLucenePrebuiltInvertedIndexInfo` and `_AliasedLucenePrebuiltInvertedIndexInfo` classes. It seems like an extra layer of abstraction that doesn't buy us very much. Since we're also going to have other types of prebuilt indexes in the future, it would also be cumbersome to introduce new classes for each type.

lintool:
@lilyjge okay, let me react and refactor. Or rather, get my agent to think about it... :)

lintool:
@lilyjge I refactored this following your concern about the extra abstraction layer.

The merged/aliased Lucene-specific mapping classes are now replaced with a single generic `_PrebuiltIndexCatalog`. `TF_INDEX_INFO` is composed declaratively from source mappings plus an alias table, and aliases now point to canonical index names instead of storing copied metadata dicts. This keeps the behavior reusable for future prebuilt index types without adding one-off classes for each source/alias pattern.

I also added focused coverage for:
- alias lookup into static metadata
- alias lookup into lazy metadata
- avoiding eager loads for unrelated prefix-filtered sources
- iteration including aliases
- existing `TF_INDEX_INFO` alias compatibility

Validation:
- `mamba run -n pyserini-dev3 python -m unittest tests.base.test_prebuilt_index_info_catalog`
- `mamba run -n pyserini-dev3 python -m py_compile pyserini/prebuilt_index_info.py tests/base/test_prebuilt_index_info_catalog.py`
- `mamba run -n pyserini-dev3 python -m unittest tests.base.test_utils`

lintool:
@lilyjge this is what Codex came up with... lmk if you like it better...

pyserini/prebuilt_index_info.py
```diff
@@ -38,6 +43,109 @@ def import_from_inverted_lucene(index_metadata):
     return info
 
 
+def _import_from_lucene_prebuilt_inverted_index_json(index_metadata):
```
lilyjge, line 46: I know we're in an in-between state right now, but I think we can probably just replace the above `import_from_inverted_lucene` with this method? or maybe leave a note so we remember to consolidate the two?

pyserini/prebuilt_index_info.py
```diff
@@ -38,6 +43,109 @@ def import_from_inverted_lucene(index_metadata):
...
+    }
+
+    for field in ['total_terms', 'documents', 'unique_terms']:
+        if field in index_metadata:
+            info[field] = index_metadata[field]
+
+    return info
+
+
+class _LazyLucenePrebuiltInvertedIndexJson(Mapping):
```
lilyjge, line 64: not quite sure whether the name is appropriate since I think this class can be reused across different types of prebuilt indexes but we can handle that once we integrate the other types

lilyjge (approved):

lintool:
@lilyjge Let me merge and do another round of refactoring to collapse unnecessary functions, etc. Yes, I think you're right, but we might want to preserve distinctions between index types so that metadata and corresponding checks can diverge.

pyserini/prebuilt_index_info.py
lintool: See below.

pyserini/prebuilt_index_info.py
lintool: See below.

PR #2584, opened 2026-06-24
[codex] Refactor archive download utility

lintool:
## Summary

- Rename the download/unpack helper and related variables to describe archive extraction rather than indexes only.
- Replace the `prebuilt` flag with `append_md5_to_dir_name` and update callers/tests.
- Suppress tqdm output when `download_url(..., verbose=False)` is used.

## Validation

- `mamba run -n pyserini-dev3 python -m unittest tests.base.test_utils`


lintool:
I will leave this as draft since there are lots of other PRs in flight before this. Will test before releasing for review.

lintool:
@sahel-sh @lilyjge @nourj98 @lingwei-gu - this is ready for review.

cleaning up a bunch of debt wrt misnamed methods, params, etc.

general sanity check is fine, I'll run repro etc. to pick up subtle failures...

pyserini/util.py
```diff
@@ -427,10 +430,10 @@ def download_encoded_queries(query_name, force=False, verbose=True, mirror=None)
     query_md5 = QUERY_INFO[query_name]['md5']
     for url in QUERY_INFO[query_name]['urls']:
         try:
-            return download_and_unpack_index(url, index_directory='queries', prebuilt=True, verbose=verbose, md5=query_md5)
-        except (HTTPError, URLError) as e:
+            return download_and_unpack_archive(url, output_dir='queries', append_md5_to_dir_name=True, md5=query_md5)
```
lilyjge: checking it's intentional we stopped passing in verbose

lilyjge (approved):

pyserini/util.py
lintool: Good catch, this should preserve the caller's setting. Fixed by passing `verbose=verbose` through `download_encoded_queries()` to `download_and_unpack_archive()`.

Validation: `mamba run -n pyserini-dev3 python -m unittest tests.base.test_utils`

PR #2598, opened 2026-07-06
Fix impact searcher .json encoder detection (endswith, not encode)

yashs33244:
`_init_query_encoder_from_str` used `query_encoder.encode('json')`, which raises `LookupError: unknown encoding: json` ('json' is not a codec). For an existing encoder file whose name does not end in `jsonl` (e.g. a `.json` cached-query file, the case this branch handles), this crashes instead of routing to `CachedDataQueryEncoder`.

Changed it to `query_encoder.endswith('json')`, which was clearly intended.

### DIFF
--- a/pyserini/search/lucene/_impact_searcher.py
+++ b/pyserini/search/lucene/_impact_searcher.py
@@ -373,7 +373,7 @@ class LuceneImpactSearcher:
     def _init_query_encoder_from_str(query_encoder):
         if query_encoder is None:
             return TokFreqQueryEncoder()
-        elif os.path.isfile(query_encoder) and (query_encoder.endswith('jsonl') or query_encoder.encode('json')):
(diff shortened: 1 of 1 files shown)


lintool:
@yashs33244 thanks for reporting this issue - what's the command line invocation that elicits this error so I can check?

yashs33244:
Thanks for taking a look! It shows up when `--encoder` points to a precomputed-queries file ending in `.json` (not `.jsonl`), which should load a `CachedDataQueryEncoder`. For example:

```
python -m pyserini.search.lucene --impact \
  --index <impact-index> \
  --encoder path/to/encoded_queries.json \
  --topics <topics> --output run.txt --hits 10
```

Since the file exists but `endswith('jsonl')` is `False`, it falls through to `query_encoder.encode('json')`, and `str.encode('json')` raises `LookupError: unknown encoding: json`. A `.jsonl` path avoids it only because the first clause short-circuits, so the intended check was `endswith('json')`.

PR #2599, opened 2026-07-06
Fix predefined query order lookup for file-path and underscore topics

yashs33244:
`get_predefined_order` tests membership with a normalized key (`Path(topics_path).stem` with `_` replaced by `-`) but then looks up `QUERY_IDS[topics_path]` using the raw path. `PREDEFINED_ORDER` and `QUERY_IDS` share the same bare hyphenated keys, so a file-path or underscore `--topics` value passes the membership test and then raises `KeyError` on the lookup.

Changed the lookup to use `normalized_path`, matching the membership test.

### DIFF
--- a/pyserini/query_iterator.py
+++ b/pyserini/query_iterator.py
@@ -84,7 +84,7 @@ class QueryIterator(ABC):
             # Lazy import:
             from pyserini.query_iterator_order_info import QUERY_IDS
 
-            order = QUERY_IDS[topics_path]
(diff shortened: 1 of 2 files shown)


lintool:
@yashs33244 can you write a test case for this?

yashs33244:
> @yashs33244 can you write a test case for this?

Added...!!

lintool (approved):

PR #2606, opened 2026-07-13
Add reproduction log for SPLADE-v3 guide

abubinfahd:
## Summary

- Added a reproduction log entry for the "A Deeper Dive into Learned Sparse Representations" guide.
- Reproduced the experiments using the current main trunk commit.

## Notes

- Environment: Ubuntu 24.04
- Python 3.12
- CPU-only execution


lintool:
hi @abubinfahd please send a single PR that has all reproduction entries.

Issue #2642, opened 2026-08-16
Topic loading is broken against Anserini master after anserini#3381


AhmadT198:
`anserini#3381` reordered the two arguments of
`TopicReader.getTopicsWithStringIdsFromFileWithTopicReaderClass`. Pyserini's call sites at
`pyserini/search/_base.py:174` and `:183` still pass `(className, file)`, so every topic load fails
against an Anserini fatjar built from master.

Anserini commit `fe3a07e4`, merged 2026-08-15:

```diff
- public static Map<String, Map<String, String>> getTopicsWithStringIdsFromFileWithTopicReaderClass(String className, String file) {
+ public static Map<String, Map<String, String>> getTopicsWithStringIdsFromFileWithTopicReaderClass(String file, String className) {
```

Three files changed there, including Anserini's own test, so Anserini's CI is green. This is a
downstream break.

It affects anyone following the **Development Installation** in `docs/installation.md`, which
instructs building Anserini from source and copying the SNAPSHOT fatjar into
`pyserini/resources/jars/`. Users on the released fatjar are unaffected.

### Symptom

```
ValueError: Unable to load topic msmarco-passage-dev-subset from /work/cache/topics/topics.msmarco-passage.dev-subset.txt!
```

The message names the topics file, which is downloaded, complete and readable. The method received
the class name where it expected a path, failed to open it, and returned null. The misdirection
sends you to inspect a file that is fine.

### Evidence

Reproduced in a clean container. Same Pyserini clone at `f8af512`, same cache, the topics file
already on disk so no download is involved, same command, same JVM. Each jar was staged in its own
directory and selected with `ANSERINI_CLASSPATH`, so the jar is the only variable:

| Jar | Built at | `get_topics('msmarco-passage-dev-subset')` |
|---|---|---|
| `2.2.0`, released | Pyserini 2.3.0 sdist | `6980` |
| `2.2.1-SNAPSHOT` | `6caeb33a`, the parent of #3381 | `6980` |
| `2.2.1-SNAPSHOT` | `cb611fa2`, current master | `ValueError` above |

The last two rows are the argument: same version string, same clone, same build command, differing
by exactly one commit. `git rev-parse fe3a07e4^` is `6caeb33a`.
It fails silently because #3381 did not change the compiled signature. `javap` reports the same
thing for both jars:

```
getTopicsWithStringIdsFromFileWithTopicReaderClass(java.lang.String, java.lang.String)
```

[...]

PR #2643, opened 2026-08-16
Add to onboarding reproduction logs, plus doc fixes for the tools/ submodule removal

AhmadT198:
This covers the Pyserini half of the onboarding path: `experiments-msmarco-passage.md`,
`conceptual-framework.md`, `experiments-nfcorpus.md`, `conceptual-framework2.md` and
`conceptual-framework3.md`. All five reproduced.

It is larger than a typical onboarding PR. The extra size comes from documentation that still
referenced the `tools/` submodule after #2632 removed it: two of the five guides pointed at paths
inside it, so those commands no longer run. If you only want the reproduction log entries, they are
`d59b816` on its own, which touches the usual five files and nothing else.

### Environment

| | |
|---|---|
| OS | Ubuntu 24.04.2 LTS, kernel 6.8.0-45-generic, x86_64 |
| CPU | Intel Core i7-10750H @ 2.60GHz, 6 cores / 12 threads |
| RAM | 15 GiB |
| GPU | none, everything ran on CPU |
| Java | OpenJDK 21.0.7 |
| Python | 3.12.3, venv |
| Pyserini | editable install at `f8af512` |
| Anserini jar | `anserini-2.2.0-fatjar.jar`, md5 `1b0a84bbdfc1674749fa9cf577b30e8b` |
| torch / transformers / faiss-cpu | 2.13.0 / 5.15.0 / 1.15.0 |
| numpy / scipy / onnxruntime / pyjnius | 2.5.2 / 1.18.0 / 1.28.0 / 1.7.0 |

Two deviations from the documented setup:

- A plain venv on system Python 3.12 rather than conda.
- The released 2.2.0 fatjar rather than one built from Anserini master. This one is forced: building
  master currently produces a jar that cannot load any topic. I traced that to a single commit,
  anserini#3381, since a jar built from its parent `6caeb33a` loads topics normally, and filed it as
  #2642.

### Results

| Guide | Result |
|---|---|
| `experiments-msmarco-passage.md` | MRR@10 `0.18741227770955546`, recall@1000 `0.8573`, map `0.1958` where the guide says `0.1957` |
| `conceptual-framework.md` | BM25 vector for docid 7187158 identical term for term; hand-computed inner product `17.949487686157227` against `LuceneSearcher`'s `17.94950` |
| `experiments-nfcorpus.md` | nDCG@10 `0.3808` |
| `conceptual-framework2.md` | nDCG@10 `0.3218`; re-encoding a document reproduced its stored vector to `3.59e-07`; by-hand dot product `11.930485963821411` matching `LuceneSearcher` |
| `conceptual-framework3.md` | nDCG@10 `0.3624`, interactive top 10 exact |

The MS MARCO passage index was built from scratch: 8,841,823 documents, 352,316,036 terms.
Everything matched except `map`, which is covered under findings.

### What changed

| Commit | Change |
|---|---|
| `488f773` | `installation.md`, the clone flag and the `tools/eval` build steps |
| `447de19` | `experiments-msmarco-passage.md`, six stale `tools/` paths |
| `d59b816` | reproduction log entries, all five guides |
| `332d790` | `conceptual-framework3.md`, `index/` to `indexes/`, plus `.gitignore` |
| `db66c19` | `conceptual-framework3.md`, encoder batch-size note |

Each commit message explains its own reasoning, so the log is probably more useful than a summary
here. Each deletion removes a command that no longer runs, rather than one that works.

**The `.gitignore` line in `332d790` is the only non-documentation change in this PR.**

One point worth stating outside the log, since it covers the largest deletion:
`java -cp <fatjar> trec_eval --version` reports **9.0.4**, the same version the removed step
compiled from `trec_eval.9.0.4.tar.gz`. The jar carries prebuilt binaries for linux-amd64,
linux-i386, macosx-aarch64, macosx-x86_64 and win-x86. `castorini/eval` does still host that
tarball, so repointing the step there was an option; I removed it because the jar the install
already requires provides the same binary. I could not find any reference to `ndeval` in the
Pyserini codebase.

### Verification

I re-ran the fixed guides end to end in a clean Ubuntu 24.04 container, with the `topics/`, `qrels/`
and `eval/` cache directories emptied, no `tools/` directory, and no `trec_eval` binary installed
anywhere in the image.

### Findings

**Fixed here:** the stale `tools/` paths, the `.gitignore` entry, and the batch-size note.

**Noticed but not fixed:**

- `map` is given as `0.1957`; I measured `0.1958` twice, and
  `integrations/core/sparse/test_prebuilt_msmarco_v1_passage.py` asserts `0.1958` for this exact
  configuration. The guide and the test suite disagree.
- `experiments-nfcorpus.md` says 32 is the default batch size; `pyserini/encode/__main__.py`
  declares `--batch-size` with `default=64`.
- 13 further docs still reference `tools/`, plus `scripts/rest_concurrency_benchmark.py` and
  `scripts/trec-covid-ranker.py`, which carry dead paths as argument defaults.

### Follow-up

Happy to send a separate PR for the remaining 13 docs and 2 scripts, if useful. I left them out
because they cover experiments I did not run, so I cannot verify a fix the way I could for the two
guides on the onboarding path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__
 pyserini/resources/jars/*.jar
 collections/*
 indexes/*
+encode/*
 .vscode/
 venv/
 
--- a/docs/conceptual-framework.md
+++ b/docs/conceptual-framework.md
@@ -524,3 +524,4 @@ If you have any questions, look at previous pull requests for examples.
 + Results reproduced by [@Hamza-Nadif](https://github.com/Hamza-Nadif) on 2026-08-04 (commit [`7b13d9b`](https://github.com/castorini/pyserini/commit/7b13d9ba2dbb07bd6ceee269564f972a394f26ac))
 + Results reproduced by [@nomsou](https://github.com/nomsou) on 2026-08-07 (commit [`b81d99c`](https://github.com/castorini/pyserini/commit/b81d99c868e39cdbc9e420425e9b123dfa55bb95))
 + Results reproduced by [@Navid-Ebadi-2003](https://github.com/Navid-Ebadi-2003) on 2026-08-09 (commit [`583bbda`](https://github.com/castorini/pyserini/commit/583bbda9413bf81a57f043c7a2c9b3009dfc34ec))
++ Results reproduced by [@AhmadT198](https://github.com/AhmadT198) on 2026-08-16 (commit [`f8af512`](https://github.com/castorini/pyserini/commit/f8af512b484848924d462a67dd0d1cdcd419eb64))
--- a/docs/conceptual-framework2.md
+++ b/docs/conceptual-framework2.md
@@ -780,3 +780,4 @@ If you have any questions, look at previous pull requests for examples.
 + Results reproduced by [@Hamza-Nadif](https://github.com/Hamza-Nadif) on 2026-08-04 (commit [`7b13d9b`](https://github.com/castorini/pyserini/commit/7b13d9ba2dbb07bd6ceee269564f972a394f26ac))
 + Results reproduced by [@nomsou](https://github.com/nomsou) on 2026-08-07 (commit [`b81d99c`](https://github.com/castorini/pyserini/commit/b81d99c868e39cdbc9e420425e9b123dfa55bb95))
 + Results reproduced by [@Navid-Ebadi-2003](https://github.com/Navid-Ebadi-2003) on 2026-08-09 (commit [`583bbda`](https://github.com/castorini/pyserini/commit/583bbda9413bf81a57f043c7a2c9b3009dfc34ec))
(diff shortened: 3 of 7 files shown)


lintool:
Thanks for sending this PR. Yes, Pyserini master is currently broken given my recent changes. I'm in the process of sorting it out, and will look at this PR thereafter.

AhmadT198:
> Thanks for sending this PR. Yes, Pyserini master is currently broken given my recent changes. I'm in the process of sorting it out, and will look at this PR thereafter.

Thanks for the heads-up, and no rush. I’ll rebase once master is stable again. If any of the breakage overlaps with what I touched here, I’m happy to help sort it out or split this PR up. Just let me know.

lintool:
@AhmadT198 I'm happy with this PR - please rebase and I'll get it merged in.

AhmadT198:
@lintool Glad to hear you're happy with it! Rebased on the latest master and pushed. Looking forward to seeing it merged, and thanks for the review.

lintool (approved):

PR #2645, opened 2026-08-17
feat: add trust_remote_code option to AutoDocumentEncoder/AutoQueryEncoder

### LINKED ISSUE #2207: Add `trust_remote_code=True` to load models from huggingface
There are some models that require `trust_remote_code=True`. This prevents repetitive prompting to press Y/N on the terminal to load the models every time it is being loaded. A typical use case is when `encoding queries and searching from a precomputed index of documents/passages` using `AutoQueryEncoder`.

Example of model requiring this argument can be found below:

```
tokenizer = AutoTokenizer.from_pretrained('bert-base-uncased')
model = AutoModel.from_pretrained('nomic-ai/nomic-embed-text-v1', trust_remote_code=True)
```

I need this as I am currently running some evals for these models on a cloud gpu using pyserini.

Please let me know if I can quickly whip up a PR to at least add it in for `AutoQueryEncoder` and perhaps  other places where `AutoModel.from_pretained(model)` is being used currently.

St4r4x:
Closes #2207

## What
Adds an optional `trust_remote_code: bool = False` parameter to `AutoDocumentEncoder` and `AutoQueryEncoder`, threaded through to both `AutoModel.from_pretrained()` and the tokenizer loader, and further threaded through `FaissSearcher`/`BinaryDenseFaissSearcher` so it also reaches the "search a precomputed index" workflow the issue itself describes. This lets models that require custom code on the Hub (e.g. `nomic-ai/nomic-embed-text-v1`, named in the issue) load without a blocking Y/N terminal prompt or needing to patch pyserini locally.

## Scope
Scoped to the two `Auto*Encoder` classes named in the issue (the generic `AutoModel.from_pretrained(...)`-based wrappers — model-specific encoders elsewhere in `pyserini/encode/` like uniCOIL/SPLADE/DPR load fixed, known architectures and don't need this knob), plus `FaissSearcher`/`BinaryDenseFaissSearcher`/`from_prebuilt_index` — those are the actual entry points for "encoding queries and searching from a precomputed index" that the issue names as the motivating use case, and the fix wasn't reachable from there in the first commit.

`load_auto_tokenizer()` in `_base.py` already forwards `**kwargs` to `AutoTokenizer.from_pretrained()`, so no change was needed there beyond passing `trust_remote_code` through as a kwarg.

All new parameters default to `False` and are appended after existing params, so existing behavior/call sites are unchanged for everyone not opting in (checked every call site, including one that already passes 3 positional args to `from_prebuilt_index`).

## Not included (happy to add if wanted)
Didn't wire this into the CLI scripts (`scripts/encode_queries.py`, `pyserini/encode/__main__.py`, `pyserini/encode/query.py`'s CLI driver) — no `--trust-remote-code` flag there yet. Let me know if that'd be useful too.

## Validation
Checked every call site of `AutoDocumentEncoder`/`AutoQueryEncoder`/`FaissSearcher` in the repo — all keyword-arg based (or safely appended after existing params), so this is fully backward compatible.

Added `tests/base/search/test_faiss_searcher_trust_remote_code.py`, a mock-based unit test on `_init_encoder_from_str` mirroring the existing SPLADE test's patch-and-assert pattern — no network or model download needed.

Couldn't run the full suite or even import `pyserini.search` for real: this repo needs a built/downloaded Anserini fatjar just to import that package (same wall as #2644), and there's no CI here to compare against either (no `.github/workflows` in this repo). Verified the new tests pass using a local, uncommitted stub of the JVM bridge (`pyserini.pyclass`) instead — confirms the logic, not a real end-to-end run.

> [!NOTE]
> AI-assisted (Claude Code); verified against the source and a local test run before each push.

### DIFF
--- a/pyserini/encode/_auto.py
+++ b/pyserini/encode/_auto.py
@@ -22,11 +22,13 @@ from pyserini.encode._base import DocumentEncoder, QueryEncoder, load_auto_token
 
 
 class AutoDocumentEncoder(DocumentEncoder):
-    def __init__(self, model_name, tokenizer_name=None, device='cuda:0', pooling='cls', l2_norm=False, prefix=None):
+    def __init__(self, model_name, tokenizer_name=None, device='cuda:0', pooling='cls', l2_norm=False, prefix=None,
+                 trust_remote_code=False):
         self.device = device
-        self.model = AutoModel.from_pretrained(model_name)
+        self.model = AutoModel.from_pretrained(model_name, trust_remote_code=trust_remote_code)
         self.model.to(self.device)
-        self.tokenizer = load_auto_tokenizer(tokenizer_name or model_name, clean_up_tokenization_spaces=True)
+        self.tokenizer = load_auto_tokenizer(tokenizer_name or model_name, trust_remote_code=trust_remote_code,
+                                              clean_up_tokenization_spaces=True)
         self.has_model = True
         self.pooling = pooling
         self.l2_norm = l2_norm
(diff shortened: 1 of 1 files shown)


St4r4x:
Extended this to also cover \`FaissSearcher\`/\`BinaryDenseFaissSearcher\`/\`from_prebuilt_index\` (via \`_init_encoder_from_str\`) — the first commit only touched \`AutoDocumentEncoder\`/\`AutoQueryEncoder\` directly, but the issue's own motivating use case ("encoding queries and searching from a precomputed index") normally goes through \`FaissSearcher\`, which didn't have a way to pass \`trust_remote_code\` through. It does now, same append-only/default-False pattern as before. Added a mock-based test for the new dispatch logic. Updated the PR description above to match.

> [!NOTE]
> AI-assisted (Claude Code) — should have flagged this in the original PR description too, apologies for the omission there.

PR #2649, opened 2026-08-21
Deliver pre-generated REST API tokens by email

lingwei-gu:
## Summary

- add an opt-in anonymous `POST /v1/token` endpoint requiring `name` and `email`
- claim credentials from a protected pre-generated JSON pool; the API server never generates tokens
- persist token identities in the YAML registry and activate issued tokens without a restart
- deliver credentials by TLS SMTP to the requester and configured individual operator mailboxes
- keep token-pool, cooldown, and SMTP settings in the YAML server config behind an explicit `--enable-token-issuance` startup gate
- return a generic `202` response without exposing credentials in HTTP responses or request logs
- enforce cooldowns for both client IP and normalized email, with one lifetime token per email
- preserve existing Bearer and `X-API-Key` credentials and reserve partially populated inventory rows
- record optional academic trace metadata, including dataset, question, harness, model, and request sequence
- document endpoint configuration and usage in the REST guide

## Configuration and security

Token delivery remains disabled unless the server starts with `--enable-token-issuance` and its YAML config defines a protected token pool and TLS SMTP delivery settings. SMTP passwords are read from an owner-only file. Production deployments should expose the endpoint over HTTPS.

The pre-generated inventory is maintained separately in private [castorini/trec26-rag#30](https://github.com/castorini/trec26-rag/pull/30). Unassigned inventory tokens are not accepted for authentication.

## Tests

- 95 focused tests pass: `tests.core.test_server_config`, `tests.core.test_token_delivery`, and `tests.core.test_rest`
- coverage includes YAML startup configuration, atomic persistence, inventory reservation, lifetime identity mapping, cooldowns, SMTP delivery, credential redaction, runtime activation, trace logging, and existing-token compatibility
- `git diff --check`


docs/usage-rest.md
```diff
@@ -53,9 +53,63 @@ With `--config` enabled:
...
+`issued_at` fields are available.
+
+Configure a TLS SMTP connection and start the endpoint with both the YAML server config and token pool:
+
+```bash
+python -m pyserini.server.rest \
+  --config /path/to/server.yaml \
+  --enable-token-issuance \
+  --token-pool /secure/path/to/lookup.json \
+  --token-email-smtp-host smtp.example.edu \
```
lilyjge: would it make sense to put all this stuff into a config file instead of sticking it in the CLI?

docs/usage-rest.md
```diff
@@ -133,8 +188,26 @@ Example request log line:
...
+anonymous so that new clients can obtain a credential.
+
+### Optional academic trace metadata
+
+Search and document requests accept optional `qid`, `question`, `run_id`, `agent`, and `step` query
+parameters. Clients are encouraged, as a courtesy, to include them whenever the values are known.
+They are collected solely for academic research on agent retrieval behavior and do not affect ranking
+or response contents.
+
+- `qid`: source-dataset question identifier
```
lilyjge: would it make sense to add a dataset field or specify a format eg dataset-qid? otherwise different datasets can repeat query ids

docs/usage-rest.md
```diff
@@ -133,8 +188,26 @@ Example request log line:
...
+- `question`: complete question body the user or agent is answering
+- `run_id`: stable identifier shared by requests from one answer attempt
+- `agent`: agent or client name and version
+- `step`: zero-based retrieval step within the run
```
lilyjge: what does this mean?

docs/usage-rest.md
```diff
@@ -133,8 +188,26 @@ Example request log line:
...
+
+Search and document requests accept optional `qid`, `question`, `run_id`, `agent`, and `step` query
+parameters. Clients are encouraged, as a courtesy, to include them whenever the values are known.
+They are collected solely for academic research on agent retrieval behavior and do not affect ranking
+or response contents.
+
+- `qid`: source-dataset question identifier
+- `question`: complete question body the user or agent is answering
+- `run_id`: stable identifier shared by requests from one answer attempt
+- `agent`: agent or client name and version
```
lilyjge: might be clearer to specify what we want, eg both model and harness?

pyserini/server/rest/app.py
```diff
@@ -376,10 +540,7 @@ async def rest_api_key_and_access_log(request: Request, call_next):
                     response = JSONResponse(
                         status_code=401,
                         content={
-                            'error': (
-                                'Unauthorized. To request an access token, '
-                                f'email {AUTH_TOKEN_REQUEST_EMAIL}.'
-                            )
+                            'error': 'Unauthorized. Request access from the service operator.'
```
lilyjge: can we be a bit more specific?

pyserini/server/rest/openapi.yaml
```diff
@@ -12,6 +12,60 @@ security:
   - ApiKeyAuth: []
   - BearerAuth: []
 paths:
+  /token:
+    post:
+      summary: Request API token delivery by email
+      description: >
+        Claims a pre-generated 256-bit API token from protected inventory, stores the supplied name
+        and normalized email, activates the credential, and sends it to that email address while
+        copying explicitly configured individual service operators. The server never generates or
```
lilyjge: why don't we generate tokens automatically at this point?

lingwei-gu: this was suggested by Jimmy... we can have one less vulnerability by not allowing token generation in the server

docs/usage-rest.md
lingwei-gu: good call moved to yaml file

docs/usage-rest.md
lingwei-gu: yes added dataset field

docs/usage-rest.md
lingwei-gu: it's i_th request for one question. so if the agent took 10 searches then this will be 0 to 9

docs/usage-rest.md
lingwei-gu: model and agent(harness) are separate fields now

pyserini/server/rest/app.py
lingwei-gu: updated to request api key for bearer token

docs/usage-rest.md
```diff
@@ -53,9 +53,73 @@ With `--config` enabled:
...
+Token requests are disabled by default. When enabled, the server claims credentials from a protected
+`lookup.json` inventory; it never generates credentials. Only entries with null `name`, `email`, and
+`issued_at` fields are available. The inventory is generated and audited offline so credential
+creation remains outside the serving process.
+
+Configure the token pool and TLS SMTP delivery in the same YAML server config:
+
+```yaml
+token_issuance:
+  pool: /secure/path/to/lookup.json
```
lilyjge, line 70: specify what this lookup.json should look like?

lilyjge: also with password file

lilyjge (commented):
sorry for the late review... mostly looks good, just some nits

docs/usage-rest.md
```diff
@@ -133,8 +198,28 @@ Example request log line:
...
+are known. They are collected solely for academic research on agent retrieval behavior and do not
+affect ranking or response contents.
+
+- `dataset`: source dataset name and version, such as `browsecomp-plus-v1`
+- `qid`: question identifier within `dataset`
+- `question`: complete question body the user or agent is answering
+- `run_id`: stable identifier shared by requests from one answer attempt
+- `agent`: retrieval harness or agent name and version
+- `model`: model provider, name, and version
+- `step`: zero-based sequence number of this retrieval request within `run_id`; the first request is `0`
```
lilyjge, line 218: @lingwei-gu I like the explanation you gave more than what's currently here, which still is a bit confusing imo. but from what you said it also sounds like this should be question based, not run id based as it says here?

pyserini/server/rest/openapi.yaml
```diff
@@ -12,6 +12,61 @@ security:
...
   - BearerAuth: []
 paths:
+  /token:
+    post:
+      summary: Request API token delivery by email
+      description: >
+        Claims a pre-generated 256-bit API token from protected inventory, stores the supplied name
+        and normalized email, activates the credential, and sends it to that email address while
+        copying explicitly configured individual service operators. Token generation happens only
+        in the offline inventory workflow, so the server never creates or returns credentials. A
```
lilyjge, line 22: would it make sense to include the token generation script here though so at least everything's in one place?

lingwei-gu: We could but that would be another PR following KISS principle to consolidate stuffs

docs/usage-rest.md
lingwei-gu: Yes it's question based... several requests could share one run_id

lingwei-gu:
> sorry for the late review... mostly looks good, just some nits

No worries. Thanks for the review

docs/usage-rest.md
lingwei-gu: I think it's good practice to keep these files as placeholders. should be defined in deployment

docs/usage-rest.md
lilyjge: so the step number should be a sequence within the question id, not the run id, right?

docs/usage-rest.md
lilyjge: wdym defined in deployment? I wouldn't put the real files here, but just what format is expected?

lilyjge (approved):
I think this is good for now, thanks @lingwei-gu ! we can keep iterating also

PR #2655, opened 2026-08-29
fix: resolve M-BEIR instructions from cache home

### LINKED ISSUE #2475: MBEIR query instructions hard coded to default Pyserini cache directory
MBEIR query instruction yamls point to an instruction file that's hard coded to the default Pyserini cache directory, eg
`instruction_file: ~/.cache/pyserini/query_instructions/query_instructions.tsv`

However, Pyserini generally allows a user-specified cache directory through the environment variable `PYSERINI_CACHE`. Example usage: https://github.com/castorini/pyserini/blob/master/pyserini/util.py#L248

This breaks things if the user has a different cache directory. Example breakage: MCP server

GioiaZheng:
## Why

M-BEIR/UniIR instruction configs are resolved through the Pyserini cache, but downloaded configs may still point `instruction_file` at the default `~/.cache/pyserini` location. That breaks custom `PYSERINI_CACHE` setups.

Fixes #2475.

## What changed

- Added a shared helper that downloads/extracts M-BEIR instruction configs from `get_cache_home()` and rewrites `query_instructions.tsv` references to the active cache path.
- Reused the helper from both the UniIR query encoder and the server backend so CLI and MCP behavior stay aligned.
- Added regression coverage for custom cache resolution, predefined M-BEIR topics, and the default cache-home test expectation on Windows.

## Validation

- `python -m unittest tests.base.test_utils.TestCacheHome tests.base.test_utils.TestMbeirInstructionConfig tests.base.test_query_iterator -v`
- `python -m py_compile pyserini\util.py pyserini\encode\optional\_uniir.py pyserini\server\backend.py tests\base\test_utils.py tests\base\test_query_iterator.py`
- `git diff --check`

## Risk / follow-up

Low. The rewrite only applies to M-BEIR configs whose `instruction_file` basename is `query_instructions.tsv`; custom instruction files are preserved by regression coverage.

### DIFF
--- a/pyserini/encode/optional/_uniir.py
+++ b/pyserini/encode/optional/_uniir.py
@@ -131,30 +131,9 @@ class UniIRQueryEncoder:
     def _get_instruction_config(self, instr_file: str = None):
         """This functions downloads all the instruction config files if not already present."""
 
-        import os
-        import tarfile
-        from pyserini.util import download_url, get_cache_home
+        from pyserini.util import get_mbeir_instruction_config
 
-        cache_dir = get_cache_home()
-        instructions_dir = os.path.join(cache_dir, 'query_instructions')
-
-        if not os.path.exists(instructions_dir):
-            query_images_and_instructions_url = "https://huggingface.co/datasets/castorini/prebuilt-indexes-m-beir/resolve/main/mbeir_query_images_and_instructions.tar.gz"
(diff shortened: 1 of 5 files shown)


lintool:
I have verified the bug and the fix. Thanks for your contributions! Merging.

lintool (approved):

PR #2676, opened 2026-09-09
[codex] Consolidate shared 2CR helpers and simplify runners

lintool:
## Summary

- Consolidate template reading across 11 runners into a shared helper using a context manager.
- Consolidate eight evaluation-command formatters, preserving complete runfile paths when inserting line continuations.
- Simplify imports, conditionals, dictionary operations, printing, and UTC timestamp usage.

## Validation

- `git diff --check`
- `python -B -c 'import ast, pathlib; [ast.parse(p.read_text(), filename=str(p)) for p in pathlib.Path("pyserini/2cr").glob("*.py")]'`

Targeted checks passed for 32 templates, missing-file errors, four shell-argument cases, and shared imports. Full runner execution was not tested; local imports lack `pandas`.

### DIFF
--- a/pyserini/2cr/_base.py
+++ b/pyserini/2cr/_base.py
@@ -14,11 +14,23 @@
 # limitations under the License.
 #
 
+import importlib.resources
+
 from pyserini.util import run_command
 
 fail_str = '\033[91m[FAIL]\033[0m'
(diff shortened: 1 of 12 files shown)


nourj98 (approved):
No complaints - LGTM

PR #2679, opened 2026-09-12
Unify 2CR score classification

### LINKED ISSUE #2675: Centralize 2CR score comparison and unify OK/OKish/FAIL rules
## Problem

Pyserini’s 2CR runners independently classify observed scores as `OK`, `OKish`, or `FAIL`. `_base.py` centralizes the display labels, but comparison logic is duplicated across modules, with different thresholds and treatment of improvements.

## Desired behavior

Introduce a shared classification helper in `pyserini/2cr/_base.py`, following the definitions proposed in [Anserini #3425](https://github.com/castorini/anserini/issues/3425), with configured tolerance treated as zero.

Let `delta = abs(observed - expected)`. Apply these checks **in order**:

1. **OK:** `delta ≤ numericalTolerance`, where `numericalTolerance` is a small, explicitly defined shared floating-point allowance.
2. **OKish:** If not already `OK`, either:
   - `observed > expected`
   - `delta < 0.0002`
3. **FAIL:** Otherwise.

Remove all existing runner-specific hardcoded tolerances, including MS MARCO’s `flaky_dict`. Do not introduce YAML tolerance configuration or benchmark-specific exceptions.

The numerical allowance and strict `< 0.0002` fallback remain shared policy constants. Normalize percentage-scale metrics to a 0–1 scale for comparison, while preserving existing score presentation.

The helper should return a shared status representation; runners remain responsible for evaluation, configuration loading, and output formatting.

## Behavior changes

- Regressions previously accepted under looser thresholds may now fail.
- Improvements outside the numerical allowance consistently become `OKish`.
- CIRAL and ODQA adopt the same three-status policy.
- Existing special-case tolerance allowances are removed.

## Acceptance criteria

- All 2CR runners delegate classification to the shared helper.
- Runner-specific comparison thresholds and `flaky_dict` are removed.
- The numerical tolerance is explicitly defined and documented.
- Comparison uses consistent score units while preserving existing display formats and evaluation precision.
- Tests cover equality, floating-point differences, improvements, regressions, the strict `0.0002` boundary, and equivalent fractional/percentage inputs.

appleweiping:
Closes #2675.

This moves the OK, OKish, and FAIL decision into a shared helper in _base.py and uses it from all 11 2CR runners. The helper applies a documented 1e-9 numerical allowance, treats larger observed scores as OKish, and keeps the 0.0002 fallback strict.

DSE and ODQA mark their scores as percentage-scale so comparison happens on a 0 to 1 scale without changing their displayed values. The MS MARCO flaky exception table and all runner-specific thresholds are removed.

Validation:
- python -m pytest tests/test_2cr_base.py -q: 4 tests and 4 subtests passed
- imported all 11 modified 2CR runner modules successfully
- python -m compileall -q pyserini/2cr tests/test_2cr_base.py: passed
- git diff --check: passed

### DIFF
--- a/pyserini/2cr/_base.py
+++ b/pyserini/2cr/_base.py
@@ -14,11 +14,42 @@
 # limitations under the License.
 #
 
+from enum import Enum
+
 from pyserini.util import run_command
 
 fail_str = '\033[91m[FAIL]\033[0m'
 ok_str = '[OK]'
 okish_str = '\033[94m[OKish]\033[0m'
+
+NUMERICAL_TOLERANCE = 1e-9
+OKISH_TOLERANCE = 0.0002
+
+
+class ScoreStatus(Enum):
+    OK = 'OK'
+    OKISH = 'OKish'
+    FAIL = 'FAIL'
+
+
+def classify_score(observed, expected, *, percentage=False):
+    """Classify an observed score against its expected value.
+
+    ``NUMERICAL_TOLERANCE`` absorbs floating-point noise. Larger improvements
+    and differences strictly below ``OKISH_TOLERANCE`` are ``OKISH``; all
+    remaining regressions are ``FAIL``. Percentage-scale inputs are normalized
+    to the same 0-1 scale before comparison.
+    """
+    scale = 100.0 if percentage else 1.0
+    observed = float(observed) / scale
+    expected = float(expected) / scale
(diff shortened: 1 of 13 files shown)


lintool:
Thanks @appleweiping for your PR, but I was able to close with #2691.

PR #2687, opened 2026-09-15
Preserve BRIGHT query prefixes with argv commands

### LINKED ISSUE #2664: BRIGHT 2CR commands depend on shell-specific ANSI-C quoting
## Problem

The BRIGHT 2CR configuration uses Bash-style ANSI-C quoting for query prefixes:

```bash
--query-prefix $'Instruct: ...\nQuery:'
```

This occurs for the `diver-retriever-4b` and `reason-embed-qwen3-4b-0928` conditions in `pyserini/2cr/bright.yaml`.

The commands currently rely on shell evaluation through `os.system`. Replacing that call with the shared `pyserini.util.run_command`, which parses command strings using `shlex.split`, changes the argument semantics:

- the leading `$` is preserved literally;
- `\n` is not expanded into a newline;
- the resulting query prefixes change the query embeddings and reproduction scores.

For now, BRIGHT should continue using `os.system` to preserve its existing behavior.

## Proposed follow-up

Refactor BRIGHT command construction so query prefixes are represented as data rather than shell syntax. Commands should eventually be passed as argv sequences, with query prefixes constructed as Python strings containing real newline characters.

This would:

- remove reliance on Bash-specific quoting;
- make command execution safer and more portable;
- allow BRIGHT to use the shared `run_command` helper;
- preserve paths and argument values containing whitespace.

## Scope

The affected conditions are:

- `diver-retriever-4b`
- `reason-embed-qwen3-4b-0928`

The refactor should include tests confirming that each query prefix is passed as one argument containing the expected newline.

lilyjge:
BRIGHT's Diver and Reason-Embed commands use Bash ANSI-C quoting for query prefixes. Passing those strings through `run_command` preserves the `$` and literal `\n`, changing the prefix used for query embeddings.

Keep compact YAML command strings and move the two affected prefixes into separate `query_prefix` fields containing real newlines. BRIGHT splits each command template before substituting values and appends the prefix as one argument, preserving its exact text and trailing spaces. Execute the resulting argv through `run_command` with live output. Displayed retrieval commands and reports use the same builder and quote individual arguments for copying into a shell.

The fix is limited to BRIGHT retrieval command construction, execution, display, and regression tests. Evaluation commands, shared helpers, and HTML escaping are unchanged.

Validation:
- `python -m unittest tests.base.test_bright tests.base.test_utils.TestRunCommand tests.base.test_2cr`: 16 tests passed.
- Compared the actual builder against all 72 original Bash commands: argument values match, with `--query-prefix` moved to the end; metadata and expected scores are unchanged.
- Executed all 72 rendered commands through an argv-printing stub under `/bin/sh`: arguments match the builder, including prefix newlines and trailing spaces.
- `git diff --check` passed. Full retrieval benchmarks were not run.

Fixes #2664.


pyserini/2cr/bright.yaml
```diff
@@ -1,6 +1,21 @@
 conditions:
   - name: bm25
-    command: python -m pyserini.search.lucene --index bright-${dataset} --topics bright-${dataset} --output $output --output-format trec --hits 1000 --bm25 --remove-query
+    command:
+      - "python"
```
lintool, line 4: yikes... not sure I like this... not very readable...

is there another way?

pyserini/2cr/bright.py
```diff
@@ -63,16 +66,26 @@
...
+    """Split the template before substitution so values remain single arguments."""
+    values = dict(dataset=dataset['dataset'], output=runfile,
+                  query_prefix=dataset.get('query_prefix', ''))
+    argv = [Template(arg).substitute(values) for arg in shlex.split(condition['command'])]
+    if 'query_prefix' in condition:
+        argv.extend(['--query-prefix', Template(condition['query_prefix']).substitute(values)])
+    return argv
+
+
+def format_run_command(argv):
```
lintool, line 79: this is more elegant than what I had before :)

lintool (approved):
lgtm - I'll run regressions and make sure everything's still good.

lilyjge:
great, thanks! I'll wait for your regressions before merging then?

lintool:
nah, go ahead and merge... I'll batch up with next series of changes, and let you know if anything breaks.

PR #2691, opened 2026-09-18
[codex] Unify 2CR reproduction score comparison

### LINKED ISSUE #2675: Centralize 2CR score comparison and unify OK/OKish/FAIL rules

lintool:
## Summary

- Centralize all 11 runners on `compare_reproduction_score(observed, expected)`, returning `ReproductionStatus`: differences ≤ `1e-9` are OK; otherwise improvements or differences strictly below `0.0002` are OKish; remaining results fail.
- Compare decimal string representations to avoid binary subtraction artifacts at the thresholds, without fixed-decimal rounding or additional tolerance.
- Remove runner-specific thresholds and MS MARCO’s `flaky_dict`. DSE and ODQA normalize percentage scores before comparison.
- Share status formatting through `format_reproduction_status`, preserving evaluation precision and score presentation.
- Add tests for equality, numerical differences, improvements, regressions, nonzero boundaries, normalized percentages, and all runners.

Closes #2675.

## Validation

- `mamba run -n pyserini-dev3 python -m unittest tests.base.test_2cr -v`
- `git diff --check`

All 7 tests pass, including 55 runner scenarios with mocked evaluation scores. Full retrieval benchmarks were not run.


pyserini/2cr/_base.py
```diff
@@ -23,6 +24,38 @@
...
+    results fail. There is no relative or configurable tolerance.
+    """
+    if percentage:
+        observed /= 100
+        expected /= 100
+
+    delta = abs(observed - expected)
+    if delta <= NUMERICAL_TOLERANCE:
+        return ScoreStatus.OK
+    if observed > expected or delta < OKISH_THRESHOLD:
```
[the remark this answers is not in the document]
lintool: Fixed in 986b7f1cf. The helper now compares decimal string representations of the scores and thresholds, so `0.5000` versus `0.5002` produces FAIL without fixed-decimal rounding or additional tolerance.

Percentage normalization now lives in the DSE and ODQA callers; the helper takes only observed and expected scores. Tests cover normalized `50.00` versus `50.02`, nonzero boundary cases, and values immediately below and above the cutoff. All 7 tests pass, including 55 runner scenarios with mocked evaluation scores.

pyserini/2cr/odqa.py
```diff
@@ -430,10 +429,8 @@ def run_topic_conditions(args, topic_arg, default_topics, yaml_path):
                     if not args.skip_eval and metric not in score.keys():
                         continue
                     if not args.skip_eval:
-                        if math.isclose(score[metric], float(expected_score), abs_tol=2e-2):
-                            result_str = ok_str
-                        else:
-                            result_str = fail_str + f' expected {expected[metric]:.4f}'
+                        status = compare_reproduction_score(score[metric] / 100, float(expected_score) / 100)
```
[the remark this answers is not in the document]
lintool: Will punt on this because we'll refactor this later.

pyserini/2cr/atomic.py
```diff
@@ -224,13 +222,8 @@ def run_conditions(args):
...
 
-                            if math.isclose(score, float(expected[metric])):
-                                result = ok_str
-                            # If results are within 0.0005, just call it "OKish".
-                            elif abs(score - float(expected[metric])) <= 0.0005:
-                                result = okish_str + f' expected {expected[metric]:.4f}'
-                            else:
-                                result = fail_str + f' expected {expected[metric]:.4f}'
+                            status = compare_reproduction_score(score, float(expected[metric]))
+                            result = format_reproduction_status(status, expected[metric])
```
lilyjge, line 226: nit but why is one casted to float and one isn't? in the method signatures both methods are expecting floats

pyserini/2cr/msmarco.py
```diff
@@ -503,16 +507,6 @@ def generate_report(args):
             out.write(Template(html_template).substitute(title=full_name, rows=all_rows))
 
 
-FlakyKey = namedtuple('FlakyKey', ['collection', 'name', 'topic_key', 'metric'])
```
lilyjge, line 506: just double checking that this removal is intended

lilyjge (approved):
the number of places that this needed to be changed makes me think that some abstraction might be useful, but that can be a later problem. nits but otherwise lgtm!

pyserini/2cr/msmarco.py
lintool: correct.

pyserini/2cr/atomic.py
lintool: Good catch—the inconsistent cast was carried over from the original comparison and formatting code. Fixed in 22908dbb9: each runner now converts the expected score to float once and reuses it for both helpers. DSE and ODQA still divide by 100 for comparison and retain percentage units for display. All 7 tests pass, including 55 runner scenarios.

pyserini/2cr/dse.py
```diff
@@ -125,13 +129,9 @@ def run_conditions(args):
...
 
-                            if math.isclose(score, float(expected[metric]), abs_tol=0.1): 
-                                result = ok_str
-                            elif abs(score - float(expected[metric])) <= 0.5:
-                                result = okish_str + f' expected {expected[metric]:.1f}'
-                            else:
-                                result = fail_str + f' expected {expected[metric]:.1f}'
-                            
+                            expected_score = float(expected[metric])
+                            status = compare_reproduction_score(score / 100, expected_score / 100)
```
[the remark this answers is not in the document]
lintool: we'll pick up the DSE repro refactoring later.

